PR #102, opened 2026-05-26
perf: skip toolpath recalculation on display-only operation field changes

zabooma:
## Summary

- Toggling an operation's eye icon (`showToolpath`) or enable checkbox (`enabled`) previously triggered a full toolpath recompute because Zustand's immutable updates always produce a new object reference, causing a cache miss in `isCacheHit`.
- Added `operationComputationEquals()` in `src/App.tsx` that compares only the 26 fields that actually affect toolpath geometry, explicitly excluding `name`, `enabled`, and `showToolpath`.
- `isCacheHit` now calls this helper instead of using reference equality (`===`) on the operation object.

## Result

Visibility and enable toggles are now instant — no spinner, no recompute. Any genuine cutting parameter change (tool, stepdown, feed, target features, pocket pattern, etc.) still invalidates the cache as before.

## Test plan

- [x] Toggle the eye icon on an operation — no spinner, toolpath result unchanged
- [x] Toggle the enable checkbox on an operation — no spinner, toolpath result unchanged
- [x] Change a cutting parameter (e.g. stepdown) — spinner appears and path updates correctly
- [x] `npm run build` passes (tsc + all 17 test suites green)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1918,9 +1918,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "Apache-2.0 OR MIT",
       "optional": true,
       "os": [
@@ -1938,9 +1935,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "Apache-2.0 OR MIT",
       "optional": true,
       "os": [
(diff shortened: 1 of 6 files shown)


zabooma:
Review note:

I did not find a blocking functional issue. The cache comparison covers the current Operation fields, and npm run build passed in an isolated worktree.

One cleanup before merge: package-lock.json appears to contain unrelated libc metadata churn. Please remove that lockfile noise unless it was intentional for this PR.

PR #103, opened 2026-05-26
fix(simulation): rebuild boundary wall geometry when cells are cut through during playback

zabooma:
## Problem

During tool playback, vertical walls on pocket boundaries would appear as open space as the tool cut through them. The static simulation result (without playback) was correct — the issue was specific to the animated play mode.

### Root cause

The boundary wall geometry is built **once** when playback starts, based on the initial grid state where all cells are full. The geometry only emits wall quads where a cell has a neighbor with no material (i.e. the outer stock profile boundary). As the tool cuts cells all the way to `stockBottomZ`, those cells become effectively empty — but their still-full neighbors were never given wall geometry facing inward, since they weren't on any boundary at creation time. The heightfield texture update kept the top surface correct, but no wall faces existed for the newly-exposed sides.

## Changes

**`src/engine/simulation/replay.ts`**
- `applyMoveToGrid` now returns `anyCellCleared: boolean` — set to `true` when a cell transitions from having material (`topZ > stockBottomZ`) to empty (`topZ <= stockBottomZ`)

**`src/engine/simulation/playback.ts`**
- `PlaybackController` accumulates a `boundaryChanged` flag whenever `anyCellCleared` is reported by any move in `advance()`
- Also set to `true` on `reset()` so boundary geometry is always correct after stop/replay
- Exposed via `getBoundaryChanged()` / `clearBoundaryChanged()`

**`src/components/simulation/SimulationViewport.tsx`**
- `rebuildPlaybackGeometry` now checks `boundaryChanged` after updating the heightfield texture
- When set, disposes the old boundary mesh and rebuilds it from the current live grid state, so newly-exposed wall faces appear immediately
- Boundary rebuild only fires when cells actually clear — no per-frame overhead during normal cutting

## Testing

1. Open an operation with a pocket or profile cut
2. Enable Tool Playback
3. Play through the operation — pocket walls should remain solid throughout, matching the final static simulation result

### DIFF
--- a/src/components/simulation/SimulationViewport.tsx
+++ b/src/components/simulation/SimulationViewport.tsx
@@ -576,6 +576,11 @@ export const SimulationViewport = forwardRef<SimulationViewportHandle, Simulatio
   const playbackMaterialMeshRef = useRef<THREE.Object3D | null>(null)
   const playbackBoundaryMeshRef = useRef<THREE.Object3D | null>(null)
   const playbackHeightfieldTextureRef = useRef<THREE.DataTexture | null>(null)
+  // Cached stock color for boundary geometry rebuilds during playback.
+  const playbackStockColorRef = useRef<THREE.Color>(new THREE.Color(0xb5beca))
+  useEffect(() => {
+    playbackStockColorRef.current = stockColor ? new THREE.Color(stockColor) : new THREE.Color(0xb5beca)
+  }, [stockColor])
   const boundaryMeshRef = useRef<THREE.Object3D | null>(null)
   const playbackFrameRef = useRef<number>(0)
   const playbackLastTimeRef = useRef<number>(0)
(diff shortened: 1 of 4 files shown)


zabooma:
Review note:

The implementation looks reasonable, and npm run build passed, but this is an engine-level bug fix and changes the ApplyMoveResult/playback boundary contract without focused tests.

Please add unit coverage for:
- applyMoveToGrid returning anyCellCleared: false for partial cuts that leave material
- applyMoveToGrid returning anyCellCleared: true when a cell transitions to stockBottomZ
- PlaybackController setting and clearing boundaryChanged when a cut-through happens

Verification I ran: npm run build on an isolated worktree for this PR.

zabooma:
Added unit coverage in `src/engine/simulation/playback.test.ts`:

- `applyMoveToGrid` returns `anyCellCleared: false` for a partial cut that leaves material above stockBottomZ
- `applyMoveToGrid` returns `anyCellCleared: true` when a cell transitions to stockBottomZ
- `applyMoveToGrid` does not re-report `anyCellCleared` on a repeated cut of an already-empty cell (no false positives once a cell is gone)
- `PlaybackController` starts with `boundaryChanged=false`, sets it to `true` after `advance()` processes a cut-through, and `clearBoundaryChanged()` clears it
- `PlaybackController` does not set `boundaryChanged` for an `advance()` whose only effect is a partial cut
- `PlaybackController.reset()` forces `boundaryChanged=true` so walls always match the base grid state after stop/replay

Verified: `npx tsx src/engine/simulation/playback.test.ts` passes in isolation, and `npm run build` (which runs all 18 test files via `scripts/run-tests.ts`) passes end-to-end.

PR #104, opened 2026-05-26
Fix operation selection UX: touch targets, hints, chevrons, feedback

zabooma:
## Summary

Seven targeted UX improvements to the operation selection panel, covering tablet touch targets, discoverability, and feedback.

- **P1 — Touch targets (tablet):** `.cam-subtab--compact` was overriding the tablet 44 px touch target rule on the Rough / Finish / Both buttons. Fixed by adding an explicit `pointer: coarse` override in `tablet.css`.
- **P2 — Add button hint:** When no geometry is selected, the "Add" button now turns amber and shows a tooltip explaining why all operations are disabled. The button stays enabled so users can still browse descriptions.
- **P3 — Menu width on narrow screens:** `.cam-add-menu--vertical` now uses `width: min(360px, calc(100vw - 2rem))` instead of separate `width` + `max-width` properties.
- **P4 — Expandable card chevron:** Each operation label button in the add menu now shows a `chevron-down` icon that rotates 180° when the card is open — makes the expand/collapse action discoverable.
- **P5 — "Use current selection" confirmation:** After a successful target update, a green "✓ Target updated" message appears for 2 s, replacing the previous silent success.
- **P6 — Duplicate icon:** The Unicode `⧉` glyph on the duplicate-operation button is replaced with `<Icon id="copy" size={14} />` for consistent rendering across platforms.
- **P7 — Operation row height:** `.cam-operation-row` min-height raised from 24 px to 40 px for easier clicking.

## Test plan

- [x] On a touch device (or Chrome devtools coarse pointer emulation), confirm Rough / Finish / Both buttons are ≥ 44 px tall
- [x] With no features selected, open the Add menu — button should be amber with tooltip
- [x] Resize browser to a narrow width — Add menu should not overflow the viewport
- [x] Click an operation label in the Add menu — chevron should rotate; click again to collapse
- [x] With a valid selection, click "Use current selection" — green confirmation should appear for ~2 s
- [x] Duplicate an operation — icon should render as a copy symbol (not `⧉`)
- [x] Operation rows in the Add menu should be visibly taller and easier to click

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ACTION_PLAN.md
+++ b/ACTION_PLAN.md
@@ -1,0 +1,74 @@
+# Operation Selection UX Action Plan
+
+## Background
+UX review of the operation selection UI in CAMPanel identified seven improvements
+for desktop and tablet/touch users.
+
+## Items
+
+### P1 — Touch targets for compact pass buttons (tablet.css)
+**Problem:** `.cam-subtab--compact` overrides the base `.cam-subtab` min-height down to
+20 px. The tablet rule that fixes `.cam-subtab` to 44 px doesn't override `--compact`
+because specificity is equal. On a coarse-pointer device the Rough / Finish / Both
+buttons are too small to tap reliably.
+
+**Fix:** Add `@media (pointer: coarse)` block in `tablet.css` that restores
+`min-height: var(--touch-target, 44px)` on `.cam-subtab--compact` and bumps padding and
+font-size to match the base `.cam-subtab` touch target.
+
+---
+
+### P2 — Inform "Add Operation" button when no geometry is selected (CAMPanel.tsx + layout.css)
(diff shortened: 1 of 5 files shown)


zabooma:
Review note:

I found one issue in src/components/cam/CAMPanel.tsx around the transient 'Target updated' confirmation. The new setTimeout is untracked, so multiple quick successful updates can race: an older timer can clear a newer confirmation early. It also is not cleaned up if the component unmounts.

Please store the timeout id in a ref, clear it before scheduling a new timer, and clear it in an effect cleanup.

Verification I ran: npm run build on an isolated worktree for this PR.

PR #105, opened 2026-05-26
feat: post-import arc/circle simplification pass for DXF

zabooma:
## Problem

DXF exporters sometimes approximate arcs, circles, and splines as sequences of short line segments (polyline approximations). These import into PureCut as dozens of individual `line` segments rather than proper `arc` or `circle` geometry, which degrades downstream toolpath quality and makes the sketch harder to work with.

## Solution

A new post-import simplification pass (`src/import/simplify.ts`) runs automatically after DXF stitching and performs three steps:

### Pass 1 — Collinear merge
Consecutive `line` segments whose directions are nearly parallel (|sin θ| < 1e-4, i.e. < 0.006°) are merged into a single segment.

### Pass 2 — Arc fitting
Greedy left-to-right scan over runs of ≥ 6 consecutive `line` segments. For each candidate run, a **Kasa least-squares circle fit** is computed. The run is accepted as an arc when:
- Every input point deviates from the fitted circle by ≤ 1% of the radius (`toleranceFraction`, configurable)
- No single sub-segment subtends more than 60° of arc (prevents a chord "home" segment absorbing into a full-circle fit)

### Pass 3 — Circle promotion
A profile that reduces to a single `arc` whose start and end points coincide within tolerance is promoted to a `circle`.

## Integration

`importDxfString` automatically applies `simplifyProfile` to every stitched shape. Pass `arcSimplifyOptions: false` in `ImportContext` to disable per-call.

## Configurable options

| Option | Default | Meaning |
|---|---|---|
| `toleranceFraction` | `0.01` | Max deviation as a fraction of fitted radius |
| `minSegments` | `6` | Min consecutive line segments to attempt arc fit |
| `maxSubtendedAngleDeg` | `60` | Max angle per sub-segment (guards against false fits) |

## Spline fitting

Left as a TODO — fitting Bézier/NURBS curves from a polyline strip is substantially more involved. Arc/circle covers the most common real-world DXF regressions.

## Tests

9 unit tests in `src/import/simplify.test.ts` covering: collinear merge, arc detection (quarter, half, three-quarter circles), full-circle promotion, mixed geometry, rejection of non-circular curves, and the min-segments guard.

All 18 tests pass, TypeScript compiles cleanly, Vite build succeeds.

## Test plan

- [ ] Import a DXF file that was exported with arc-as-polyline approximation — arcs should come in as `arc` segments
- [ ] Import a DXF circle exported as a closed polyline — should become a `circle`
- [ ] Import a DXF with genuine line segments — no unintended arc detection
- [ ] Verify `arcSimplifyOptions: false` disables the pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -1,0 +1,71 @@
+# Plan: DXF Post-Import Arc/Circle Simplification
+
+**Branch:** `feat/dxf-arc-simplification`
+**Status:** In progress
+
+## Problem
+
+DXF files sometimes export true arcs and circles as dense polylines (many short line
+segments). SPLINE entities are also sampled into line strips during import. The result
+is that imported geometry has hundreds of tiny lines where a handful of arc segments
+would be correct and far more compact.
+
+## Scope
+
+- **In scope:** collinear-line merging, arc detection, full-circle detection
+- **Out of scope (TODO):** spline fitting (Bézier/NURBS recovery from line approximations)
+
+## Solution
+
+Add a post-import simplification pass in `src/import/simplify.ts` that runs on each
+`SketchProfile` after stitching in `importDxfString`.
+
+### Passes (in order)
+
+1. **Collinear-line merge** — consecutive `line` segments whose directions are nearly
+   parallel (sine of included angle < 1e-4) are merged into a single segment.
+
+2. **Arc fitting** — greedy left-to-right scan over runs of consecutive `line` segments.
+   For each run, try from the longest possible sub-run down to `minArcSegments` (default 6).
+   Use the Kasa least-squares circle fit; accept if max point deviation ≤
+   `radiusToleranceFraction × radius` (default 1%). Replace accepted runs with a single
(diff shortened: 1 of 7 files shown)


zabooma:
Review note:

I found a blocking geometry issue in the DXF simplifier. In src/import/simplify.ts, the arc/full-circle detection can rewrite intentional faceted geometry. I reproduced this on the PR branch: a closed regular 8-gon and 12-gon simplify to a single circle. That changes user-authored polygonal DXF geometry into curved cuts.

The default minArcSegments: 6 plus 'all points lie on a circle' is too permissive. Before merge, please add negative tests for regular polygons and tighten acceptance, for example by requiring a much higher segment count for full circles or a smaller max per-segment angular span.

Verification I ran: npm run build, npx tsx src/import/simplify.test.ts, plus the custom polygon reproduction above.

zabooma:
No need for this for now

PR #108, opened 2026-05-28
perf(simulation): shader-driven playback walls + tab responsiveness

zabooma:
## Summary

- Replaces the per-tick boundary-mesh rebuild from #103 with a **static shader-driven boundary mesh**. Walls are emitted once at every grid edge (interior + outer), one floor face per cell; the vertex shader samples both adjacent heightfields each frame to compute wall extent, and the fragment shader discards floors where the cell still has material and walls where the step is zero. No CPU mesh rebuilds during playback.
- Throttles `playbackProgress` / `displayPose` setState to ~10 Hz, fixing the "Maximum update depth exceeded" cascade that appeared near 90% of an edge-out rough.
- Adds spinners + render-loop gating so the simulation tab feels responsive:
  - Operation selection while in simulation view is now wrapped in `startSimulationTransition`, surfacing the existing `isComputing` overlay during the recompute.
  - "Play tool" click sets a new `isPlaybackBuilding` state and defers the heavy mesh build to `requestAnimationFrame`, so the spinner paints before the main thread is blocked.
  - New `isActive` prop skips `renderer.render()` when the simulation tab isn't visible, so switching to another tab is no longer held up by a heavy sim frame.
- Strips the now-dead `anyCellCleared` / `boundaryChanged` plumbing from `replay.ts` and `playback.ts`, and removes `playback.test.ts` (the entire file was tests for the removed contract).

## Background

The previous fix in #103 was semantically correct (it detected when a cell transitioned through to `stockBottomZ` and rebuilt the wall mesh so newly exposed walls appeared), but the rebuild walks the whole grid and re-allocates a multi-MB `Float32Array`. At the bottom-Z pass of an edge-out rough that fires every tick — for the T-53 body file it pinned the main thread, blew the React update-depth guard, and made the rest of the UI sluggish.

Moving wall visibility into the shader removes the rebuild entirely while still producing correct walls the moment a cell transitions, because every potential wall is already in the mesh from frame 1 and only collapses to zero extent until a neighbor empties.

## Test plan

- [x] `npm run build` (typecheck + 19 test files pass + vite build)
- [x] Hard-reload, open `work/T-53-body.camj`
- [x] Play the simulation — verify it plays smoothly through the bottom-Z pass of the edge-out rough with no console errors and no UI sluggishness
- [x] Rotate the model under the stock during playback — verify the bottom of the stock is visible (with holes where cuts go through)
- [x] Compare upper-Z pocket walls vs. current cut walls — both should shade consistently (lit face always faces the empty/lower side now)
- [x] Click between simulation and other tabs — switch should feel immediate in both directions
- [x] Select a different operation while in simulation view — spinner appears during the recompute
- [x] Click "Play tool" on a large project — spinner appears while the playback mesh builds, then disappears


zabooma:
Review finding: the new shader-driven playback boundary mesh scales too large for the existing simulation detail range. `createShaderDrivenBoundaryGeometries()` emits one floor quad plus left/top wall quads for every cell, with right/bottom edge extras. At the current max detail (`targetLongAxisCells = 1500`), a square grid is about 2.25M cells, which becomes roughly 40.5M vertices. Each vertex carries 12 float attributes (`position`, `normal`, `aLeftUv`, `aRightUv`, `aIsTop`, `aIsFloor`), so the geometry alone is around 1.9 GB of typed attribute buffers before Three.js/driver overhead, and the temporary JS arrays during construction are larger. Even the default 240x240 case is about 1M vertices / ~50 MB for this mesh.

This is likely to freeze or crash playback at high detail, which undercuts the responsiveness fix. Please either cap/reduce playback detail for this path, generate a much smaller representation, or add a guard/fallback before enabling the shader-driven boundary mesh at large grid sizes. A focused `gpuMesh.test.ts` assertion for the new geometry vertex-count/scaling behavior would also help keep this from regressing.

zabooma:
Good catch — addressed in 0ebd119.

- Added `SHADER_DRIVEN_BOUNDARY_MAX_CELLS = 500 * 500` in `SimulationViewport.tsx`. Above that threshold the playback path falls back to the static `createDynamicProfileBoundaryGeometries` (built once at playback start, no rebuilds). At 500×500 the shader-driven mesh allocates roughly ~108 MB; above that we trade fidelity (the cosmetic missing-walls-at-cut-through bug from before #103 reappears at very high detail) for not OOMing.
- Exported `SHADER_BOUNDARY_VERTICES_PER_CELL = 18` from `gpuMesh.ts` so the scaling contract is explicit.
- Added `testShaderDrivenBoundaryVertexCountScales()` in `gpuMesh.test.ts` that builds at 16, 24, and 48 cell grids and asserts the per-cell vertex count stays within `[18, 22] × cells + small boundary surcharge`. A regression that bumps the constant or changes the chunk shape now fails the test instead of silently raising memory.
- Left a TODO at the new builder pointing at indexed geometry + Uint8/Uint16 attribute packing as the way to raise the cap (estimate: roughly 4× reduction in attribute memory, lifting the safe ceiling closer to ~700×700 before fallback).

PR #124, opened 2026-05-29
Optimize feature-first toolpath block ordering

### LINKED ISSUE #120: Optimize feature-first operation block ordering and transitions
## Finding

`feature_first` operations are split into per-feature sub-operations and then concatenated in target feature order. `perFeatureOperations()` preserves the incoming feature order (`multiFeature.ts:20-34`), and `mergeToolpathResults()` simply flatMaps each part's moves (`multiFeature.ts:55-69`). Pocket, edge-route, V-carve, and V-carve-recursive all use this path.

That keeps each feature complete before moving to the next, but it does not optimize the order of feature blocks or bridge state between blocks. If selected features are not spatially ordered, the generated program can finish one feature, retract, then rapid across the job even when another nearby feature block was available.

## Suggested direction

- Add a block-ordering pass for `feature_first` results that chooses the next feature block by nearest travel from the previous block endpoint.
- Preserve the core feature-first guarantee: complete all passes for a feature before moving to another feature.
- Keep deterministic tie-breaking by original target order.
- Consider normalizing/bridging the first move of each block so concatenated output does not rely on a zero-length initial rapid to reposition from the previous block.
- Add tests for pocket/edge feature-first operations where target order is intentionally non-spatial and verify rapid travel is reduced or block order is nearest-neighbor.

## References

- `src/engine/toolpaths/multiFeature.ts:20-34`
- `src/engine/toolpaths/multiFeature.ts:55-69`
- `src/engine/toolpaths/pocket.ts:1060-1066`
- `src/engine/toolpaths/edge.ts:306-310`

zabooma:
## Summary
- Add nearest-neighbor ordering for feature-first toolpath result blocks while preserving contiguous per-feature blocks
- Normalize reordered block transitions so the next block starts from the previous endpoint
- Apply the ordering to pocket, edge-route, V-carve, and recursive V-carve feature-first generation
- Archive the completed plan: planning/archive/FEATURE_FIRST_BLOCK_ORDERING_Plan.md

Closes #120

## Verification
- npx tsx src/engine/toolpaths/toolpaths.test.ts
- npm run build

### DIFF
--- a/planning/archive/FEATURE_FIRST_BLOCK_ORDERING_Plan.md
+++ b/planning/archive/FEATURE_FIRST_BLOCK_ORDERING_Plan.md
@@ -1,0 +1,44 @@
+---
+status: Done
+created: 2026-05-29
+---
+
+# Feature-First Block Ordering Plan
+
+## Goal
+
+Optimize `feature_first` toolpath output so per-feature blocks are still completed one feature at a time, but the concatenated block order follows nearest travel from the previous block endpoint instead of blindly preserving target feature order. This should reduce unnecessary safe-Z rapid travel for pocket, edge-route, V-carve, and V-carve-recursive operations that share the multi-feature merge path.
+
+## Approach
+
+- Add a deterministic nearest-neighbor ordering pass in `multiFeature.ts` for generated per-feature `ToolpathResult` blocks.
+- Use each block's first and last meaningful move positions to score travel from the current endpoint, preserving original order as the tie-breaker.
+- Keep the core `feature_first` behavior intact: all moves from a feature block remain contiguous and are not interleaved with other features.
+- Normalize the concatenated block transition if needed so the first move of a reordered block starts from the previous block endpoint rather than relying on stale `from` coordinates from its isolated generation.
+- Route pocket-specific merging through the same ordered block list while preserving warning, bounds, clamp, and `stepLevels` aggregation behavior.
+
+## Files affected
+
+- `src/engine/toolpaths/multiFeature.ts` — add shared block ordering/transition helpers and apply them in `mergeToolpathResults` / `mergePocketToolpathResults`.
+- `src/engine/toolpaths/toolpaths.test.ts` — add focused tests for non-spatial target order in pocket and edge feature-first operations, and update merge expectations where ordered output changes.
+
+## Tests
+
+- Add or update structural tests in `src/engine/toolpaths/toolpaths.test.ts` to verify:
+  - `feature_first` still keeps each feature's full depth/pass block contiguous.
+  - non-spatial target feature order is reordered by nearest block travel.
+  - deterministic tie-breaking falls back to original target order when distances match.
+  - pocket `stepLevels` aggregation remains deduped and sorted after block ordering.
+- Run `npm run build` before completing the work.
+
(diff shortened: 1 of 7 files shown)


zabooma:
## Code Review

### Overview
Adds an optional `orderBlocks: 'nearest'` to `mergeToolpathResults` / `mergePocketToolpathResults`. In `feature_first` mode for pocket, edge-route, V-carve, and V-carve recursive, per-feature blocks are reordered greedily (nearest-neighbor by XY of block start) instead of following user-specified target order. Inter-block transitions are normalized so the first move of a reordered block starts from the actual previous endpoint.

### Strengths
- Scope is minimal — default behavior unchanged; only feature-first call sites opt in.
- Deterministic tie-break via original index; anchor block is always the first input.
- Single helper threaded through all four operation types via the shared merge function.
- Warnings/bounds/clampIds continue to be aggregated from the original `parts`, so dropping empty-move parts in `orderPartsByNearestBlock` is safe.
- Tests cover ordering, tie-breaking, and `stepLevels` invariants.

### Concerns / Suggestions

**1. Normalized rapid may not retract Z first (`multiFeature.ts` `mergeMoves`)**
When the rewritten transition is a rapid from `previousEnd` (typically at cut depth) to a point at safe-Z, it becomes a single diagonal XYZ rapid. Most controllers execute XYZ rapids as dogleg or fully parallel motion — there's a non-zero risk of grazing material if XY outpaces Z retract. The safer CAM convention is: rapid Z to safe-Z first, then XY traverse, then plunge.
- Not strictly a regression — pre-PR `flatMap` already concatenated blocks with stale `from` — but since this code path now explicitly stitches transitions, it's a good place to enforce safe-Z retract before XY travel.
- Suggest: when `previousEnd.z < safeZ` and the next block's start XY differs, emit two rapids (`previousEnd → (prev.x, prev.y, safeZ)` then `→ next.from`). Plumbing safe-Z through is non-trivial — at minimum a code comment acknowledging the assumption would help.

**2. Dead fallback in `mergeMoves`**
`previousEnd = part.moves.at(-1)?.to ?? previousEnd` — the `?? previousEnd` is unreachable because `continue` above guarantees `part.moves.length > 0`. Can simplify to `part.moves.at(-1)!.to`.

**3. Tie-break test could be stronger**
The tie-break test confirms original-index ordering with one input arrangement. Consider also asserting that swapping the input order swaps the result, to lock in the \"original target order as tie-breaker\" semantics from the plan.

**4. `samePoint` Z-strictness vs XY-only distance scoring**
`samePoint` requires Z equality, but `xyDistanceSquared` doesn't consider Z. Two blocks at identical XY but different Z will still trigger transition normalization, which is correct — worth a brief comment in the helper.

### Risks
- Observable behavior change: feature_first output for multi-feature pocket/edge/V-carve operations now produces different G-code paths than before. Any golden snapshots or recorded G-code comparisons will see diffs. The change is opt-in at the call site, but unconditionally on for all four operation types in this PR.

### Verdict
Approve with the safe-Z transition concern (#1) flagged. Core logic is clean and correct; test coverage matches the plan; legacy `orderBlocks: 'input'` semantics preserved. The Z-retract question is worth at least a comment if not a fix before merge.

PR #126, opened 2026-05-30
Snap waterline ring entries across levels

### LINKED ISSUE #121: Snap waterline ring entries across levels to preserve direct plunges
## Finding

Waterline finish already clusters rings into columns and tries to machine each column top-to-bottom, but the code has a TODO noting that closed rings at adjacent Z levels can miss exact XY alignment because Clipper offset/simplification shifts the nearest entry vertex slightly. When that happens, `transitionToCutEntry()` sees different XY and emits a retract+rapid+plunge instead of the intended direct plunge down the same wall.

The relevant code rotates each closed contour to the nearest entry (`finishSurfaceWaterline.ts:1712-1720`) and then calls `transitionToCutEntry(..., maxLinkDistance = 0)` (`finishSurfaceWaterline.ts:1783`), so even tiny drift is enough to force a safe-Z round trip.

## Suggested direction

- When a closed ring in the same waterline column starts within a small tolerance of the previous ring endpoint, snap/insert the entry point so it exactly matches the previous XY before calling `transitionToCutEntry()`.
- Use a tolerance tied to stepover/tool radius, as the inline TODO suggests, with a conservative cap to avoid distorting real geometry.
- Add a regression test for a vertical-wall pocket or cylinder with multiple waterline levels that asserts column descent emits direct plunges and not retract+rapid+plunge cycles between adjacent rings.

## References

- `src/engine/toolpaths/finishSurfaceWaterline.ts:1582-1585`
- `src/engine/toolpaths/finishSurfaceWaterline.ts:1712-1720`
- `src/engine/toolpaths/finishSurfaceWaterline.ts:1783`
- Shared same-XY plunge behavior: `src/engine/toolpaths/pocket.ts:236-245`

zabooma:
## Summary
- Add bounded snapping for closed waterline ring entries so adjacent same-column levels can keep exact same-XY direct plunges
- Snap the rotated entry vertex in place when it is within tolerance, avoiding an extra short entry segment
- Add regression coverage for drifted closed-ring entry snapping and same-column waterline descents that must not round-trip through safe Z
- Archive the completed plan: planning/archive/WATERLINE_ENTRY_SNAP_Plan.md

Closes #121

## Verification
- npx tsx src/engine/toolpaths/finishSurface.test.ts
- npm run build

### DIFF
--- a/src/engine/toolpaths/finishSurface.test.ts
+++ b/src/engine/toolpaths/finishSurface.test.ts
@@ -10,6 +10,7 @@ import { defaultTool, newProject, rectProfile, type Operation, type Project, typ
 import { normalizeProject, useProjectStore } from '../../store/projectStore'
 import { convertProjectUnits } from '../../utils/units'
 import { generateFinishSurfaceToolpath, maxContourGap } from './finishSurface'
+import { snapClosedContourEntryToAnchor } from './finishSurfaceWaterline'
 import { generateRoughSurfaceToolpath } from './roughSurface'
 import { toClipperPath, normalizeWinding, DEFAULT_CLIPPER_SCALE, applyContourDirectionBySide, isClockwise, normalizeToolForProject } from './geometry'
 import { loadSTLTransformedGeometry } from '../csg'
@@ -1782,6 +1783,27 @@ function testWaterlineConventionalWindsOppositeOfClimb(): void {
   }
 }
 
+function testWaterlineSnapsDriftedClosedRingEntryToPreviousEndpoint(): void {
+  console.log('Testing waterline snaps drifted closed-ring entries to previous XY...')
+  const previousEndpoint = { x: 6.5, y: 3.5 }
+  const contour: Point[] = [
+    { x: 6.503, y: 3.498 },
+    { x: 13.5, y: 3.5 },
+    { x: 13.5, y: 6.5 },
+    { x: 6.5, y: 6.5 },
+  ]
+
+  const snapped = snapClosedContourEntryToAnchor(contour, previousEndpoint, 0.01)
+  assert(snapped.length === contour.length + 1,
+    `expected snap to insert an exact entry point, got ${snapped.length} points`)
+  assert(snapped[0].x === previousEndpoint.x && snapped[0].y === previousEndpoint.y,
+    `expected snapped entry to equal previous endpoint, got (${snapped[0].x}, ${snapped[0].y})`)
+  assert(snapped[1] === contour[0], 'expected original contour geometry to be preserved after inserted entry')
+
+  const unsnapped = snapClosedContourEntryToAnchor(contour, previousEndpoint, 0.001)
+  assert(unsnapped === contour, 'expected contour outside tolerance to remain unchanged')
+}
+
 // ── Run all tests ────────────────────────────────────────────────────────
 
 testFinishSurfaceCoversLowerTopPlateau()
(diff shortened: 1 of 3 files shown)


src/engine/toolpaths/finishSurface.test.ts
```diff
@@ -1829,6 +1851,7 @@ testApplyContourDirectionBySideReversesOpenPolylinesWhenNeeded()
 testApplyContourDirectionBySideAcceptsNaturalWindingHint()
 testWaterlineClimbWindsCorrectlyOnBothSides()
 testWaterlineConventionalWindsOppositeOfClimb()
+testWaterlineSnapsDriftedClosedRingEntryToPreviousEndpoint()
```
zabooma, line 1854: This only covers `snapClosedContourEntryToAnchor` in isolation. The plan called for an end-to-end waterline test that builds a vertical-wall pocket / cylinder mesh, runs `generateFinishSurfaceWaterline`, and asserts adjacent same-column closed rings descend with a direct plunge (no safe-Z retract → rapid → plunge between them). Without that we are not verifying the user-visible fix — only that the helper inserts a point when called directly. The integration is small (apply `entrySnapTolerance`, call `snapClosedContourEntryToAnchor` post-rotation) but it is also where it can quietly break: e.g. `entrySnapTolerance` could be computed as 0 for some unit/stepover combo, the rotated nearest vertex could end up just outside the tolerance, or `transitionToCutEntry`'s same-XY shortcut could regress. Worth adding the regression the plan describes.

src/engine/toolpaths/finishSurfaceWaterline.ts
```diff
@@ -1033,6 +1033,27 @@ function projectedContourStartPoint(
...
+  const first = contour[0]
+  const distance = Math.hypot(first.x - anchor.x, first.y - anchor.y)
+  if (distance <= 1e-9) {
+    return contour
+  }
+  if (distance > tolerance) {
+    return contour
+  }
+
+  return [{ x: anchor.x, y: anchor.y }, ...contour]
```
zabooma: Insert vs. replace: this prepends `anchor` to the contour rather than moving the first vertex to `anchor`, so the emitted ring becomes `anchor → A → B → C → D → anchor` instead of `A → B → C → D → A`. The leading `anchor → A` is a sub-tolerance cut move (≤0.25 mm), and the closing leg goes `D → anchor` instead of `D → A`. For waterline this is harmless geometrically, but it does mean every snapped ring picks up one extra short cut segment at the start and the implicit closure no longer lands on a real contour vertex. If the intent was strictly "preserve original geometry," replacing `contour[0]` with `anchor` (or, equivalently, snapping `A` outright when within tolerance) would be closer to that intent and avoids the tiny entry stub. Worth confirming the insert-then-traverse form is what you actually want here, not just what avoids touching the original points array.

zabooma (commented):
Helper and its tolerance look sound. Two things worth a look before merging.

src/engine/toolpaths/finishSurface.test.ts
zabooma: Addressed in b8fa5f0. I strengthened the generator-level `testWaterlineColumnDescentReusesSameXYWithPlunge` case so it now runs the vertical-wall pocket-block waterline operation and asserts there are zero same-column safe-Z retract/rapid/plunge round trips between adjacent descending rings, in addition to requiring direct cut-Z-to-cut-Z plunges.

Re-ran:
- `npx tsx src/engine/toolpaths/finishSurface.test.ts`
- `npm run build`

src/engine/toolpaths/finishSurfaceWaterline.ts
zabooma: Addressed in b8fa5f0. The helper now snaps by replacing the rotated entry vertex with the anchor (`[anchor, ...contour.slice(1)]`) instead of prepending an extra point, so there is no extra entry stub and the closed-ring traversal stays the same length. The isolated helper test was updated to assert replacement semantics.

Re-ran:
- `npx tsx src/engine/toolpaths/finishSurface.test.ts`
- `npm run build`

PR #130, opened 2026-06-02
Improve adaptive waterline tip caps

### LINKED ISSUE #127: Waterline adaptive: tip-cap quality is poor on island peaks (cone tip, nose/chin/forehead)
## Summary

The adaptive waterline finish strategy produces visibly poor surface quality at the tops of islands — anywhere the projected tip-cap path is the dominant finishing motion. On a simple cone the apex shows concentric ridges / a pock-marked ring pattern; on the OldMan bust the nose tip, chin, and forehead show the same artifact (rough, terraced, and slightly bumpy crowns instead of a smooth blend into the surrounding flank).

Screenshots from the simulator (to attach to this issue):
- Cone — top of the cone (apex). Concentric ring banding clearly visible around the peak.
- OldMan — closeups of the nose tip / forehead / chin area showing terraced, scaly texture on convex tips.

These are exactly the regions handled by `processTipPath` in the adaptive waterline pipeline, not the band fills. The band fills (which cover most of the surface) look acceptable. The defect is local to island peaks.

## Where this lives

- Strategy implementation: [src/engine/toolpaths/finishSurfaceWaterline.ts](src/engine/toolpaths/finishSurfaceWaterline.ts)
  - `processTipPath` — [src/engine/toolpaths/finishSurfaceWaterline.ts:596](src/engine/toolpaths/finishSurfaceWaterline.ts:596)
  - Local-top detection / island-birth handling — [src/engine/toolpaths/finishSurfaceWaterline.ts:694](src/engine/toolpaths/finishSurfaceWaterline.ts:694)
  - Topmost-level tip handling (no higher parent) — [src/engine/toolpaths/finishSurfaceWaterline.ts:820](src/engine/toolpaths/finishSurfaceWaterline.ts:820) area
  - `safeToolTipZAt` heightmap sampling used by tip caps — [src/engine/toolpaths/finishSurfaceWaterline.ts:1081](src/engine/toolpaths/finishSurfaceWaterline.ts:1081)
  - Round-join inset helper `offsetClipperPathsRound` — top of file
- Plan / design notes (this is a known follow-up area from the original adaptive plan): [planning/WATERLINE_ADAPTIVE_REFINEMENT_Plan.md](planning/WATERLINE_ADAPTIVE_REFINEMENT_Plan.md) — see the section *Per-vertex Z source: heightmap (tip caps) vs. linear (bands)* which explicitly calls out tip caps as the place where heightmap sampling is used and where convex plateau / quantization tradeoffs live.

## What the code does today

For each ring identified as an island top (CCW outer ring, either born between two coarse levels or has no higher parent + significantly larger matched lower):

1. The boxy first-slice coarse ring at `tipZ` is suppressed (`suppressCoarsePath`) so the rounder inset cap rings can replace it.
2. `processTipPath` generates inward offsets of `tipPath` using `offsetClipperPathsRound` with round joins and tight `ArcTolerance`, stepping inward by `stepoverDistance` up to `maxRingsPerBand` rings (or until collapse).
3. Per-vertex Z for each ring vertex is sampled via the shared heightmap: `surfaceZAt(point)` → which goes through `safeToolTipZAt(x, y, heightMap, tool)` (the kinematic-safe tool-tip Z, not raw geometric surface Z). The sampled Z is clamped into `[tipZ, peakZ]` where `peakZ = higher.z` (the next coarse level above the tip's first slice).
4. Falls back to a linear ramp from `tipZ` at the ring boundary to `peakZ` at the deepest inset when the heightmap can't resolve a point.
5. Each inset ring is emitted as its own `emitLevel({ projectZAtPoint, source: 'projectedCap' })` so columns walk them top-down.

## Why the result still looks bad

Several plausible contributing causes — likely a combination, not a single root cause:

### 1. Heightmap cell quantization dominates a small footprint
Tip caps span a small XY area (one island top). `safeToolTipZAt` reads the heightmap, which has a finite cell size (`chooseHeightMapCellSize`). On a small cap, each cell occupies a meaningful fraction of the footprint, so cell-level Z steps produce the visible ring terracing on the cone apex. The cone is rotationally symmetric, so cell-grid quantization aliasing shows up as banded concentric ridges instead of a smooth dome. The plan already calls this out as a known tradeoff for tip caps — the cone visualizes it cleanly.

### 2. `safeToolTipZAt` is a discrete neighbor scan
It's the kinematic-safe (tool-tip drop) Z, which involves checking the tool sphere against discrete heightmap neighbors. On steep convex tips that's an additional source of staircase texture: tiny lateral motion changes which neighbor cell wins. The plan notes that switching band fills to the same source produced "bumpy texture wherever bands cover the surface" — the same mechanism is still active inside tip caps; we just accepted it there.

### 3. Z is clamped to `[tipZ, peakZ]` but `peakZ = higher.z` is too low
`peakZ` is set from the next coarse level *above* the tip's first-slice z (`higher.z`). The actual model peak sits in `(upper.z, higher.z]` — but on a sharp cone or a sharp nose, the true peak is at or near the upper end of that interval. Clamping at `higher.z` means rings near the center can't ever reach the true peak; they're capped at the previous coarse z. Combined with discrete heightmap samples, this can produce a flat plateau or ring at `higher.z` around the very center. (When the tip is at the topmost level there's no higher level at all; the comment notes this and reuses `upper.z`, which makes the cap effectively flat in that special case.)

### 4. First-slice ring is miter-derived and never re-derived from the mesh
The plan explicitly notes: *"The outermost cap ring still inherits the miter shape of the upstream `buildWaterlineLevels` slice offset; a follow-up could re-derive that first ring from the raw mesh slice (we have it in `sliceMaterialByZ`) with round joins."* Round joins fix subsequent insets, but the outermost cap ring is still as polygonal as the first slice that produced it. On a cone this shows as a coarse outer polygon that the inner rings shrink off of, contributing to the visible ring structure.

### 5. Each ring is emitted as a single column level — no smoothing between rings
`emitLevel(..., source: 'projectedCap')` makes every inset its own level. The column walker descends them top-down, so between ring `n` and ring `n+1` the tool retracts/plunges (or steps) at discrete `ringZ` values that are computed from a linear `tipZ + (distance/denom)*zSpan` estimate. Even though per-vertex Z follows the heightmap inside a ring, the *cluster representative z* used for ordering is linear. If the heightmap returns Z values that disagree noticeably with the linear estimate, the descent order can create visible scallops between rings as well as within them.

### 6. Tip-cap ring spacing equals the band stepover
`stepoverDistance` for tip caps is the same as for band fills. On a small convex peak, the rings near the center collapse quickly and the last few rings end up at almost-but-not-quite-the-same Z, which is visually a ring stack rather than a smooth crown.

## Repro

Both fixtures live in `work/`:

- **Cone** — `work/Cone.camj` (uses `work/cone.stl`). Open it, run the waterline finish operation, and render the simulation. The apex shows concentric ring banding (matches the first attached screenshot).
- **OldMan** — `work/old-man-in-box.camj` (already referenced in the adaptive plan; `work/Old-man-simple.camj` and `work/Oldman-splash-final.camj` are also available if a simpler setup is useful). Same waterline finish setup. Render the simulation and zoom on the tip of the nose, the chin, and the forehead. Each shows terraced / bumpy texture localized to the convex peak.

The band-fill surface in between these peaks looks fine; this is specifically a tip-cap quality regression area.

## Suggested directions for whoever picks this up

In rough order of likely impact (none of these are mandates — pick what's actually needed after measuring):

1. **Re-derive the outermost tip-cap ring directly from the mesh slice** using `sliceMaterialByZ` at `tipZ` with round joins instead of inheriting the miter-derived `tipPath`. This removes the visible polygonal outer ring and gives all subsequent insets a smooth starting boundary. The plan already flags this as a follow-up.
2. **Refine the heightmap cell size locally for tip-cap sampling** (or sample with bilinear/biquadratic interpolation between cells rather than nearest-cell). The cone case is essentially an aliasing problem — finer or interpolated sampling should kill the ring banding.
3. **Replace the per-ring `ringZ` representative with one derived from the actual sampled heightmap profile** along the ring (e.g., the mean of `projectZAtPoint` evaluated around the ring boundary), so column ordering matches what's actually being cut.
4. **Raise `peakZ` above `higher.z` for sharp/small tips** — sample the heightmap inside the cap footprint to estimate the true peak Z and use that as the upper clamp instead of the next coarse level. (Carefully — must not exceed stock top, and must respect tool-tip-safe Z if shape-aware safety is required.)
5. **Densify rings near the collapse point** — once the inset shrinks below some fraction of stepover, switch to a finer ring spacing or emit a centered single point so the crown closes cleanly instead of stacking near-identical rings.
6. **Selectively apply the heightmap correction only where it disagrees with the linear ramp by more than a threshold** (the plan's documented future-refinement idea), and otherwise use the smooth linear projection. This would target the convex-plateau case without re-introducing all-over cell quantization noise.

## Out of scope

- Band-fill Z source (intentionally linear per the plan — see notes there).
- Intersecting-add wall cleanup (separate follow-up section in the same plan).
- Renaming or restructuring the operation.
- A true 3D constant-scallop algorithm.

## Acceptance

- Cone apex renders as a smooth dome in the simulator at typical stepovers (no concentric ring banding visible on the peak).
- OldMan nose tip, chin, and forehead render without the terraced/scaly texture currently seen on those convex peaks.
- Band-fill surface quality elsewhere is unchanged or improved (no regressions in the band-fill path).
- Existing waterline regression tests in [src/engine/toolpaths/finishSurface.test.ts](src/engine/toolpaths/finishSurface.test.ts) still pass; ideally add a tip-cap quality regression case (e.g., asserting Z monotonicity of projected cap rings on a synthetic cone, or asserting no rings cluster within an epsilon of each other in Z).

Image

Image

zabooma:
## Summary

- Generate adaptive waterline tip caps from the first real stepdown using micro-Z slices, including split/collapsed tip branches without cone-specific logic
- Emit each waterline column's real boundary rings before projected fill rings, and allow bounded intra-column cut links to reduce nearby retract/plunge round trips
- Add `waterlineTipStepdown` operation plumbing plus synthetic cone/collapsed-branch/order/linking regressions
- Archive the completed plan: `planning/archive/WATERLINE_TIP_CAP_QUALITY_Plan.md`

Closes #127

## Verification

- `npx tsx src/engine/toolpaths/finishSurface.test.ts`
- `npm run build`

### DIFF
--- a/planning/archive/WATERLINE_TIP_CAP_QUALITY_Plan.md
+++ b/planning/archive/WATERLINE_TIP_CAP_QUALITY_Plan.md
@@ -1,0 +1,57 @@
+---
+status: Done
+created: 2026-05-30
+---
+
+# Waterline Tip-Cap Quality Plan
+
+## Goal
+
+Fix issue #127: adaptive waterline tip caps should render smooth convex island peaks instead of producing concentric banding, terraced crowns, or bumpy cap texture. The user-visible target is a smoother cone apex and smoother OldMan nose/chin/forehead tips while preserving the existing band-fill quality elsewhere.
+
+## Approach
+
+- Keep the existing adaptive waterline strategy and focus only on `projectedCap` generation inside `generateProjectedWaterlineLevels`.
+- Add a small synthetic cone-style regression fixture/test in `finishSurface.test.ts` so the failure can be measured without relying on large `work/` files or screenshot inspection.
+- Improve tip-cap Z projection so caps do not blindly expose heightmap cell quantization:
+  - use the first generated stepdown shape as the base for every local top instead of special-casing cone-like topology
+  - create real intermediate "tip micro-step" waterline levels upward from that first stepdown, defaulting to half the operation stepdown
+  - stop creating tip micro-step levels when the sliced/offset shape collapses into air above the island
+  - feed those real intermediate contours through the same projected-band interpolation used between regular waterline levels
+- Handle split/merge tip stacks generically: if one active micro-level branch disappears while another island continues upward, emit inward projected cap rings for the disappearing branch instead of dropping it.
+- Keep generated toolpaths readable in the preview by emitting each column's real waterline boundaries before the generated projected fill rings.
+- Reduce safe-Z round trips inside each waterline column by allowing short direct cut links between successive nearby rings, bounded by the adaptive stepover and disabled across protected clipped breaks.
+- If the first-slice boundary remains visibly polygonal after the Z fix, re-derive or smooth the tip cap's starting boundary from the stored mesh slice material using round offset joins for the specific matched island.
+- Preserve current protections around intersecting add features, containing pockets, tabs, clamps, region filtering, and adaptive band fills.
+
+## Files affected
+
+- `src/engine/toolpaths/finishSurfaceWaterline.ts` — adjust projected-cap generation, Z projection, peak estimation, and possibly the round-join starting boundary used for island tips.
+- `src/engine/toolpaths/finishSurface.test.ts` — add or update waterline adaptive tip-cap regression coverage around a synthetic cone/peak model, plus assertions that band-fill behavior is not regressed.
+- `src/engine/toolpaths/INDEX.md` — update only if the waterline entry needs a more precise description after the implementation.
+- `planning/WATERLINE_TIP_CAP_QUALITY_Plan.md` — archive when complete.
+
+## Tests
+
+- Add a focused waterline regression test that generates projected cap cuts on a synthetic convex peak and asserts:
+  - projected cap Z values are monotonic toward the peak within the cap
+  - cap rings do not cluster into near-identical Z shelves at the crown
+  - the highest projected cap reaches near the model/tool-safe peak instead of staying flat at the coarse top level
+- Add a second synthetic regression where overlapping uneven peaks split at micro-Z levels and the lower branch collapses before the taller peak; assert the lower branch still receives projected cap cuts.
+- Assert the cone fixture emits the real waterline boundary before projected cap fill paths so the output does not begin with interior fill rings.
+- Assert nearby projected cap fill starts without an immediate safe-Z plunge when the preceding ring is close enough to link.
+- Run `npx tsx src/engine/toolpaths/finishSurface.test.ts` during iteration.
+- Run `npm run build` before committing.
+
+## Open questions / risks
+
+- Simulator visual quality is the real acceptance criterion, but automated tests can only approximate it with Z-distribution and monotonicity assertions. I will keep the regression test focused on the failure mechanics described in issue #127.
+- Using heightmap samples too aggressively caused visible texture across band fills before; this work must stay local to `projectedCap` and avoid changing `projectedBand` Z behavior.
+- Local heightmap refinement can become expensive if it rebuilds global grids. Any higher-resolution or interpolation work should be bounded to cap sampling or reuse existing data.
+
+## Out of scope
+
+- Replacing adaptive waterline with true 3D constant-scallop finishing.
+- Changing parallel finish or rough-surface strategies.
+- Redesigning adaptive band fills, region behavior, intersecting-add wall cleanup, or waterline UI controls.
+- Starting the Vite dev server.
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -95,6 +95,7 @@ function operationComputationEquals(a: Operation, b: Operation): boolean {
     && a.waterlineMicroStepover === b.waterlineMicroStepover
     && a.waterlineRefinementThreshold === b.waterlineRefinementThreshold
     && a.waterlineMaxRingsPerBand === b.waterlineMaxRingsPerBand
+    && a.waterlineTipStepdown === b.waterlineTipStepdown
   )
 }
 
--- a/src/engine/toolpaths/finishSurface.test.ts
+++ b/src/engine/toolpaths/finishSurface.test.ts
@@ -90,6 +90,21 @@ function addQuad(
   lines.push('  endfacet')
 }
 
+function addTriangle(
+  lines: string[],
+  a: [number, number, number],
+  b: [number, number, number],
+  c: [number, number, number],
+): void {
+  lines.push('  facet normal 0 0 0')
+  lines.push('    outer loop')
+  for (const point of [a, b, c]) {
+    lines.push(`      vertex ${point.join(' ')}`)
+  }
+  lines.push('    endloop')
+  lines.push('  endfacet')
+}
+
 function makeSteppedStlDataUrl(): string {
   const lines = ['solid stepped']
   addBoxFaces(lines, 0, 0, 0, 20, 10, 1)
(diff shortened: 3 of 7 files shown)


zabooma:
## Code Review

### Overview
Replaces the previous heightmap-sampled "shrink the boxy first-slice ring inward with round joins" tip-cap algorithm with a structurally simpler approach: start each local top from the first real stepdown shape, walk back upward in real sliced micro-Z slices (default `stepdown/2`, configurable via new `waterlineTipStepdown`), and feed each step through the same band-fill projection used between coarse levels. Splits/collapses are handled generically — when a branch loses its matched upper contour, an inward projected cap is emitted instead of dropping it. The PR also reorders columns (real waterlines top-down, projected fills bottom-up) and allows bounded intra-column linking to skip retract+plunge round trips.

### Strengths
- **Significant simplification**: removes `offsetClipperPathsRound`, the heightmap-rasterize helper, and the special-cased cone-tip ramp; reuses one `emitProjectedBandFill` for both bands and caps.
- **Real-slice-based Z** removes the heightmap-cell quantization that was previously leaking into cap texture — addresses the root cause of #127 rather than masking it.
- **Solid synthetic regression coverage**: cone fixture asserts Z range, monotonic rise toward center, ordering of real waterlines before projected fills, link-without-plunge, no coarse ring inside the crown. Uneven-twin-cone asserts the collapsed lower branch still receives projected cap cuts.
- **Operation field is fully plumbed**: types, defaults, normalize, equality check (`App.tsx`), and `convertOperation` unit handling.

### Concerns

**1. Possible plunge-safety regression near tall intersecting adds.** The PR deletes `heightMapWithIntersectingAddTops`, which used to stamp each intersecting-add's `topZ` into the heightmap so `safeToolTipZAt` would treat (e.g.) the `old-man-in-box` wedge as an obstacle, not just the mesh underneath. Cap Z is now derived from real slices, so cap projection itself is fine, but downstream code that calls `safeToolTipZAt(point.x, point.y, baseHeightMap, tool)` now sees the mesh-only heightmap and can return an unsafe tip Z near a tall add. The diff comment claims the heightmap is kept "for intersecting-add plunge/split safety later," but the helper that actually made that safety work for *intersecting adds taller than the mesh* is gone. Suggest verifying the `old-man-in-box.camj` waterline doesn't regress, or restoring the rasterize helper just for the safety queries.

**2. Collapsed-branch emission depends on `maxContourGap([], [lowerPath])`.** When the tip stack walks past the lower peak in the twin-cone test, `processTipStack` calls `emitProjectedBandFill` with `matchedUpperPaths = []` and `upper.contourPaths = []`. The early return is `if (gap <= minimumGap) return true`. If `maxContourGap` returns 0 when one side is empty (a reasonable implementation), the collapsed branch emits nothing and the assertion on inner cuts of the lower peak would still pass via earlier non-collapsed micro-levels. Worth confirming the collapsed-only path actually emits something — add an assertion checking the *highest* Z inner cut on the lower peak is closer to its 3 mm apex than to where it last had a matched upper, otherwise the "collapsed-branch fill" claim is not really tested.

**3. Plan archived in the same PR.** `planning/archive/WATERLINE_TIP_CAP_QUALITY_Plan.md` is added in this PR while #127 is still being closed. Plans should stay in `planning/` (status `In progress`) until the user has tested in the simulator. Move the archive to a follow-up commit after sim verification.

**4. Transition-move source tagging may mislabel travel as cut.** After `transitionToCutEntry`, the inserted move's `source` is overwritten with `ringEntry.source` when the move is `kind === 'cut'`. A linked transition between two cap rings will be labeled `projectedCap` even when it's a horizontal travel rather than a true cap cut. Likely cosmetic (affects preview color/source filtering), but flag if `source` is consumed by post-processing or G-code emission rules.

**5. Cluster ordering change is a real behavior shift.** Real waterlines descend top-down, projected fills now climb bottom-up within the same column. This is intentional and benefits link reuse, but it changes the order in which cap rings hit the simulator — worth a brief mention in the operation docs / commit message if other agents inspect emitted move order.

**6. Minor.** `columnLinkDistance = min(stepoverDistance * 1.5, tool.radius)` is reasonable, but capping at the *full* tool radius can still translate to a noticeable horizontal cut between rings on small tools. Consider `tool.radius * 0.5` if you see scoring between adjacent cap rings in the sim.

### Risk Summary
- **Highest risk:** plunge safety near tall intersecting adds (concern #1) — direct regression vector for `old-man-in-box.camj`.
- **Medium risk:** collapsed-branch test may not actually exercise the empty-upper code path (concern #2).
- **Low risk:** premature archive, source tagging, ordering — cosmetic / process.

Recommend a sim pass on `old-man-in-box.camj` and `cone` before merging, and tightening the collapsed-branch assertion to guarantee the empty-upper path emits rings.

zabooma:
Follow-up pushed in `91e3648` addressing the actionable review items I could verify locally:

- Restored an add-aware safety heightmap for waterline safety queries. Intersecting add `topZ` is stamped into a copied heightmap and used by `safeToolTipZAt` in both lifted-Z and contour-splitting paths, without feeding add geometry into the projected slice generation.
- Tightened `testWaterlineTipCapFillsCollapsedBranch` so it now asserts the collapsed lower branch reaches near its local apex, not just that some projected cap cuts exist.
- Reduced the nearby-ring direct-link cap from `tool.radius` to `tool.radius * 0.5` to keep the plunge/retraction optimization more conservative.

Verification:
- `npx tsx src/engine/toolpaths/finishSurface.test.ts`
- `npm run build`
- Loaded `work/Oldman-splash-final.camj` in Chrome at `localhost:1421`, generated/displayed the 3D Surface finish toolpath, and saw no console errors. Existing warnings were localhost analytics disabled and non-manifold STL fallback.

I left the plan archive state unchanged because the repo instructions say to archive before opening a PR. This is not a declaration that the visual issue is validated; Frank still needs to confirm the actual machining preview behavior.

PR #215, opened 2026-06-26
feat: emit canned drilling cycles (G81/G82/G83/G73) from post-processor

### LINKED ISSUE #172: Canned drilling cycles (G81/G82/G83/G73) not emitted
`MachineDefinition.cannedCycles` schema exists but `runPostProcessor` never emits canned cycles; all drill moves post as expanded G0/G1 (drill-type differentiation lives only at toolpath generation). Drilling works, just without canned-cycle G-code.

Pre-existing. Documented by the Phase-3 audit — see the NOTE in `src/engine/toolpaths/camOperationSmoke.test.ts` (~line 454).

Migrated from `planning/INDEX.md` tech-debt.

zabooma:
Closes #172

## What

Wires `MachineDefinition.cannedCycles` into `runPostProcessor` so drilling operations export canned-cycle G-code (`G81`/`G82`/`G83`/`G73` … `G80`) on machines that support them, instead of always expanding to `G0/G1`.

## How

The drilling generator already computes every cycle parameter but flattens each hole into `rapid`/`plunge` moves. Rather than reverse-engineer cycles in the post-processor, the generator now attaches structured `DrillCycle` metadata to its result **alongside the existing moves**. The post-processor reads that metadata and emits modal canned-cycle blocks when the active machine definition supports the cycle; otherwise it falls back to today's expanded-move path.

- `cannedCycles` schema gains `chipBreakDrillCommand` (G73) and `cancelCommand` (G80); added to the three canned-capable defs (uccnc/linuxcnc/mach3). grbl/grblhal/generic keep `cannedCycles: null`.
- New `DrillCycle` type + optional `drillCycles?` on `ToolpathResult`; populated in `drilling.ts` in lockstep with moves.
- Post-processor emits `<cmd> X Y Z<bottom> R<retract> [Q<peck>] [P<dwell>] F<plungeFeed>` (modal), preceded by the `G98`/`G99` retract mode and terminated with `G80`.

## Invariants preserved

- **Toolpath `moves` are unchanged** — the 3D simulation/preview consume `toolpath.moves` only and are unaffected; the canned metadata is post-processor-only.
- Machines with `cannedCycles: null` (GRBL etc.) still export expanded `G0/G1` with no canned words.
- Any drill type whose command is null (e.g. `chip_breaking` on a def without G73) falls back to expanded moves and pushes a warning.

## Tests

`postprocessor.test.ts`: G81/G82/G83/G73 emission (with Z/R/Q/P/F + closing G80) on a canned-capable def, plus a GRBL regression guard asserting no canned words. Smoke-test NOTE repointed to reflect implemented behavior.

## Verify

`npm run build` (tsc + full test suite + vite) green.

## Notes for reviewers

Implemented via the project's Claude CLI/DeepSeek worker in an isolated worktree; diff, commit, and build were reviewed and re-run independently before this PR.

### DIFF
--- a/src/engine/gcode/definitions/linuxcnc.json
+++ b/src/engine/gcode/definitions/linuxcnc.json
@@ -70,8 +70,10 @@
     "drillCommand": "G81",
     "drillWithDwellCommand": "G82",
     "peckDrillCommand": "G83",
+    "chipBreakDrillCommand": "G73",
     "peckStepWord": "Q",
-    "retractMode": "G98"
+    "retractMode": "G98",
+    "cancelCommand": "G80"
   },
   "coolant": {
     "floodOnCommand": "M8",
--- a/src/engine/gcode/definitions/mach3.json
+++ b/src/engine/gcode/definitions/mach3.json
@@ -70,8 +70,10 @@
     "drillCommand": "G81",
     "drillWithDwellCommand": "G82",
     "peckDrillCommand": "G83",
+    "chipBreakDrillCommand": "G73",
     "peckStepWord": "Q",
-    "retractMode": "G98"
+    "retractMode": "G98",
+    "cancelCommand": "G80"
   },
   "coolant": {
     "floodOnCommand": "M8",
--- a/src/engine/gcode/definitions/uccnc.json
+++ b/src/engine/gcode/definitions/uccnc.json
@@ -74,8 +74,10 @@
     "drillCommand": "G81",
     "drillWithDwellCommand": "G82",
     "peckDrillCommand": "G83",
+    "chipBreakDrillCommand": "G73",
     "peckStepWord": "Q",
-    "retractMode": "G98"
+    "retractMode": "G98",
+    "cancelCommand": "G80"
   },
   "coolant": {
     "floodOnCommand": "M8",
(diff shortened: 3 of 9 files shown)


zabooma:
Addressed the P1 from review (raised via Codex; couldn't be posted directly due to a usage limit) in 23817c4.

**Root cause confirmed and reproduced:** `chipBreakDrillCommand` was added to the `cannedCycles` schema as a required field. Legacy `.camj` projects / custom machine definitions saved before this PR lack the key, so `validateMachineDefinition` threw and `normalizeMachineDefinitions` silently dropped the machine (clearing `selectedMachineId`). Validating the `mach3` and `linuxcnc` definitions embedded in `src/engine/test-fixtures/model-in-pocket.camj` reproduced it.

**Fix:** default the field to null, matching how the sibling `cancelCommand` already handles legacy absence:
```ts
chipBreakDrillCommand: z.string().nullable().default(null),
```
A null `chipBreakDrillCommand` already means "no G73 command", and the post-processor already falls back to expanded moves for that case — so no other change was needed.

**Regression test:** `testLegacyCannedCycleDefaults` in `postprocessor.test.ts` validates a legacy canned-cycle def missing both `chipBreakDrillCommand` and `cancelCommand`, asserting it does not throw and that the two fields default to `null` / `G80`.

**Verified:** the previously-failing fixture defs now validate (`chipBreak=null`); `npm run build` green (tsc + all 75 test files + vite).

PR #246, opened 2026-07-01
feat: reduced slot feed for fully engaged pocket cuts

### LINKED ISSUE #173: Pocket offset: reduced-feed first innermost loop
Follow-up plan for optionally cutting the first innermost offset loop at reduced feed before restoring normal pocket feed. we need to add a new operation property for this, vakue would be a percentage of the operation's feed.

Migrated from `planning/INDEX.md` Pending approval. Original draft preserved at `planning/archive/POCKET_OFFSET_ENTRY_FEED_REDUCTION_Plan.md` (git history).

zabooma:
Implements #173, evolved through review into an engagement-based model.

## What

Adds an optional **Slot Feed (%)** setting for pocket operations: cut moves where the tool is fully engaged (slotting into virgin material) run at `feed × percent/100`; everything else runs at normal feed. Geometry is unchanged — this is feed metadata plus two ordering fixes that the feature surfaced.

## Engagement model

One rule, applied once per Z level to all pocket passes (offset rough, parallel rough, finish floor): **a cut fragment is fully engaged exactly when no kerf already cut at that level lies within ~0.9× the tool diameter of it.** Prior kerfs directly behind the motion direction (the tool's own trail) don't count, so a straight slot stays engaged however long it runs. Prior cut segments are indexed in chunk-sized pieces on a uniform grid; moves are classified in quarter-slot-distance chunks and split where the classification changes.

This yields, with no special cases: each disjoint section's own inner start reduced; the first pass through a pinch corridor reduced (and only that pass — later cuts riding its kerf are normal); links through cleared material, zigzag end links, ring corners, and the self-overlapping back side of thin loops all at normal feed; parallel boundary passes reduced and fill lines adjacent to a kerf normal.

## Ordering fixes (apply regardless of slot feed)

- **Pocket and surface finish cut the floor before the walls** — with walls first, a wall pass at final depth slots through whatever axial stock roughing left, at full feed.
- **Pocket offset finish floors cut inner-first** like the rough pass (per disjoint floor area, outer contours only). Surface-clean roughing intentionally stays outer-first: its coverage boundary hangs the tool off the material edge at light engagement, the opposite of a pocket wall.

## Plumbing

- `ToolpathMove.feedScale` multiplier; postprocessor emits `(feed || default) × feedScale` with the existing modal F-word logic.
- Optional `pocketSlotFeedPercent` field (no migration; `undefined`/100 = off, byte-identical move stream, covered by parity tests). UI: "Slot Feed (%)" for rough pockets and finish pockets with Finish Floor.
- Offset ring trees are built once per band and traversed per level (also removes pre-existing per-level Clipper recomputation).
- Simulation: playback bar shows the current move's real feed (dot + value); move subdivision preserves `feedScale`; the tool animation advances on a reference-feed budget so reduced cuts visibly slow down (seeking stays geometric).

## Testing

9 toolpath tests (simple/island/partial-depth-island/multi-level offset, parallel, offset+parallel finish floor, surface ordering, disabled parity), postprocessor modal-F test, playback tests (metadata preservation, feed-scaled advance, geometric seek), plus probe scripts verifying stamped-move sets and byte-identical disabled parity against the pre-feature stream. `npm run build` green (81 test files).

Closes #173

### DIFF
--- a/src/app/useSimulationModel.ts
+++ b/src/app/useSimulationModel.ts
@@ -191,6 +191,7 @@ export function useSimulationModel({
     // units-per-second, so divide by 60. This becomes the "1×" playback speed so
     // users can intuitively speed up or slow down relative to the real feed rate.
     const feedPerSecond = selectedOperation.feed > 0 ? selectedOperation.feed / 60 : undefined
+    const plungeFeedPerSecond = selectedOperation.plungeFeed > 0 ? selectedOperation.plungeFeed / 60 : undefined
     // `project.meta.units` uses 'inch'; the playback UI shows the short label 'in'.
     const units: 'mm' | 'in' = project.meta.units === 'inch' ? 'in' : 'mm'
 
@@ -205,6 +206,7 @@ export function useSimulationModel({
       maxSegmentLength,
       units,
       feedPerSecond,
+      plungeFeedPerSecond,
     }
   }, [centerTab, generateToolpathForOperation, project, selectedOperation, selectedToolpath, simulationDetailCells, simulationMode])
 
--- a/src/app/useToolpathGeneration.ts
+++ b/src/app/useToolpathGeneration.ts
@@ -73,6 +73,7 @@ export function operationComputationEquals(a: Operation, b: Operation): boolean 
     && a.rpm === b.rpm
     && a.pocketPattern === b.pocketPattern
     && a.pocketAngle === b.pocketAngle
+    && a.pocketSlotFeedPercent === b.pocketSlotFeedPercent
     && a.stockToLeaveRadial === b.stockToLeaveRadial
     && a.stockToLeaveAxial === b.stockToLeaveAxial
     && a.finishWalls === b.finishWalls
--- a/src/components/cam/CAMPanel.tsx
+++ b/src/components/cam/CAMPanel.tsx
@@ -1468,6 +1468,24 @@ export function CAMPanel({
                       onCommit={(value) => updateOperation(selectedOperation.id, { plungeFeed: value })}
                     />
                   </label>
+                  {selectedOperation.kind === 'pocket'
+                    && (selectedOperation.pass === 'rough'
+                      || (selectedOperation.pass === 'finish' && selectedOperation.finishFloor)) ? (
+                    <label
+                      className="properties-field"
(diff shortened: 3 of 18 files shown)


zabooma:
Review findings:

1. [P2] `src/engine/operationBooklet/report.ts:151`-`156` estimates all cut/lead moves at `operation.feed`, but PR 246 now lets those same moves carry `move.feedScale` and the postprocessor emits `(operation.feed || tool.defaultFeed) * (move.feedScale ?? 1)` in `src/engine/gcode/postprocessor.ts:396`-`398`. Any setup booklet for a slot-feed pocket will under-report feed-controlled time, potentially by a large margin when the reduced-feed fragments are long. Please price cut/lead time with the effective scaled feed and add an operation-booklet regression test using a `feedScale < 1` move.

2. [P3] The PR description still includes the generated-tool attribution footer. `AGENTS.md` says PR descriptions should not append `Generated with ...` or similar attribution footers, so please remove the Claude Code footer before merge.

Verification run locally in the PR worktree:
- `npx tsx src/engine/toolpaths/toolpaths.test.ts`
- `npx tsx src/engine/gcode/postprocessor.test.ts`
- `npx tsx src/engine/simulation/playback.test.ts`
- `npm run build`

zabooma:
Both findings addressed:

1. **Booklet feed time** — fixed in 5de9dc2: `feedControlledTimeSeconds` now prices cut/lead moves at `operation.feed × (move.feedScale ?? 1)`, matching what the postprocessor emits. Added `testFeedTimeUsesScaledSlotFeed` (20 mm at 100% + 20 mm at 10% must report 16.5 s, not 3 s) plus a Slot Feed settings-row assertion.

2. **Attribution footer** — removed from the PR description.

`npm run build` green (81 test files).

zabooma:
Re-check result: both review findings are addressed in `5de9dc2`.

Verified:
- `feedControlledTimeSeconds` now prices cut/lead moves with `operation.feed * (move.feedScale ?? 1)`, matching the postprocessor
- `src/engine/operationBooklet/operationBooklet.test.ts` now covers the scaled slot-feed estimate and Slot Feed report row
- PR body no longer includes the generated-tool attribution footer

Validation run locally in the PR worktree:
- `npx tsx src/engine/operationBooklet/operationBooklet.test.ts`
- `npx tsx src/engine/gcode/postprocessor.test.ts`
- `npm run build`

No remaining findings from my previous comments.

PR #249, opened 2026-07-03
Add hole-capable region masks

### LINKED ISSUE #168: Rest regions can't represent holes: stock-to-leave / annular bands over-cover the cleared center
## Context

PR #164 / commit `a86c627` reworked rest-region generation. Rest regions are emitted as `region` features whose `profile` is a **single closed loop** — there is no way to represent a region **with a hole**. To avoid an annular region's hole being emitted as a phantom positive "inside" region, `areaPathsToDrafts` now **drops hole contours**:

`src/engine/toolpaths/restRegions.ts`
- `signedContourArea` (~line 381)
- `areaPathsToDrafts` filters `signedContourArea(contour) > 0` (~line 399) — i.e. keeps only positively-wound outer contours, discarding holes.

This is correct for the common corner-cusp case (no holes), but it mishandles genuinely **annular** rest.

## Symptom

When a pocket is cut with a radial **stock-to-leave** (`operation.stockToLeaveRadial > 0`), the previous tool leaves a continuous band of material around the wall (an annulus), not just corner cusps. The pipeline:

1. `generateAreaRestRegionDrafts` (~line 410) builds corner triangles + a **residual** pass (`removeThinSlivers`, ~line 374) that captures the wall band.
2. The band is an annulus: outer contour = pocket wall, inner contour (hole) = inner edge of the band.
3. `areaPathsToDrafts` drops the hole → emits the **filled outer** contour = the whole pocket interior.

Result: the rest region covers the already-cleared center instead of just the leftover band, so a finishing tool clipped to it re-machines the entire pocket. Wasteful (lots of air/`re`-cutting) — not a gouge, but wrong.

Islands can produce the same annular shape and hit the same limitation.

## Root cause

The region-mask machinery is union-only and single-contour:
- `RestRegionDraft.profile` is one `SketchProfile` (no holes).
- `buildRegionMask` (`src/engine/toolpaths/regions.ts`) does `unionPaths(...)` over region features and `normalizeWinding(..., false)` — so even if a hole were emitted as a separate feature, the winding is normalized and it would be unioned (filled), not subtracted.

## Proposed approaches (pick one)

1. **Hole-capable region masks (preferred, most general).** Allow a region feature to be tagged subtractive; have `buildRegionMask` difference the subtractive ones from the additive ones. Rest emits the band's outer as additive + its hole as subtractive.
2. **Annulus decomposition.** Decompose an annular rest region into hole-free simple polygons (slab/trapezoid decomposition) and emit each as its own draft.
3. **Keyhole/bridge contour.** Connect the hole to the outer with a zero-width bridge so a single contour excludes the hole under non-zero fill — but `buildRegionMask`'s `normalizeWinding` currently defeats this, so it would need adjusting.

## Acceptance criteria

- A rectangular pocket with `stockToLeaveRadial > 0` produces a rest region that **excludes the cleared center** (machines only the perimeter band).
- Same for a pocket with an island that leaves an annular rest ring.
- New test: assert the generated rest region(s) for a stock-to-leave rect pocket do **not** contain the pocket-center point (and do contain a point in the wall band). Compare with the existing `testPocketRestRegionsFindCornerCusps` style in `toolpaths.test.ts`.

## Related

- PR #164 (introduced the hole-drop in `areaPathsToDrafts`).
- A prior task chip for this was raised and dismissed during the PR work; this issue is the durable record.

zabooma:
## Summary
- Add include/exclude region mask mode with legacy region defaults normalized to include
- Compose region masks in feature order so include/exclude nesting can create holes and islands
- Preserve hole-capable pocket rest regions and apply generated mask modes to created rest region features
- Add Properties mask-mode controls, exclude tree badge, dashed canvas styling, and updated legend entries

Closes #168

## Verification
- npx tsx src/store/featureReferencesMigration.test.ts
- npm run build


zabooma:
## Review: exclude region masks aren't honored by the non-pocket operations

The core composition is correct. Pocket, edge, carving, waterline, and drilling clip through the paths this PR migrated to the ordered `containsPoint` / `applyRegionMaskToPaths` logic (e.g. [`pocket.ts:1618`](https://github.com/PureCutCNC/purecutcnc/blob/fd124e10c8b64ee1bf8448ac86bd2112b74a0bf7/src/engine/toolpaths/pocket.ts#L1618) → `clipToolpathResultToRegionMask`, [`resolver.ts:353`](https://github.com/PureCutCNC/purecutcnc/blob/fd124e10c8b64ee1bf8448ac86bd2112b74a0bf7/src/engine/toolpaths/resolver.ts#L353)), so nested include/exclude composition works there — confirmed with an exclude→include→exclude pocket.

The gap is four operations that were **not** migrated: they still clip by intersecting `regionMask.paths` directly, and that field no longer represents the mask once a region is `exclude` (or the first region is exclude, which sets `baseIncludesSubject`).

### 1. `regionMask.paths` intersection inverts exclude-first masks

- [`surface.ts:260`](https://github.com/PureCutCNC/purecutcnc/blob/fd124e10c8b64ee1bf8448ac86bd2112b74a0bf7/src/engine/toolpaths/surface.ts#L260) — `executeClipPaths(subjectPaths, regionMask.paths, ctIntersection)`
- [`surfaceStepdown3d.ts:258`](https://github.com/PureCutCNC/purecutcnc/blob/fd124e10c8b64ee1bf8448ac86bd2112b74a0bf7/src/engine/toolpaths/surfaceStepdown3d.ts#L258) — `intersectClipperPaths(outlinePaths, regionMask.paths)`
- [`restRegions.ts:593-594`](https://github.com/PureCutCNC/purecutcnc/blob/fd124e10c8b64ee1bf8448ac86bd2112b74a0bf7/src/engine/toolpaths/restRegions.ts#L593-L594) — `intersectClipperPaths(sourceBand/reachableBand, regionMask.paths)`

For an exclude-only mask, `buildRegionMask` sets `paths = excludePaths` (the excluded area) with `baseIncludesSubject = true`. Intersecting the subject against that keeps only the **interior of the excluded region** — the inverse of what the user asked. (Whenever the first targeted region is `exclude`, `regionMask.paths` also drops the "base = whole subject" semantics.)

Isolated repro — stock 20×20 (area 400), one `exclude` region 10×10 (area 100):

```
CORRECT (applyRegionMaskToPaths — pocket/edge/resolver):   machined area 300  (everything EXCEPT the excluded region)
LEGACY  (intersect subject, regionMask.paths — surface/rest): machined area 100  (ONLY the excluded region)

point (2,2)  outside exclude, SHOULD machine:  correct = true    surface/rest = false
point (10,10) inside exclude,  should skip:    correct = false   surface/rest = true
```

Fix: route these through `applyRegionMaskToPaths`, the same way `resolver.ts:353` now does. ([`drilling.ts:57`](https://github.com/PureCutCNC/purecutcnc/blob/fd124e10c8b64ee1bf8448ac86bd2112b74a0bf7/src/engine/toolpaths/drilling.ts#L57) is already fine — it uses `regionMask.containsPoint`.)

### 2. Parallel finish surface composes regions independently

[`finishSurfaceParallel.ts:758`](https://github.com/PureCutCNC/purecutcnc/blob/fd124e10c8b64ee1bf8448ac86bd2112b74a0bf7/src/engine/toolpaths/finishSurfaceParallel.ts#L758) builds a mask **per region** in a loop: `clipTupleContoursToRegionMask(baseContours, buildRegionMask([region]))`. A standalone `exclude` region resolves to "machine everything except this area", so a mixed op (`[include A, exclude B]`) emits scanlines over the whole model minus B in the exclude pass instead of carving a hole in A. Composing all region features into one `buildRegionMask(regionFeatures)` (or skipping `exclude` features in the loop and applying them as holes) fixes it.

### Minor

[`resolveFeatures.ts:438`](https://github.com/PureCutCNC/purecutcnc/blob/fd124e10c8b64ee1bf8448ac86bd2112b74a0bf7/src/store/helpers/resolveFeatures.ts#L438) copies `definition.regionMaskMode` without the `?? 'include'` default that `rawFeatureAdapter` ([`:481`](https://github.com/PureCutCNC/purecutcnc/blob/fd124e10c8b64ee1bf8448ac86bd2112b74a0bf7/src/store/helpers/resolveFeatures.ts#L481)) applies. Benign today because `normalizeProject` normalizes definitions, but the asymmetry is a latent trap.

### Scope question

If `exclude` isn't intended for these four operations yet, the real gap is that the Properties panel lets you set `exclude` on any region an operation can consume, with no guard or warning — so they silently cut the inverse rather than no-op'ing.

PR #253, opened 2026-07-04
feat: construction geometry layer (sketch-only reference features)

### LINKED ISSUE #199: Construction geometry layer (points / lines / circles / axes)
**Roadmap:** Phase 2 / P1 — construction geometry layer. This issue holds the implementation plan; the original roadmap seed is preserved under *Background* at the bottom.

---

# Design: Construction geometry layer

## 1. Summary

Introduce a third class of sketch entity — **construction** — that lives only in the 2D sketch view. It renders in a muted, lighter style, sits in its own **Construction** section of the project tree, and is **hard-excluded from the model**: 3D preview, voxel simulation, CSG, toolpaths, and G-code never see it. It stays fully available to *sketch-time* tooling: snapping, mirroring, constraint anchoring, trim/cut references, and dimensions. It is created from a new third button on the creation-target toolbar (alongside Feature and Region) and can be converted to/from a real feature, or to/from a region when its profile is closed.

Construction is implemented as a first-class sketch-feature *role* carrying any profile — which subsumes the roadmap's "points / lines / circles / axes" (each is just a construction feature of that shape).

## 2. Representation (resolved)

**Decision: a new `FeatureOperation` value `'construction'`** (not a separate boolean flag). This mirrors how `'region'` is modelled and how the tree already sections by `operation`.

**The critical consequence — and the work that delivers the issue's "CAM/CSG can't accidentally consume it" guarantee:** the codebase currently uses `feature.operation !== 'region'` as its de-facto "this is a machinable feature" test in **~40 sites across 15 files** (full surface: `grep -rn "!== 'region'\|=== 'region'" src`). A new operation value that is *not* handled at those sites would be treated as machinable and silently cut. So the plan is **not** "add a value and sprinkle `&& !== 'construction'`" — it is:

- Introduce one module `src/store/helpers/featureRoles.ts` exporting the single source of truth:
  - `isConstruction(f)` → `f.operation === 'construction'`
  - `isRegion(f)` → `f.operation === 'region'`
  - `isMachinable(f)` → `f.operation !== 'region' && f.operation !== 'construction'`  *(a real, cuttable feature)*
  - `modelFeatures(project)` → `project.features.filter(f => !isConstruction(f))`  *(everything CSG/3D/sim may see; regions still build display walls as today, construction is fully absent)*
- Replace the scattered `operation !== 'region'` "machinable" checks with `isMachinable(f)`, and route CSG / toolpath collection through `modelFeatures()`.
- Add a **guard test** that fails the build if any construction feature reaches CSG input or a toolpath target list. This is the belt-and-suspenders that makes the exclusion structural rather than convention.

Trade-off accepted with this representation: converting feature → construction → feature loses the original add/subtract intent (operation is the storage), so convert-back defaults to `subtract`. See §9.

## 3. Data model (`src/types/project.ts`)

- `FeatureOperation` → add `'construction'`: `'add' | 'subtract' | 'region' | 'model' | 'line' | 'construction'`.
- `FeatureFolder.section` → widen to `'features' | 'regions' | 'construction'`.
- `CreationTarget` (`src/store/types.ts`) → `'feature' | 'region' | 'construction'`.
- `LATEST_PROJECT_VERSION` → bump `'2.0'` → `'2.1'` (see §11).
- Operation lives on `FeatureDefinition` and the baked `SketchFeature`; construction propagates to linked instances through the existing operation-change path (`featureSlice` `updateFeature`), so linked copies share the role like they share the shape.

## 4. Creation toolbar (third choice)

- `src/components/layout/toolbar/CreationActions.tsx` — add a third `renderCreationTargetButton('construction', <icon>, 'Create construction')`; add a new icon SVG under `src/assets/icons/` and `npm run sync-icons`.
- `src/components/layout/ToolRail.tsx` — mirror the tablet/rail target button (currently feature/region only).
- Creation → operation mapping (`src/store/helpers/buildShapeFeature.ts`, `src/store/slices/pendingAddSlice.ts`): when `creationTarget === 'construction'`, set `operation: 'construction'`, name `Construction N`, and drop into the construction section.
- Shape availability: like region hides `text`, decide construction's allowed shapes (default: allow all incl. open polyline/spline, since construction is often reference lines/axes).

## 5. Project tree (own section/folder)

Mirror the Regions treatment in `src/components/feature-tree/FeatureTree.tsx`:

- Split into three sets: `machiningFeatures` = `isMachinable`, `regionFeatures` = `isRegion`, `constructionFeatures` = `isConstruction`; folders by `section`.
- New **Construction** section root alongside Features/Regions: collapse, add-folder(`'construction'`), show-all / hide-all, select-root — parallel to `setAllRegionsVisible` / `selectRegionsRoot` / `toggleRegionFolderVisible` in the store slices (`treeVisibilitySlice`, `featureSlice`) and `store/types.ts`.
- Row treatment: a construction badge/class analogous to `tree-row--region` / the "mask" badge.
- `AppShell.tsx` region count/visibility helpers (lines ~316-317) get construction counterparts.

## 6. Distinct rendering ("different colour, less intensive")

`src/components/canvas/previewPrimitives.ts` `drawFeature` — add an early `isConstruction` branch (this is the single place feature body color is chosen): desaturated grey-blue, **dashed** thinner stroke, no depth-weight fill, and skip the `showInfo` Z label (construction has no machining Z). Add CSS if a tree-side swatch is needed.

## 7. Hard exclusion from the model (the guarantee)

- **CSG** (`src/engine/csg.ts`, entry collection ~524/822/974): filter to `modelFeatures()` so construction never enters CSG. 3D preview (`Viewport3D`) and voxel simulation (`SimulationViewport`) both derive from CSG → excluded for free. Audit both for any direct `project.features` reads.
- **Toolpaths / resolver** (`src/engine/toolpaths/*`, `resolver.ts`): construction is never a machining target nor a region mask (`regions.ts` machinable/region split, `edge.ts`, `multiFeature.ts`, `toolSelection.ts`).
- **CAM target selection & validity** (`src/components/cam/CAMPanel.tsx`, `operationValidity.ts`, `store/helpers/operationDefaults.ts`, `operationsSlice.ts`): exclude construction from candidate/"select all" lists and reject it if targeted (`compatibleFeatureIdsForOperation`).
- **Guard test** (new): assert `modelFeatures()` and every toolpath target list contain zero construction features.

## 8. Available to sketch tooling (snapping / mirror / constraint anchor / cut)

- **Snapping** — `resolveSketchSnap` (`snappingHelpers.ts`) reads the profile list built from `project.features`; construction is included automatically. Verify it is not filtered by operation upstream.
- **Mirror / trim / cut references** — these act on selected sketch geometry, not on `modelFeatures()`, so construction participates as a source/axis/reference without being machined.
- **Constraint anchor** — construction geometry may be referenced by other features' constraints as an anchor (it just does not carry its own — see §10).

## 9. Conversion (construction ↔ feature ↔ region)

Extend the existing operation-change path (`featureSlice.updateFeature`, which already propagates to the definition + linked siblings and applies region special-cases at ~559/699):

- **→ feature:** set `operation` to `add`/`subtract` (default `subtract`; respect the "first machining feature forced to `add`" rule). Open profiles only valid where a machining op accepts them (guard via the `projectStore` open-profile check, line ~119, which already whitelists `line`/`model`/`region` — add `construction`).
- **→ region:** requires a **closed** profile; set `operation: 'region'`, move to Regions section, apply region Z-strip.
- **feature / region → construction:** set `operation: 'construction'`, move to Construction section.
- Reuse the existing tree "operation menu" (`FeatureTree.tsx` ~1067) to surface these as convert actions.

## 10. Dimensions yes, constraints deferred (resolved)

- **Dimensions apply** to construction features — the `LocalDimension` plumbing is independent of machining and works unchanged.
- **Own constraints deferred** for v1: keep the constraint-add UI disabled for construction (revisit later). Construction can still be an *anchor* for other features' constraints.

## 11. Persistence & versioning

- New additive fields; forward/back compatible on load (missing/unknown ⇒ treated as before).
- **Version bump to `'2.1'`** (resolved): an older build opening a file containing construction geometry would not recognise `operation: 'construction'` and would treat it as a normal (subtract) feature — i.e. machine it, the exact failure we guard against. Bumping `LATEST_PROJECT_VERSION` makes older builds show the existing "newer file" warning (`openProjectFromText` → `loadWarning`). Confirm the loader's operation-union handling degrades gracefully for `'2.1'` files opened best-effort.
- Verify `.camj` round-trip and `normalize.ts` (line ~487 machinable check) handle the new value.

## 12. File touchpoints (checklist)

- Types: `src/types/project.ts`, `src/store/types.ts`
- New: `src/store/helpers/featureRoles.ts` (+ tests), guard test, construction icon SVG
- Creation: `CreationActions.tsx`, `ToolRail.tsx`, `buildShapeFeature.ts`, `pendingAddSlice.ts`, `commands/creationShapes.ts`, `usePointerGestures.ts`, `useCanvasKeyboard.ts`, `SketchCanvas.tsx`
- Tree: `feature-tree/FeatureTree.tsx`, `PropertiesPanel.tsx`, `treeVisibilitySlice.ts`, `featureSlice.ts`, `AppShell.tsx`
- Render: `canvas/previewPrimitives.ts` (+ CSS)
- Exclusion: `engine/csg.ts`, `engine/toolpaths/{resolver,regions,edge,multiFeature}.ts`, `engine/operations/toolSelection.ts`, `components/cam/{CAMPanel,operationValidity}.tsx/.ts`, `store/helpers/{operationDefaults,naming,normalize}.ts`, `store/slices/{operationsSlice,tabsSlice}.ts`, `store/projectStore.ts`, `text/index.ts`
- Replace `operation !== 'region'` "machinable" checks with `isMachinable()` throughout the above.

## 13. Phasing (mergeable slices)

1. **Model + gate:** add `'construction'` value, `featureRoles.ts`, replace machinable checks with `isMachinable`, route CSG/toolpaths through `modelFeatures()`, guard test. No UI yet.
2. **Create + show:** toolbar/rail third target, creation mapping, tree Construction section, distinct rendering.
3. **Exclusion polish:** CAM target/validity exclusion, Viewport3D/Simulation audit, version bump + loader warning.
4. **Conversions:** construction ↔ feature ↔ region with guards.
5. **Dimensions verified; grouping guard** (construction groups only with construction — mostly free from the section split; add an explicit cross-section group guard).

## 14. Testing

- Unit: `featureRoles` predicates; `buildShapeFeature`/`pendingAdd` construction path; conversion state transitions (incl. open→region rejection).
- Guard: zero construction features in CSG input / toolpath targets / CAM candidate lists.
- Structural (source-read style, like `regionPresentation.test.ts`): construction badge/section rendered; `drawFeature` construction branch present.
- Round-trip: `.camj` with construction survives save/load; `'2.1'` warning path.

## 15. Open items for implementation

- Which shapes the construction target offers (default: all, including open polyline/axis lines).
- Exact construction palette/dash spec.
- Whether "mirror/cut" tools need any explicit opt-in for construction sources or already include them.

---

## Background (roadmap seed)

A construction layer that gives dimensions, mirror, and placement durable references.

- Construction points, lines, circles, axes; rendering/visibility; snap participation.
- **Hard exclusion from CAM targets and CSG** — construction geometry must never be consumed as a machining target or boolean input.
- Open decision (from the roadmap): store as a special feature kind, or as a flag on compatible geometry. The choice must guarantee CAM/CSG can't accidentally consume it. → **Resolved above: new `operation` value + centralized `isMachinable`/`modelFeatures` gate + guard test.**

_Prerequisite for deeper constraints and layout helpers._

zabooma:
Implements the construction geometry layer designed in #199 — a third class of sketch entity for reference geometry that lives only in the sketch view and can never be machined.

## What's in here

**Representation & the exclusion guarantee (plan §2, §7)**
- New `FeatureOperation` value `'construction'`; `FeatureFolder.section` and `CreationTarget` widened.
- New [`src/store/helpers/featureRoles.ts`](../blob/feat/issue-199-construction-geometry/src/store/helpers/featureRoles.ts) — the single source of truth: `isMachinable` / `isRegion` / `isConstruction`, `modelFeatures()` (the CSG gate), `sectionForOperation`. The ~40 scattered `operation !== 'region'` "machinable" checks in store/engine/CAM now go through these predicates.
- Hard exclusion wired at every seam: `csg.ts buildScene` (3D preview + simulation derive from it), toolpath target splitting (`regions.ts`), per-feature operation fan-out (`multiFeature.ts`), tool sizing, CAM validity + target builders (early reject with a clear hint when construction is in the selection), operation fallback targets, and the 3D camera fit.
- **Guard test** `src/engine/constructionExclusion.test.ts` fails the build if construction ever becomes a machining target, region mask, or CSG input.

**Creation & display (plan §4–§6)**
- Third creation-target button (desktop toolbar + tablet rail) with a new `construction` icon; all shape tools work, text stays feature-only; open polyline/spline/composite paths become construction (not `line`) under this target.
- Own **Construction** tree section mirroring Regions: folders, show/hide all, per-folder visibility, `ref` badge, muted styling; statusbar Construction toggle.
- 2D canvas renders construction muted grey-blue, dashed, thinner, never filled, no Z label.

**Sketch tooling stays available (plan §8)**
- Snapping, move/mirror/rotate/offset, and cutting (construction as cutter) all work — they're selection-driven and unfiltered.
- Dimensions apply. Own constraints are deferred per the plan: `beginConstraint` no-ops on construction; construction can still be another feature's constraint reference.

**Conversions (plan §9)**
- Tree operation menu: closed features offer Add / Subtract / Region / Construction; open profiles offer Line ↔ Construction only (so an open profile can never become a region).
- Multi-select operation dropdown gains Construction.
- Z edits are stripped for construction (like regions); converting moves the feature to the right tree section; cross-section folder moves/assignments are now rejected/re-rooted (also fixes a latent hole where a region dropped into a features folder vanished from the tree).
- First-feature rule: converting the first row to construction is allowed (it leaves the model); the first *machinable* feature is still forced to Add.

**Persistence (plan §11)**
- `LATEST_PROJECT_VERSION` → `'2.1'`, stamped **by content**: a file is saved as 2.1 only when it actually contains construction geometry, so pre-construction builds warn on exactly the files they could silently machine wrong.
- The legacy open-profile → `line` load migration skips construction, so open construction geometry survives round trips.

## Tests
- `featureRoles.test.ts` — predicate partition + sectioning.
- `constructionExclusion.test.ts` — the exclusion guard (functional + structural).
- `constructionWorkflows.test.ts` — creation target, conversions, section integrity, deferred constraints, version stamping, open-profile round trip.
- `constructionPresentation.test.ts` — structural UI/CSS assertions (mirrors `regionPresentation.test.ts`).

`npm run build` green (lint + tsc + all 87 test files + vite).

Closes #199


zabooma:
Review findings requiring action:

1. `src/store/slices/featureSlice.ts:571-583`, plus `updateFeatures` at `:721-724`, still enforce the Add rule against physical row 0 instead of the first machinable feature. With construction at row 0, the real first model feature can be changed to subtract/region, or converted to construction while a later subtract becomes first machinable. That creates the invalid state already covered by `isFirstFeatureValid([construction, subtract])`. The UI locks have the same mismatch: `FeatureTree.tsx:315` passes `index === 0`, and `PropertiesPanel.tsx:1262` still uses `operation !== 'region'`. Please compute/enforce first machinable via `isMachinable` in the store and UI.\n\n2. `src/components/feature-tree/PropertiesPanel.tsx:1162` adds Construction to the multi-select operation dropdown, but that path calls `updateFeatures`, which only mutates the selected rows. It does not update `featureDefinitions` or linked siblings like `updateFeature` does, while resolver read paths take `operation` from `definition.operation`. For linked instances, multi-select conversion to Construction can leave resolved hit/snap/edit paths seeing the old machinable operation. Please route operation changes through the same propagation logic as `updateFeature`, or add equivalent propagation to `updateFeatures`.\n\n3. `src/components/feature-tree/PropertiesPanel.tsx:1281-1302` still does not handle Construction in the single-feature Operation control. A closed construction feature reaches the Select with `value="construction"` but options only include Subtract/Add/Region, and an open construction feature is displayed as Line. That makes the properties panel inconsistent with the tree operation menu and gives construction rows a blank/misleading operation state. Please add Construction/Line handling there too.\n\nVerification run while reviewing:\n- `npx tsx src/store/constructionWorkflows.test.ts` passed\n- `npx tsx src/engine/constructionExclusion.test.ts` passed\n- `git diff --check 4ac6b770877b337ca3bfc751c67a0d93256ca3ff...e50e29f152219c1e7ce5ec9980457ebea11bfe26` passed

zabooma:
All three findings addressed in 8b92b1f — thanks, all real.

**1. Base-solid rule vs physical row 0** — the rule now tracks the first *machinable* feature end to end:
- `sanitizeOperationPatch` (new in `normalize.ts`, shared by `updateFeature` and `updateFeatures`) computes the first machinable feature *after* the patch and forces `add` when the edited row would be it. The construction exemption falls out naturally (a row converting to construction is no longer machinable, so it can't be the base).
- `enforceFirstMachinableAdd` (also `normalize.ts`) re-asserts the invariant after any operation edit, cascading to the successor — so converting the base to construction (or region) while a later subtract exists turns that subtract into the new `add` base, exactly the semantics `reorderFeatures` already had. Note: like `reorderFeatures`, the cascade adjusts the raw row without rewriting its shared definition — that divergence tolerance is pre-existing (`reorderFeatures`, and previously the row-0 sibling force).
- UI locks follow: `FeatureTree` passes `feature.id === firstMachiningFeature?.id` instead of `index === 0`, and `PropertiesPanel`'s `firstMachiningFeature` uses `isMachinable`.
- One deliberate behavior change worth flagging: converting the base feature to **region** used to be silently rewritten to `add` (row-0 force); it now succeeds and the cascade protects the successor instead. That seemed more honest than converting a "make this a region" click into "make this add" — shout if you'd rather keep the old force.

**2. `updateFeatures` propagation** — added `propagateOperationToLinkedInstances` (`featureDefinitions.ts`): bulk operation changes now write through to each changed row's shared definition and mirror onto linked siblings *outside* the selection (with folder-section reconciliation), matching `updateFeature`'s P1b semantics, then the base-solid cascade runs. Covered by `testBulkLinkedPropagation` (reference copy → bulk convert one instance → sibling + definition follow, both directions).

**3. Single-feature Operation control** — now construction-aware: closed features get a `Construction` option alongside Add/Subtract/Region; open profiles get a `Line ↔ Construction` select (mirroring the tree menu) instead of rendering construction rows as a locked "Line".

New coverage: `testFirstMachinableRule` (force with construction at row 0; cascade on base→construction) and `testBulkLinkedPropagation` in `constructionWorkflows.test.ts`, plus structural assertions for the panel/tree lock changes. Full `npm run build` green (all 88 test files).

PR #255, opened 2026-07-05
Print Design: CAD-style printing for the 2D sketch/design view

### LINKED ISSUE #254: Implement printing
### Goal
Add printing for the 2D design view: the stock/sketch/design content a user sees in the Sketch workspace, with CAD-style paper and scale controls. This is not operation-booklet printing, G-code documentation, 3D/simulation printing, or a full project report.

### User-facing behavior
- Add a `Print Design...` command that opens a dedicated print dialog.
- Print target is the 2D sketch/design view, using the current project state and visibility flags.
- Default printed content:
  - visible stock outline/profile
  - visible sketch features, including text-resolved geometry and imported-model silhouettes/top-view assets where available
  - visible origin marker
  - visible dimensions when `showDimensions` is enabled
  - visible tabs/clamps when their project visibility is enabled
- Optional content toggles in the dialog:
  - grid
  - backdrop image
  - feature labels/info
  - tabs/clamps
  - toolpath overlays, only when visible in the sketch view
  - simple footer/title block with project name, units, scale, paper, and date
- Do not print the app shell, side panels, CAM operation settings, G-code, operation booklet content, 3D preview, or simulation output.

### CAD-style print controls
- Paper presets: Letter, Legal, Tabloid, A4, A3, plus custom size.
- Orientation: portrait / landscape.
- Margins: default sane margins, editable in project units or paper units.
- Print area:
  - `Visible design extents` (default): bounds of visible stock/features/tabs/clamps/origin/backdrop, matching the sketch model rather than the UI viewport.
  - `Stock extents`: stock profile bounds even when features are outside stock.
  - `Current sketch view`: the current pan/zoom viewport; enabled when the sketch canvas is mounted.
- Scale:
  - `Fit to page` (default): preserve aspect ratio and center on the printable area.
  - `Actual size (1:1)`: project units map to physical paper units (`1 in = 1 in`, `1 mm = 1 mm`).
  - `Custom`: percentage or ratio (`1:2`, `2:1`, etc.).
- Show computed output size and a warning when the selected scale clips the drawing on the chosen paper.
- Output is centered on the printable area by default, with optional X/Y offsets for users who need registration.

### Rendering approach
Use a print-specific vector renderer instead of screenshotting the canvas. The existing sketch canvas renderer is immediate-mode `CanvasRenderingContext2D`; it is good for interaction, but not for accurate paper scale or crisp print output.

Create a DOM-free print renderer under a new engine folder, for example:

- `src/engine/designPrint/types.ts` — print options, paper presets, scale modes, content toggles.
- `src/engine/designPrint/layout.ts` — page size, margin, printable-area, bounds, scale, clipping, and centering calculations.
- `src/engine/designPrint/svg.ts` — converts project geometry into an SVG using project-space coordinates and physical output sizing.
- `src/engine/designPrint/html.ts` — wraps the SVG in a printable HTML document with `@page`, white background, print CSS, footer/title block, and no app UI.
- `src/engine/designPrint/index.ts` and `INDEX.md` — public exports and folder documentation.

Implementation details:
- Reuse existing geometry helpers where possible:
  - `getVisibleSceneBounds2D()` / extracted bounds logic for visible-design extents.
  - `getStockBounds()` and `getProfileBounds()` for stock and profile bounds.
  - `getFeatureGeometryProfiles()` so text features print as resolved geometry.
  - pure dimension helpers from `src/sketch/dimensions.ts` (`dimensionLayout`, `dimensionLabelText`, etc.) for dimension geometry and labels.
- Add SVG-specific path/shape helpers rather than trying to reuse canvas drawing calls directly.
- Use print-friendly colors on white paper. Preserve useful feature color distinctions in color mode, and offer a monochrome mode for shop drawings.
- Keep selection/hover/editing state out of the print output unless explicitly added later.
- Use physical CSS/SVG sizes so actual-size scale is not dependent on screen DPI.

### UI integration
- Add `PrintDesignDialog` under `src/components/export/PrintDesignDialog.tsx`.
- Add app state in `src/App.tsx` to open/close the dialog.
- Add a `printDesign` command to `src/commands/fileCommands.ts` so desktop toolbar and tablet top command bar can share it.
- Add a print icon source under `src/assets/icons/print.svg` and regenerate `public/icons.svg` with `npm run sync-icons`.
- Add the command to:
  - desktop/global toolbar file group
  - tablet/compact top command bar file group, with responsive checks so buttons do not overlap
  - Tauri native File menu as `Print Design...` with `CmdOrCtrl+P`
- Intercept `Cmd/Ctrl+P` in the web app when focus is not in an editable field and open `PrintDesignDialog` instead of printing the app shell.
- Extend `useDesktopIntegration` with an `onPrintDesign` callback and handle the new native menu event.

### Platform printing
- Add a platform/helper path for printing generated HTML, shared by browser and Tauri where possible.
- Prefer a generated print document opened from the print dialog action, then call `print()` on that document, instead of temporarily mutating/hiding the main app UI.
- Browser implementation should avoid popup-blocker failures by starting the print target synchronously from the user action or using a controlled hidden iframe.
- Desktop implementation should use the same generated document flow first; do not add a native Tauri print plugin unless the WebView print path proves insufficient.
- The print document should be self-contained: inline SVG, inline CSS, no dependency on app layout classes.

### Current-view support
- Extend `SketchCanvasHandle` with a read-only method such as `getVisibleWorldBounds(): Bounds2D | null`.
- `App.tsx` passes that to `PrintDesignDialog`; when unavailable, the dialog disables `Current sketch view` and falls back to visible-design extents.
- This is a read-only handle addition; it must not mutate sketch state or broaden canvas editing behavior.

### Tests and verification
- Add focused unit tests for print layout math:
  - Letter and A4 printable-area calculations.
  - mm and inch actual-size scale conversion.
  - fit-to-page scale preserves aspect ratio.
  - custom scale/ratio parsing.
  - clipping detection when actual size is larger than paper.
- Add SVG/HTML smoke tests with a small deterministic project:
  - stock + feature paths appear.
  - hidden features are omitted.
  - dimensions appear only when enabled.
  - generated SVG has physical size/viewBox values consistent with selected scale.
- Run:
  - `npx tsx src/engine/designPrint/designPrint.test.ts`
  - `npm run build`
- Manual checks before PR:
  - Browser print preview shows only the design document, not the app shell.
  - `Fit to page`, `Actual size`, and `Custom scale` change the preview and warnings correctly.
  - `Cmd/Ctrl+P` opens `Print Design...` outside text inputs and does not steal normal text-field behavior.
  - Tablet/top-command-bar layout does not overlap after adding the print button.

### Out of scope for this issue
- Operation setup sheets/booklets.
- Printing G-code or CAM operation settings.
- 3D preview or simulation print output.
- Multi-page tiling for oversized 1:1 prints. V1 should warn about clipping and offer fit/custom scale; tiling can be a follow-up issue.
- Native printer settings beyond the browser/OS print dialog.

zabooma:
Closes #254

## What this adds

A `Print Design…` command that prints the 2D sketch/design view with CAD-style paper and scale controls, per the design in the issue.

### Engine (`src/engine/designPrint/`, DOM-free)
- **layout.ts** — paper presets (Letter, Legal, Tabloid, A4, A3, custom), orientation, margin/printable-area math, print-bounds resolution (visible design extents / stock extents / current sketch view), fit / actual-size (1:1) / custom scale (`1:2`, `50%`, `0.5`), centering with X/Y registration offsets, and clipping detection.
- **svg.ts** — print-specific vector renderer producing a self-contained SVG in physical millimetres (actual size is DPI-independent). Renders stock outline, visible features (text resolved via `getFeatureGeometryProfiles`, **construction geometry as dashed unfilled reference marks**, imported-model silhouette paths), tabs/clamps, origin marker, dimension annotations (gated on `showDimensions`), optional grid/backdrop/feature-labels/toolpath overlays, and a footer/title block (project name, units, scale, paper, date). Print-friendly palette on white with a monochrome mode for shop drawings; selection/hover/editing state never prints.
- **html.ts** — wraps the SVG in a printable HTML document (`@page` sized to the paper, no app CSS).

### UI & wiring
- `PrintDesignDialog` (`src/components/export/`) with live paper preview, warnings for clipping and invalid custom scales (Print disabled on invalid scale), and computed output size.
- `printDesign` file command: desktop toolbar button, tablet top-command-bar button (hidden below 1080px so the bar never overlaps), `print` icon (`npm run sync-icons`).
- `Cmd/Ctrl+P` opens the dialog when focus is not in an editable field (web); Tauri gets a native `File → Print Design…` menu item with `CmdOrCtrl+P` routed through a new `onPrintDesign` desktop-integration callback.
- Printing goes through a hidden same-origin iframe (`src/platform/printDocument.ts`) created synchronously from the user action — no popups, and the app shell is never mutated or printed.
- `SketchCanvasHandle.getVisibleWorldBounds()` — read-only handle addition backing the "Current sketch view" print area; the dialog omits that option when the canvas isn't laid out.
- `getVisibleSceneBounds2D` moved to `src/sketch/sceneBounds.ts` (pure) so the engine can reuse it; re-exported from `viewTransform.ts` for existing callers.

## Tests & verification
- `npx tsx src/engine/designPrint/designPrint.test.ts` — printable-area math (Letter/A4), mm+inch actual-size conversion, fit aspect-ratio preservation, custom-scale parsing, clipping (including offset-induced), bounds resolution, SVG/HTML smoke tests (hidden features/construction omitted, dimensions gated on `showDimensions`, construction dashed+unfilled, XML escaping, physical `viewBox`/`@page` sizing).
- `npm run build` green (lint + icons + tsc + all 88 structural test files + vite).
- Manual browser verification on the example project: dialog preview updates across fit/actual/custom scales; clipping and invalid-scale warnings show correctly; captured print document contains only the design SVG + `@page` rule (no app shell); `Ctrl+P` opens the dialog outside text fields and is ignored while renaming the project; toolpath overlays render when toolpaths are visible.

## Notes / follow-ups
- The Tauri desktop path uses the same iframe + `window.print()` flow per the issue ("no native print plugin unless the WebView path proves insufficient"). WKWebView's `window.print()` support should be confirmed on a real desktop build; if it's a no-op there, a follow-up can add a native fallback.
- The Rust menu change compiles by symmetry with the existing items but wasn't `cargo check`ed here (no Rust toolchain run in this worktree).
- Multi-page tiling for oversized 1:1 prints is out of scope (v1 warns about clipping), as specified in the issue.


zabooma:
Review findings:

- [P1] `src/platform/printDocument.ts:55-61` waits on `doc.images`, but the backdrop produced by `src/engine/designPrint/svg.ts:344-359` is an inline SVG `<image>` element, not an HTML ``. With the Backdrop image option enabled, `win.print()` can run before a large backdrop data URL has loaded, so the printed output can omit the reference image even though the preview includes it. Please wait for SVG image elements as well, or otherwise defer `print()` until the inline SVG/image resources are actually loaded.

- [P2] `src/components/export/PrintDesignDialog.tsx:64-79` uses `getVisibleSceneBounds2D(project)` for the default `visible` print area, while the default content options leave `backdrop: false` (`src/engine/designPrint/types.ts:153-160`). Because `getVisibleSceneBounds2D` includes any visible backdrop in the bounds (`src/sketch/sceneBounds.ts:80-87`), a project with a large tracing image opens Print Design with the real CAD geometry scaled down to fit an image that is not printed. The bounds should be derived from the enabled print layers, or the default backdrop toggle should match the bounds behavior so the first print preview reflects what will actually be printed.

- [P2] `src/styles/tablet.css:593-599` hides the only tablet top-bar Print Design button below 1080px and says the fallbacks are Cmd/Ctrl+P and the desktop menu. On a touch-only web tablet at common 1024px widths, there is no desktop menu and no guaranteed hardware keyboard path, so the new print dialog becomes unreachable from the app UI. Please keep a touch-accessible entry point for compact tablet, for example in an overflow/file menu or drawer, instead of removing the command entirely.

zabooma:
All three findings addressed in 2092c81:

- **[P1] SVG `<image>` not awaited before print** — `printHtmlDocument` now waits for SVG `<image>` elements in addition to `document.images`: `decode()` where available, otherwise `load`/`error` listeners with a 1.5s cap (SVGImageElement has no `complete` flag, so the cap covers events that fired before the listeners attached).

- **[P2] Bounds include the unprinted backdrop** — `resolvePrintBounds` now derives visible-design extents from the enabled content layers: `getVisibleSceneBounds2D` grew include flags for tabs/clamps, and the backdrop widens the bounds exactly when its toggle will draw it (regardless of sketch visibility, matching the print semantics). The dialog passes its content toggles through, so flipping backdrop/tabs/clamps rescales the preview to precisely what the output contains, and the initial orientation heuristic also excludes the default-off backdrop. Covered by new unit tests.

- **[P2] Print unreachable on compact touch tablets** — the `display: none` hide is gone. Below 1080px the command bar tightens its gaps (fits at 1024px even with 44px coarse-pointer touch targets — verified with emulation; buttons keep their `--control-hit-target` minimums) and the center section becomes swipe-scrollable with `justify-content: safe center` (flex-start fallback) so nothing can overlap and no edge becomes unreachable.

PR #258, opened 2026-07-06
Add gear creation workflow

### LINKED ISSUE #218: Implement gear feature creation
## Goal

Add a Gear creation tool alongside Slot, Regular Polygon, Rounded Rectangle, and the other creation shapes. The user selects the gear centre and outside radius on the canvas, with exact radius entry following the same dimension-entry workflow used by existing creation tools. All gear-specific parameters are entered directly in the workflow dialog, and the result is ordinary sketch geometry that can be selected, edited, machined, saved, and loaded like any other feature.

## Approach

- Add `Gear` as a secondary creation shape in the existing creation toolbar/menu. It should reuse the current creation target controls and the `CanvasWorkflowPanel` creation flow rather than introducing a new modal pattern.
- Interaction flow:
  - Activate Gear.
  - Click/tap the gear centre.
  - Move/click/tap the outside point to set the **outside radius**. In gear terminology this is the addendum circle radius / outside radius; the plan intentionally does not use pitch radius as the canvas-selected radius.
  - `Radius`/dimension entry opens the same numeric-entry path as Circle/Regular Polygon: `triggerDimensionEdit()`, `DimensionEditState`, `computeDimensionEditPreviewPoint()`, and `commitCreationDimensionEdit()` get a `gear` branch that resolves the outside-radius point from the centre.
  - After the outside radius exists, show a compact Gear Parameters dialog inside the workflow panel. Focus moves into dialog inputs only for this numeric-entry phase, then returns to the canvas after Confirm/Cancel.
- Gear parameter terminology and fields:
  - **Number of teeth**: integer, default 24, bounded to a practical range such as 3-200.
  - **Outside radius**: prefilled from the canvas; editable through the existing radius dimension entry, not duplicated as the main dialog control.
  - **Whole tooth depth**: radial distance from the addendum circle to the root circle. Default from standard full-depth proportions; root diameter is derived as `2 * (outsideRadius - wholeDepth)`.
  - **Tooth flank profile**: `Involute` default, plus `Straight-sided` for simple/trapezoidal decorative or non-meshing gears.
  - **Pressure angle**: shown for involute gears, default 20 degrees.
  - **Root form**: `Rounded root fillet` default, plus `Flat root` and `Sharp root`. When rounded, expose **root fillet radius**.
  - **Crest form**: `Flat crest` default, plus `Rounded crest` if the geometry helper can produce it cleanly without unstable tiny arcs.
  - **Bore diameter**: optional shaft hole diameter. Empty or 0 disables it.
- Geometry generation:
  - Add a pure helper such as `buildGearProfile(params: GearProfileParams): SketchProfile` in `src/sketch/gearProfile.ts` rather than putting complex gear math into `project.ts`, `SketchCanvas.tsx`, or a store slice.
  - Store the generated gear as a normal closed single-loop `SketchProfile`, using existing `Segment` primitives. Do not add a parametric gear schema to `.camj` in this issue.
  - For involute gears, derive the standard module/pitch radius from outside diameter and tooth count (`module = outsideDiameter / (teeth + 2)`, `pitchRadius = module * teeth / 2`) and use the entered whole depth for the root radius. Sample each involute flank with a bounded number of line segments, then connect roots/crests according to the selected root/crest forms.
  - For straight-sided gears, generate trapezoidal teeth using the same outside/root radii and tooth pitch, with the selected root/crest forms.
  - Validate inputs before committing: outside radius > 0, whole depth > 0, root radius > 0, bore diameter < root diameter, finite pressure angle, and no root/crest fillet larger than the available tooth space.
- Bore handling:
  - `SketchProfile` is one closed loop, so the optional shaft hole should not be encoded as an inner contour in the same profile.
  - When Bore diameter > 0, create a separate concentric circle feature named like `Gear Bore N`, inserted immediately after the gear feature, with operation `subtract` and matching Z range.
  - When a bore is created, also create a feature group/folder named like `Gear N` and place both the outer gear and bore inside it. The group should live in the same feature-tree section as the gear, keep the two shapes adjacent and easy to manage, and be created as part of the same history step.
  - The outer gear follows the current `buildShapeFeature()` / `addFeature()` operation rules. The created gear/bore group should remain easy to find after creation, ideally with the group or both features selected.
- Keep tablet behavior first-class:
  - The workflow panel uses intent labels, not visible keyboard-key labels.
  - In canvas-placement phases the canvas keeps focus; in dialog-entry phases inputs take focus; Confirm/Cancel return focus to the canvas through `focusCanvasAfterAction()`.
  - The Gear Parameters dialog must fit the existing tablet workflow panel width. Use compact labels, selects, and numeric inputs; avoid nested cards.

## Files affected

- `src/sketch/gearProfile.ts` *(new)* - pure `GearProfileParams`, defaults/validation helpers, and `buildGearProfile()` geometry construction.
- `src/sketch/gearProfile.test.ts` *(new)* - focused geometry tests for tooth count, closure, outside/root radius bounds, profile modes, root forms, invalid inputs, and bore validation boundaries.
- `src/store/types.ts` - add `PendingAddTool` variant for `shape: 'gear'`, gear parameter types if they are shared with the store, and store action signatures.
- `src/store/slices/pendingAddSlice.ts` - add `startAddGearPlacement`, gear parameter updates, and placement completion for centre/radius creation.
- `src/store/slices/featureSlice.ts` - add a small gear creation action that inserts the outer gear and optional bore in one history step, creating a feature group/folder for the pair when a bore is present.
- `src/components/canvas/previewPrimitives.ts` - draw the pending gear preview using `buildGearProfile()` and the current parameter defaults.
- `src/components/canvas/triggerDimensionEdit.ts` - open exact outside-radius entry for a pending gear after the centre is selected.
- `src/components/canvas/manualEntry.ts` - extend `DimensionEditState` / `computeDimensionEditPreviewPoint()` for gear outside-radius entry.
- `src/components/canvas/useCreationWorkflow.ts` - commit or cancel gear radius entry and transition to the gear parameter dialog.
- `src/components/canvas/useCanvasKeyboard.ts` - keep desktop keyboard accelerators mapped to the same gear workflow state transitions.
- `src/components/canvas/SketchCanvas.tsx` and, if the JSX would grow too much, `src/components/canvas/GearParameterPanel.tsx` *(new)* - render the Gear Parameters controls inside the creation workflow panel.
- `src/commands/creationShapes.ts` - wire Gear into creation command activation.
- `src/components/layout/toolbar/shared.ts` and `src/components/layout/toolbar/CreationActions.tsx` - add Gear to the secondary creation tool list.
- `src/assets/icons/gear.svg` *(new)* and generated `public/icons.svg` - add the toolbar icon through the SVG-first icon pipeline.
- `src/INDEX.md`, `src/store/INDEX.md`, and nearest affected indexes - update documentation for the new sketch helper/test and store behavior.

## Tests

- `npx tsx src/sketch/gearProfile.test.ts`
  - involute gear closes and produces the requested number of teeth.
  - all generated outer points stay within outside radius tolerance.
  - root points respect `outsideRadius - wholeDepth`.
  - straight-sided profile produces stable closed geometry.
  - rounded/flat/sharp root forms do not self-intersect for representative tooth counts.
  - invalid inputs fail with clear validation errors.
- `npx tsx src/store/gearCreation.test.ts` *(new or folded into an existing store creation test)*
  - gear creation adds one outer gear feature with ordinary profile geometry.
  - optional bore creates a concentric circle feature after the gear.
  - optional bore creation creates a `Gear N` group/folder containing both the gear and bore features.
  - gear + bore + group creation is one undo step.
  - first-feature operation behavior still follows existing `addFeature()` rules.
- UI/manual verification during implementation:
  - desktop: create gear by click radius, exact radius entry, parameter dialog Confirm/Cancel.
  - tablet/coarse pointer: same flow with touch; focus returns to the canvas after each action.
- `npm run build`

## Open questions / risks

- This plan treats the clicked radius as the outside/addendum radius. If we want pitch radius instead, that should be changed before implementation because it affects the dialog defaults and geometry math.
- Involute curves will be approximated with bounded line segments. That is appropriate for the current `SketchProfile` model, but the segment count must be capped so high-tooth-count gears do not make the canvas or toolpath generation sluggish.
- Bore-as-separate-feature is the right fit for the current one-loop `SketchProfile` model, but it means the gear is not one parametric multi-contour object. Grouping the gear and bore keeps the user-facing feature tree manageable, but it is still not a native multi-loop profile. A true multi-loop/profile-with-holes schema is out of scope here.
- Rounded crest support should be dropped or deferred if it creates tiny arcs or self-intersections near small tooth counts. Flat crest is the reliable default.

## Out of scope

- Parametric editing of gear parameters after creation.
- Multi-loop `SketchProfile` / native holes in a single feature.
- Gear trains, racks, sprockets, worms, bevel/helical gears, backlash simulation, strength checks, or AGMA/ISO compliance beyond using standard terminology and common involute defaults.
- Automatic machining operation setup for gears or bores.
- Toolbar redesign beyond adding one secondary creation tool and icon.

zabooma:
## Summary
- Add a gear creation tool alongside the existing sketch creation tools, with center/radius placement and a parameter dialog for gear-specific values
- Generate configurable gear profiles with tooth count, whole depth, flank profile, pressure angle, root form, root fillet radius, crest form, crest radius, and optional bore diameter
- Create a grouped gear+bore feature set when a center bore is requested
- Add gear parameter reference diagrams, toolbar/icon wiring, preview/manual-entry integration, and focused gear geometry/store tests

Closes #218

## Verification
- `npx tsx src/sketch/gearProfile.test.ts`
- `npm run build`


zabooma:
## Code review — Add gear creation workflow

Reviewed the diff. The wiring is thorough and consistent — I verified ID generation (global counter, no gear/bore collision), the extrude depth convention (`Math.min(stock.thickness, 10)` matches every other shape), preview/commit parity, dimension-edit + keyboard variable scoping, and the toolbar index shifts. Those are all correct. Two findings:

### 1. Involute teeth self-intersect at the tip for high pressure angle + low tooth count
**`src/sketch/gearProfile.ts:258`** — `leftCrestAngle`/`rightCrestAngle` are computed as `center ∓ halfToothAngleAtPitch ± roll`, but nothing clamps them when the involute `roll` exceeds `halfToothAngleAtPitch` (a pointed/crossed tooth), and `validateGearProfileParams` (line 114) has no guard for it. With parameters that are all inside the allowed ranges (`teeth ≥ 6`, `pressureAngleDeg ≤ 35`, default `crestForm: 'flat'`), the two flanks cross at each tooth tip.

Running the actual `buildGearProfile`:
- `teeth=6, pa=35` → 6 self-intersections
- `teeth=8, pa=35` → 8
- `teeth=12, pa=35` → 12

i.e. one crossing per tooth. The result is a `closed: true` profile with a bowtie at every tip, which then feeds CSG/toolpath generation as malformed geometry.

**Suggested fix:** clamp the crest so the flanks degenerate to a point (pointed tooth) instead of crossing, or reject the combination in `validateGearProfileParams`.

*(The default gear — teeth 24, pa 20 — is clean.)*

### 2. Gear folder and feature/bore names use divergent counters (cosmetic)
**`src/store/helpers/gearFeature.ts:112`** — the group folder is numbered by a gear-folder counter (`gearFolderCount + 1`), while the gear feature (`Gear ${features.length + 1}`) and bore (line 124, `Gear Bore ${features.length + 1}`) are numbered by total feature count. After the first gear+bore, a second one yields folder "Gear 2" containing feature "Gear 3" / "Gear Bore 3". No functional impact — but a single gear-scoped counter for all three would read better.

*(I did not flag the grouped-insert path pushing history unconditionally vs. `addFeature`'s `transactionStart` guard — gear creation is never invoked inside an open history transaction, so it isn't reachable in practice.)*

PR #262, opened 2026-07-06
fix: decouple 3D preview model rebuild from selection

### LINKED ISSUE #261: Perf: decouple 3D preview model rebuild from selection changes
## Problem (verified by measurement)

The 3D preview rebuilds the **entire CSG boolean model on every selection change**, even though the model does not depend on selection.

- The scene-build `useEffect` in [`Viewport3D.tsx`](src/components/viewport3d/Viewport3D.tsx) lists `selection.selectedNode` (and `collidingClampIds`) in its dependency array, so every selection change re-runs it.
- It calls `buildScene(project, selectedClampId, selectedTabId, collidingClampIds)` in [`csg.ts`](src/engine/csg.ts), whose expensive step is `buildBooleanModel(project, visibleFeatures)` — the manifold union/difference of all features. **That depends only on project geometry, not selection.**
- Selection is used *solely* to color the selected fixture: `selectedClampId`/`selectedTabId` feed `tab.id === selectedTabId` / the clamp highlight, which only set a color + opacity on the tab/clamp meshes.
- Net effect: selecting a **feature** passes `null` for both clamp/tab ids — identical model inputs — yet the whole manifold model is rebuilt for zero visual change.

**Measured cost:** ~500 ms on a typical project, ~2300 ms on a heavy imported one. Because all three centre views are always mounted (`AppShell.tsx`), this freeze also stalls the sketch view on every click.

Deferral (gating the rebuild on the 3D tab being visible) was tried and rejected — it makes switching to the 3D tab slow and needs a spinner. We want the root-cause decoupling instead.

## Goal

Decouple selection from the expensive model rebuild. Selecting features/clamps/tabs must **not** trigger `buildBooleanModel`. The model rebuilds only when geometry actually changes (features / stock / tabs / clamps geometry). Clamp/tab selection-highlight and colliding-clamp coloring become cheap, separate updates (swap material color/opacity on the already-built fixture mesh), not a full scene rebuild.

## Plan

**`src/engine/csg.ts`**
1. Extract fixture-highlight coloring into two pure helpers:
   - `applyClampHighlight(mesh, selected, colliding)` — sets `material.color` + `material.opacity` (the exact values currently baked into `buildClampMesh`).
   - `applyTabHighlight(mesh, selected)` — same for tabs.
2. `buildClampMesh` / `buildTabMesh` build geometry with a neutral `MeshStandardMaterial`, then call the helper for the initial appearance. **Signatures keep their `selected`/`colliding` params** so `SimulationViewport.tsx` (the other `buildClampMesh` caller) is unaffected.
3. `buildScene(project)` drops the `selectedClampId` / `selectedTabId` / `collidingClampIds` params and builds fixtures unhighlighted. (Verified: `Viewport3D` is the only `buildScene` caller.)

**`src/components/viewport3d/Viewport3D.tsx`**
4. The scene-build effect no longer depends on `selection.selectedNode` or `collidingClampIds`; it calls `buildScene(project)`. Keeps the existing **150 ms debounce + `buildRequestRef` cancellation**. It stores the returned `clampMeshes` / `tabMeshes` maps in refs.
5. A new lightweight effect keyed on `selection.selectedNode` + `collidingClampIds` recolors the already-built clamp/tab meshes via the helpers — no rebuild.
6. After each rebuild, re-apply the current highlight (via a ref to the latest highlighter) so freshly-built fixtures immediately reflect the current selection/collision even when selection didn't change.

**Test**
7. Add a structural test (`src/engine/viewport3dSelectionDecoupling.test.ts`, same style as `constructionExclusion.test.ts`) asserting: `buildScene` takes only `project`; the scene-build effect's dep array excludes `selection.selectedNode`; a separate highlight effect keyed on `selection.selectedNode` exists; the `applyClampHighlight` / `applyTabHighlight` helpers exist.

## Verify

- Temporarily log in the boolean-model build path to confirm selecting features no longer triggers a rebuild, and that clamp/tab highlight + colliding-clamp coloring still update on the already-built meshes. Remove temp logging before the PR.
- `npm run build` (lint + tsc + tests + vite) green before committing.

Branch: `fix/issue-<NN>-decouple-3d-rebuild`. PR at the end with `Closes #<NN>`.

zabooma:
Closes #261

## Problem

The 3D preview rebuilt the **entire CSG boolean model on every selection change**, even though the model doesn't depend on selection.

The scene-build `useEffect` in `Viewport3D.tsx` listed `selection.selectedNode` (and `collidingClampIds`) in its dependency array, so every click re-ran `buildScene → buildBooleanModel` — the manifold union/difference of all features. That boolean model depends only on geometry; selection merely tints the selected clamp/tab. Selecting a **feature** passed `null` for both fixture ids (identical model inputs) yet rebuilt the whole model for zero visual change.

**Measured:** ~500 ms on a typical project, ~2300 ms on a heavy import. Because all three centre views are always mounted (`AppShell.tsx`), the freeze also stalled the sketch view on every click.

## Fix (root-cause decoupling)

- **`csg.ts`** — extract fixture coloring into pure `applyClampHighlight(mesh, selected, colliding)` / `applyTabHighlight(mesh, selected)` helpers that mutate an already-built mesh's material **in place** (same colors/opacity as before). `buildClampMesh`/`buildTabMesh` call them for the initial appearance and **keep their `selected`/`colliding` params**, so `SimulationViewport.tsx` (the other `buildClampMesh` caller) is unchanged.
- **`csg.ts`** — `buildScene(project)` drops the `selectedClampId`/`selectedTabId`/`collidingClampIds` params and builds fixtures unhighlighted.
- **`Viewport3D.tsx`** — the scene-build effect is now keyed on geometry inputs only (`selection.selectedNode` and `collidingClampIds` removed); it keeps the **150 ms debounce + `buildRequestRef` cancellation** and stores the returned `clampMeshes`/`tabMeshes` maps in refs. A separate lightweight effect recolors those already-built meshes on selection/collision change — no rebuild. After each rebuild the current highlight is re-applied (via a ref-held latest-highlighter) so freshly-built fixtures reflect the current selection even when selection didn't change.

Deferral (gating the rebuild on the 3D tab being visible) was explicitly **not** used — it makes switching to the 3D tab slow and would need a spinner.

## Why this is airtight

Selecting a feature/clamp/tab only ever updates the `selection` slice (`selectFeature`/`selectTab`/`selectClamp` return `selection`-related partials, never a new `project`). So none of the scene-build effect's deps — `clearRenderedObjects`, `disposeObjectMaterial`, `originVisible`, `project`, `rebuildGridHelpers` (all stable across selection) — change, and `buildBooleanModel` cannot run on a selection change. Collision changes are likewise a cheap recolor now.

## Tests

`src/engine/viewport3dSelectionDecoupling.test.ts`:
- **Runtime:** `applyClampHighlight`/`applyTabHighlight` recolor an already-built fixture **in place** — same mesh + same material instance, only color/opacity change (proving the highlight is a recolor, not a rebuild); collision vs selection are visually distinct; clearing restores the original appearance.
- **Structural:** `buildScene` takes only `project` (can't consume selection/collision); the Viewport3D scene-build effect is keyed on geometry inputs only; a dedicated selection-keyed highlight effect exists.

`npm run build` (lint + tsc + all 89 test files + vite) is green.

### DIFF
--- a/src/components/viewport3d/Viewport3D.tsx
+++ b/src/components/viewport3d/Viewport3D.tsx
@@ -22,7 +22,7 @@ import type { ToolpathVisibility } from '../toolpathVisibility'
 import type { ToolpathResult } from '../../engine/toolpaths/types'
 import { useProjectStore } from '../../store/projectStore'
 import { modelFeatures } from '../../store/helpers/featureRoles'
-import { buildOriginTriad, buildScene } from '../../engine/csg'
+import { applyClampHighlight, applyTabHighlight, buildOriginTriad, buildScene } from '../../engine/csg'
 import { getStockBounds, rectProfile } from '../../types/project'
 import { getFeatureGeometryProfiles } from '../../text'
 import { buildToolpathLinePositionChunks, toolpathPointToWorldTuple } from './toolpathOverlay'
@@ -676,6 +676,10 @@ export const Viewport3D = forwardRef<Viewport3DHandle, Viewport3DProps>(function
   const controlsRef = useRef<ReturnType<typeof createOrbitControls> | null>(null)
   const frameRef = useRef<number>(0)
   const objectsRef = useRef<THREE.Object3D[]>([])
+  // The most recently built fixture meshes, keyed by id, so selection/collision
+  // highlight can recolor them in place without a full scene rebuild (issue #261).
+  const clampMeshesRef = useRef<Map<string, THREE.Mesh>>(new Map())
+  const tabMeshesRef = useRef<Map<string, THREE.Mesh>>(new Map())
   const toolpathObjectsRef = useRef<THREE.Object3D[]>([])
   const originObjectRef = useRef<THREE.Object3D | null>(null)
   const buildRequestRef = useRef(0)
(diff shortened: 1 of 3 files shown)


zabooma:
Review result: no actionable findings.

I checked the changed files (`src/components/viewport3d/Viewport3D.tsx`, `src/engine/csg.ts`, and `src/engine/viewport3dSelectionDecoupling.test.ts`) against the existing `buildScene` / fixture mesh contracts. The selection/collision state is now isolated to in-place material updates, `SimulationViewport` still uses the retained `buildClampMesh(clamp, selected, colliding)` API, and `collidingClampIds` is derived immutably from toolpath results.

Verification run in a detached PR worktree:
- `npx tsx src/engine/viewport3dSelectionDecoupling.test.ts`
- `npm test`
- `npm run build`

PR #265, opened 2026-07-07
Add creation parameter reference diagrams

### LINKED ISSUE #259: Refactor complex creation dialogs to use parameter reference diagrams
## Context

Issue #218 / PR #258 introduced the gear creation dialog with a third-column visual reference for each parameter. That pattern works well for parameters whose meaning is hard to infer from text alone, especially geometry controls like root form, fillet radius, crest form, and bore diameter.

This follow-up tracks applying the same UI pattern to other complex feature creation dialogs where parameter meaning would benefit from a compact diagram.

## Scope

Evaluate and refactor complex creation workflows such as:

- Slot creation
- Regular polygon / n-gon creation
- Rounded rectangle creation
- Chamfered rectangle creation
- Any other creation panel with non-obvious geometric parameters

Simple primitives such as plain rectangle, circle, and ellipse should only be changed if the diagram adds clear value.

## Goals

- Use the same general pattern as the gear dialog: label, control, compact visual reference column.
- Keep diagrams small, technical, and directly tied to the parameter being edited.
- Prefer dynamic inline diagrams when the drawing changes with the selected option or value.
- Avoid moving these into the global icon sprite unless a diagram is reused outside a specific dialog.
- Preserve tablet usability and existing creation/dimension-entry behavior.

## Acceptance Criteria

- Complex creation dialogs that have non-obvious geometric parameters use a consistent parameter-reference layout.
- Each new diagram clearly explains the associated parameter without adding explanatory body text to the panel.
- Existing keyboard/manual-entry flows still work.
- Existing creation previews still work.
- Focused tests are added where behavior changes; UI-only refactors should be verified with the existing build gate.
- `npm run build` passes.

## Notes

This is a UI consistency/refactor follow-up, not part of the gear feature delivery itself.

zabooma:
## Summary
- Add compact reference diagrams for regular polygon side count and rounded/chamfered rectangle corner parameters
- Move those creation fields into a focused canvas parameter-reference component
- Reuse the gear dialog reference-column styling for consistent creation panel layout

Closes #259

## Verification
- npm run build

### DIFF
--- a/src/components/canvas/CreationParameterReferences.tsx
+++ b/src/components/canvas/CreationParameterReferences.tsx
@@ -1,0 +1,185 @@
+/**
+ * Copyright 2026 Franja (Frank) Povazanj
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { KeyboardEvent, ReactNode } from 'react'
+import type { PendingAddTool } from '../../store/types'
+
+type PendingNgon = Extract<NonNullable<PendingAddTool>, { shape: 'ngon' }>
+type PendingRectCorner = Extract<NonNullable<PendingAddTool>, { shape: 'roundrect' | 'chamferrect' }>
+
+interface NgonParameterPanelProps {
+  pendingAdd: PendingNgon
+  setPendingNgonSides: (sides: number) => void
+}
+
+interface RectCornerParameterPanelProps {
+  pendingAdd: PendingRectCorner
+  setPendingRectCorner: (corner: number) => void
+}
+
+interface CreationParameterFieldProps {
+  children: ReactNode
+  label: string
+  reference: ReactNode
+}
+
+function stopPanelKey(event: KeyboardEvent<HTMLInputElement>): void {
+  event.stopPropagation()
+}
+
+function clampNgonSides(value: string, fallback: number): number {
+  const parsed = Math.round(Number(value))
+  return Number.isNaN(parsed) ? fallback : Math.max(3, Math.min(50, parsed))
+}
+
+function clampCorner(value: string, fallback: number): number {
+  const parsed = Number(value)
+  return Number.isNaN(parsed) || parsed < 0 ? fallback : parsed
+}
+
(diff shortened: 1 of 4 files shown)


zabooma (commented):
Reviewed — clean, type-safe extraction with no behavioral regressions. Two optional cleanup nits inline; nothing blocking.

src/components/canvas/CreationParameterReferences.tsx
```diff
@@ -0,0 +1,185 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { KeyboardEvent, ReactNode } from 'react'
+import type { PendingAddTool } from '../../store/types'
+
+type PendingNgon = Extract<NonNullable<PendingAddTool>, { shape: 'ngon' }>
+type PendingRectCorner = Extract<NonNullable<PendingAddTool>, { shape: 'roundrect' | 'chamferrect' }>
```
zabooma: Nit: `NonNullable<PendingAddTool>` is redundant here — `PendingAddTool` is already non-nullable (the `| null` lives on the store field `pendingAdd: PendingAddTool | null`, not on the type). `Extract<PendingAddTool, { shape: 'ngon' }>` is enough. Harmless either way.

src/components/canvas/CreationParameterReferences.tsx
```diff
@@ -0,0 +1,185 @@
...
+  return (
+    <div className="canvas-workflow-panel__meta canvas-workflow-panel__parameter-fields">
+      <CreationParameterField label="Sides (3-50)" reference={<NgonSidesReference sides={pendingAdd.sides} />}>
+        <input
+          key={pendingAdd.session}
+          className="canvas-workflow-panel__count-input"
+          type="text"
+          inputMode="numeric"
+          defaultValue={pendingAdd.sides}
+          onBlur={(event) => {
```
zabooma, line 150: Optional cleanup: the clamp → commit → write-back sequence is repeated four times (onBlur + Enter, in both this panel and `RectCornerParameterPanel`), and the two panels are structurally identical apart from the clamp fn and label. `GearParameterPanel` already has a `commitNumber(value, fallback, apply, clamp)` helper that does exactly this — reusing that pattern would collapse most of both panels. Not blocking.

src/components/canvas/CreationParameterReferences.tsx
zabooma: Resolved in `cb001fb`: removed the redundant `NonNullable` wrapper so the extracted pending types now use `Extract<PendingAddTool, ...>` directly. Verification: `npm run build` passed.

src/components/canvas/CreationParameterReferences.tsx
zabooma: Resolved in `cb001fb`: added a local `parseNumber` / `commitNumber` helper and routed both blur and Enter handlers through it for the n-gon and rectangle-corner fields. The helper keeps the previous fallback behavior for invalid side counts and negative corner values. Verification: `npm run build` passed.

Issue #270, opened 2026-07-10
Make complex SVG/DXF artwork importable as line paths or nesting-aware solids


zabooma:
## Context

A user supplied the same complex 6-inch Aztec drawing as SVG and DXF under `work/aztec vcarve test/`.

Measured from the current importers:

- SVG: 2.6 MB, 2,980 `<path>` elements.
- DXF: 11 MB, 2,980 closed polylines on one layer.
- Parsed DXF result: 2,980 closed profiles, 0 open profiles, 114,840 segments, maximum 512 segments in one profile.
- On the current development machine, DXF inspection took about 0.8 s and parsing/stitching about 1.1 s. The larger practical cost is feature creation/rendering and the 3D boolean preview.
- The SVG was exported by Vectric and all 2,980 paths are explicitly `fill:none` with a visible stroke. That is strong evidence that these are path contours, not 2,980 additive solids.

Today `importShapes` maps every closed SVG/DXF profile to `add`. That feeds thousands of solids into sequential Manifold unions, produces a misleading model, and can make the 3D preview fail. Import also selects every created feature and expands the imported folder, which is a poor large-file default.

A simple nesting pass is useful for genuinely filled/solid artwork, but it is not a universal answer for this file. A containment experiment on the DXF found one outer root, nesting depth up to 11, and a parity result of 2,753 Add / 227 Subtract. Since the SVG source is stroke-only, Paths should be recommended for that SVG. DXF Auto should still preserve the established material-oriented import default because that behavior has worked well across other tested DXF files.

Related: #208 covers evidence-backed geometry repair/diagnostics. This issue is about feature meaning, CAM eligibility, and large-import behavior rather than repairing broken contours.

## Product semantics

- **Line** may contain an open or closed profile. It is machining/path geometry and never contributes material to the 3D boolean model.
- **Add/Subtract** remain closed material-modeling features.
- **Engrave / Follow line** follows a Line profile directly at fixed depth, whether open or closed.
- **V-carve** may target Subtract features or closed Line features. Open Line features remain invalid for V-carve.
- When V-carve targets multiple closed Line features, interpret the selected Line contours as one even-odd contour set. Do not run them through the current same-winding non-zero union, which would fill nested contours and can collapse this artwork into the wrong region.
- Region and Construction semantics remain unchanged.

[...]

PR #293, opened 2026-07-14
V-Carve medial: geometric medial-axis V-carve, retiring the recursive skeleton op

### LINKED ISSUE #177: V-carve: robust geometric medial-axis solver (retire skeleton experiments)
**Priority:** High · migrated from TODO.md (root, being retired).

Implement the real robust geometric V-carve solver and retire the approximate skeleton experiments.

- Current contour-parallel V-carve works but needs tight contour spacing and produces a very busy toolpath.
- Approximate skeleton experiments are not good enough for the main workflow.
- Target: robust geometric medial-axis / straight-skeleton implementation that works on serif text and outline artwork.
- Keep the contour-parallel fallback only until the robust solver is trusted; long-term this becomes the preferred V-carve operation for outline text and artwork.

### LINKED ISSUE #279: V-Carve medial: geometric medial-axis V-carve operation (implements #177 direction)
Implements the direction of #177 (robust geometric medial-axis solver) as a **new** operation kind `v_carve_medial` ("V-Carve medial"). The existing `v_carve` (offset) and `v_carve_recursive` (skeleton) generators are untouched; deciding their fate is the open question that keeps this issue open after the implementation lands.

Implementation lives on branch `vcarve-skeleton-redesign-0ab6ba` in `src/engine/toolpaths/vcarveMedial/`.

## Design

Instead of approximating the skeleton by Clipper offset-stepping with topology-event heuristics, the medial axis is computed directly from the Voronoi diagram of a dense boundary sampling (the Delaunay-dual construction):

1. **Sample** every boundary loop (outer + islands) at `stepover` spacing, preserving original vertices exactly so corners stay sharp.
2. **Triangulate** the samples with `delaunator` (new dependency; tiny, robust). Circumcenters of triangles that fall inside the region converge to the interior medial axis; adjacent interior triangles yield the axis edges (midpoint-inside guard for reflex pockets).
3. **Exact clearance**: each node's clearance is the exact distance to the original boundary segments (uniform-grid segment index with expanding-ring queries), not to the samples — depth accuracy is independent of the sampling step.
4. **Contact-spread filter** (λ-medial-axis criterion evaluated against exact geometry): keep a node only if the boundary stays within `clearance·(1 + tol)` over an angular spread ≥ ~11°, with `tol = 0.003 + (h/c)²/8` covering chord-sagitta discretization error. Vertices introduced by curve flattening (circles, fillets, font beziers) spawn spokes that see the boundary in one narrow direction only — filtered out; genuine corners (the sharper, the wider the spread) and osculating tips (rounded stroke terminals) always survive.
5. **Cluster contraction**: nodes merged on a 0.75·step grid, each cluster represented by its **deepest** member (the true medial axis locally maximizes clearance, and its exact clearance carries over). De-braids parallel micro-chains; cell-based clustering bounds the merge diameter so long chains can never collapse transitively.
6. **Spur pruning** (short non-descending leaf edges) and **small-component removal** (the medial axis of a connected region is connected; tiny leftover fragments are discretization remnants).
7. **Corner tips**: branch tips extended to every convex corner with turn > 15° at zero clearance, so the V-bit surfaces exactly at the corner point (crisp serifs). Convexity is evaluated region-side (outer loops: convex vertices; islands: their concave vertices).
8. **Dot collapse**: a skeleton whose entire extent is ≤ 3 steps (an "i" dot, a period) becomes a single plunge at the deepest point.

### Depth model

`depth = clearance / tan(vBitAngle/2)`, clamped to `maxCarveDepth` and the band height — the V flanks touch both walls everywhere along the skeleton. Bands/regions come from the shared `resolvePocketRegions` resolver, so feature targeting, islands, region masks, tabs and clamps behave like the other v-carve ops.

### Toolpath emission

- Graph → chains (junction-to-junction paths, cycles, plunge points), simplified with a depth-aware Douglas-Peucker (deviation = max of XY error and Z error scaled by the flank slope; preserves the max-depth clamp kink).
- **Coverage pruning**: a leaf branch is dropped when every point's groove protrudes at most ~¾ step (in clearance units) below the surface the remaining skeleton already cuts — from the identity “groove at q covers the tip at p ⇔ dist(p,q) ≤ q.clearance − p.clearance”. Removes terminal micro-branches that “do nothing”; corner branches survive because clearance falls steeply along them; zero-clearance tips are exempt.
- **Medial-ball linking**: chain ends within 90% of both endpoints' clearance are joined tool-down (both points lie inside the groove opening already carved there), plus a 3-step budget for same-or-deeper entries; never from safe height. Greedy nearest-chain ordering with junction continuation.

### Parameters (no schema change)

- `stepover` — skeleton sampling step (UI label "Step Size"); also scales all cleanup tolerances
- `maxCarveDepth` — depth clamp for wide areas
- `debugToolpath` — tags cut moves `source: 'medial-axis'`
- Per-region sample budget (~40k) coarsens the step with a warning; empty results auto-refine at half step twice (features thinner than the step).

## Included bugfix (affects existing ops too)

`expandFeatureGeometry` leaked the parent's `definitionId`/`transform` reference fields onto synthetic text glyph children via object spread. `resolvePocketRegions` treats any feature carrying a `definitionId` as an instance to resolve by id — impossible for synthetic ids — and flattened **every glyph to an empty path**, so raw text-feature targets produced "Band … resolved to empty subject geometry" for *all* pocket-like operations (`pocket`, `v_carve`, `v_carve_recursive` alike). The synthetic children now strip those fields (their profiles are already fully resolved world space). Regression-tested.

## Results (60° V-bit, step 0.25 mm / 0.01")

| Case | v_carve_recursive | v_carve_medial |
|---|---|---|
| Serif text "Rag" (Droid Serif Bold) | 6,813 cuts · 22.5 s | ~200 cuts · 0.17 s |
| 5-point star | 187 cuts · 59 s | 10 cuts · 11 ms |
| Ring | 96 cuts · 2.0 s | 63 cuts · 23 ms |
| "Frankie Doo goes to Bancroft" (Droid Sans Bold Italic, 25 regions) | — | 985 cuts · 94 retracts · ~1 s |

Old output showed fragmented bowls ('g'), doubled strokes and stray cuts; the new skeletons are connected, single-stroke, contained, with corner tips reaching the surface.

## Testing

- `vcarveMedial/vcarveMedial.test.ts` (16 cases): exact-clearance property, corner tips at zero depth, reflex/island corner behavior, junction topology, circle spoke filtering, ring with island, 60°/90° depth math, max-depth clamp, determinism, multi-feature separation, raw-text regression, dot collapse, link-vs-retract behavior.
- Extended `camOperationSmoke` (generate + post G-code), `toolSelection`, `constructionExclusion` kind coverage.
- New Playwright e2e: create V-Carve (medial) via the feature context menu with auto-imported V-bit.
- `scripts/vcarve-medial-preview.ts` renders side-by-side old/new SVG comparisons.

## Remaining before this is "done"

- [ ] **Decide the fate of the other two variants** — keep `v_carve` (offset) as-is? deprecate/remove `v_carve_recursive` (per #177 "retire skeleton experiments")? migration for saved projects using them?
- [ ] Example image `public/operation-examples/vcarve-medial-example.png` for the add-operation card
- [ ] In-app 3D preview / simulation sanity pass on real projects
- [ ] Future: optional flat-area clearing when depth clamps (wide regions currently bottom out along the centerline only); multi-pass depth staging; smarter chain ordering (Eulerian) to cut the remaining retracts

zabooma:
## Summary

Replaces the approximate skeleton V-carve experiments with a **robust geometric medial-axis solver** (issue #177), delivered as a new operation kind **`v_carve_medial`** ("V-Carve medial"), and **retires `v_carve_recursive`** in its favor. `v_carve` (offset) is kept as the area-clearing variant.

Full design writeup and the decision record are in #279.

## The new operation

Instead of approximating the skeleton by Clipper offset-stepping with topology-event heuristics, the medial axis is computed directly from the Voronoi diagram of a dense boundary sampling (the Delaunay-dual construction):

- **Exact depth** — each skeleton node's clearance is the exact distance to the original boundary segments (not to the samples), so `depth = clearance / tan(halfAngle)` makes the V flanks touch both walls everywhere. Independent of sampling step.
- **Contact-spread filter** (λ-medial-axis criterion on exact geometry) removes the spokes that curve-flattening spawns, so circles, fillets and font béziers stay clean while genuine corners and rounded stroke terminals survive.
- **Crisp corners** — branch tips extend into every convex corner >15° at zero depth, so the bit surfaces exactly at the point (sharp serifs by construction).
- Cleanup passes: deepest-member cluster contraction (de-braids parallel micro-chains), coverage-based redundant-leaf pruning, medial-ball chain linking (minimizes retracts), and dot-collapse (an "i" dot becomes a single plunge).

`stepover` is the skeleton sampling step (UI "Step Size"); `maxCarveDepth` clamps wide areas.

### Results (60° V-bit, equal step size)

| Case | old `v_carve_recursive` | `v_carve_medial` |
|---|---|---|
| Serif "Rag" (Droid Serif Bold) | 6,813 cuts · 22.5 s | ~200 cuts · 0.17 s |
| 5-point star | 187 cuts · 59 s | 10 cuts · 11 ms |
| "Frankie Doo goes to Bancroft" (25 regions) | — | 985 cuts · ~1 s |

The old output fragmented bowls ('g'), doubled strokes and left stray cuts; the new skeletons are connected, single-stroke, contained, with corner tips reaching the surface.

## Removing `v_carve_recursive`

- Deleted the generator + its test; removed the kind from `OperationKind` and every wiring site (defaults, descriptions, CAMPanel add-menu + property editors, validity/quick-ops, tool selection, booklet, region resolver).
- **Saved projects migrate on load**: `normalizeOperation` rewrites `kind: 'v_carve_recursive'` → `'v_carve_medial'`. Parameters are compatible (`stepover` is the step size for both; `maxCarveDepth` and the V-bit carry over). The bundled `purecutcnc.camj` example op was updated in place.
- Because #270 already made closed Line features valid V-carve targets, `v_carve_medial` inherits closed-Line support (smoke/resolver/validity tests converted from the recursive kind).

Net diff across the branch: **−1,489 lines**.

## New dependency

`delaunator` (small, widely-used, robust Delaunay triangulator).

## Testing

- New `vcarveMedial/vcarveMedial.test.ts` (16 cases): exact-clearance property, corner tips at zero depth, reflex/island behavior, junction topology, circle spoke filtering, ring-with-island, 60°/90° depth math, clamp, determinism, dot-collapse, chain-linking, raw-text target.
- New `store/operationMigration.test.ts`: recursive→medial load migration + param preservation.
- Extended `camOperationSmoke`, `toolSelection`, `constructionExclusion`, `operationValidity`, `operationDefaults` for the kind change.
- New e2e: create V-Carve (medial) via the feature context menu with an auto-imported V-bit.
- `npm run build` green (lint + tsc + 112 test files + vite); `npm run test:e2e` 31/31.

## Follow-ups (tracked on #279, not blocking)

- Add `public/operation-examples/vcarve-medial-example.png` for the add-operation card (the retired skeleton PNG can be repurposed).
- In-app 3D preview / simulation sanity pass on a real project.
- Optional, future: flat-area clearing when depth clamps; multi-pass depth staging.

Closes #279
Closes #177


zabooma:
Review findings:

1. **[P1] Tool-down chain links can gouge clamped V-carves** — [`toolpath.ts:310-314`](https://github.com/PureCutCNC/purecutcnc/blob/73a65b7f55855cd1f29b86040d28c23a983c3460/src/engine/toolpaths/vcarveMedial/toolpath.ts#L310-L314) checks the raw medial clearance when deciding whether to bridge two chains, but `depthOf()` caps the actual cutting depth. For example, with clearance 1, slope 1, max depth 0.1, and a 0.8 gap, the raw check permits a bridge while each V-groove reaches only 0.1 sideways, so the emitted link cuts across unplanned material. The `linkBudget` alternative has the same issue: entering at an equal/deeper Z does not show that the segment was already machined. Use capped effective clearance and a swept-volume containment check, or retract.

2. **[P2] The advertised preview script cannot run** — [`scripts/vcarve-medial-preview.ts:32`](https://github.com/PureCutCNC/purecutcnc/blob/73a65b7f55855cd1f29b86040d28c23a983c3460/scripts/vcarve-medial-preview.ts#L32) imports `vcarveRecursive`, which this PR deletes. Running its documented command fails with `ERR_MODULE_NOT_FOUND`. Remove the script or rewrite the comparison around saved output rather than the retired generator.

3. **[P2] The new operation card displays a missing-image fallback** — [`operationDescriptions.ts:70`](https://github.com/PureCutCNC/purecutcnc/blob/73a65b7f55855cd1f29b86040d28c23a983c3460/src/types/operationDescriptions.ts#L70) names `vcarve-medial-example.png`, but that asset is absent from `public/operation-examples/`. Expanding the add-operation card therefore visibly shows “Missing image.” Add the asset in this PR or temporarily reuse an existing one.

Verification: the focused medial suite and `npm run build` passed on this PR head; GitHub build and e2e checks are green.

zabooma:
Thanks for the review — addressed in `9585fe9`:

**1. [P1] Chain-link gouge — fixed.** You're right that the check used the raw medial clearance while `depthOf()` clamps the actual cut depth. Links now test against the **effective carved radius** `min(clearance, maxDepth·slope)` — the V-bit cone's surface footprint — so a bridge is permitted only when the whole segment stays inside the disk the tool has already carved at its current point. Your exact example (clearance 1, slope 1, maxDepth 0.1, gap 0.8) now retracts: effective radius 0.1, link reach 0.09 < 0.8. I also **dropped the separate z-based `linkBudget` path** you flagged (equal/deeper-Z doesn't prove the segment was machined) — linking is now gated solely by that containment test. Added `testLinkRespectsClampedDepth` encoding the scenario. Real text is unaffected (its junctions aren't depth-clamped): the test file still generates 985 cuts / 0 warnings.

**2. [P2] Broken preview script — removed.** Rather than rewrite it, I deleted `scripts/vcarve-medial-preview.ts` — its whole purpose was the old-vs-new comparison, which is moot now that the recursive generator is gone. While there I found **52 more debug scripts** that imported the deleted `vcarveRecursive` generator (same `ERR_MODULE_NOT_FOUND`); those are the retired skeleton experiments (#177), so I removed them too. None was a build/test/lint gate.

**3. [P2] Missing example PNG.** The repo owner is adding `vcarve-medial-example.png` separately; leaving that one to them.

`npm run build` is green (lint + tsc + 112 tests + vite).

zabooma:
Re: finding #3 (missing example image) — handled in `c91632a`. The retired skeleton op's `vcarve-skeleton-example.png` was orphaned by its removal and happens to depict exactly the medial-axis toolpath (a letter T: centerline branches meeting at a junction, tips into the corners), so I renamed it to `vcarve-medial-example.png` (a `git mv`, tracked as a rename). The add-operation card now shows a real image instead of the fallback. All three findings resolved.

PR #342, opened 2026-07-21
fix: closed-line v-carve ignores enclosing add islands and breaks even-odd fill in feature_first mode

### LINKED ISSUE #340: v-carve offset and medial do not produce toolpath for the closed lines
v-carve offset and medial  accept closed lines as target but do not produce toolpath for the closed lines.

Image

Image

To reproduce: create a triangle line shape and add v-carve operations to it.

zabooma:
Two fixes for issue #340:

**1. resolver.ts** — Before subtracting an add island from line areas, check whether the add fully encloses every active line target. If it does, the add is parent material the line carves into — skip subtraction. If it doesn't, the add is a true island inside the line area — subtract it.

**2. multiFeature.ts** — `isFeatureFirst` returns false for v-carve operations that include line targets, so line-line even-odd fill is preserved. Previously `feature_first` mode split the two circles of a letter-O into separate per-feature sub-operations, losing the even-odd hole.

### Verification
- Full build passes
- All 26 unit tests pass (14 resolver + 12 smoke)
- All 68 e2e tests pass
- User's test file `v-carve-closed-line-test.camj` now produces correct donut toolpaths (66 medial moves, 2344 offset moves)

Closes #340

### DIFF
--- a/src/engine/toolpaths/multiFeature.ts
+++ b/src/engine/toolpaths/multiFeature.ts
@@ -61,21 +61,19 @@ export function isFeatureFirst(operation: Operation, project?: Project): boolean
   if ((operation.machiningOrder ?? 'level_first') !== 'feature_first') return false
   if (operation.target.source !== 'features') return false
   if (operation.target.featureIds.length <= 1) return false
-  // V-carve operations that target line features must be processed together
-  // — line-line even-odd fill cannot work when targets are split across
-  // separate per-feature sub-operations (issue #340).
+  // V-carve operations that target a closed line must be processed together
+  // — line-line even-odd fill cannot survive when targets are split across
+  // separate per-feature sub-operations (issue #340). This applies only when
+  // a line target is actually present: other v-carve targets (e.g. multiple
+  // disjoint subtracts) still split per feature as usual.
   if (operation.kind === 'v_carve' || operation.kind === 'v_carve_medial') {
-    if (project) {
-      const hasLineTarget = operation.target.featureIds.some((id) => {
-        const feature = project.features.find((f) => f.id === id)
-        if (!feature) return false
-        const definition = project.featureDefinitions[feature.definitionId]
-        return definition?.operation === 'line'
-      })
-      if (hasLineTarget) return false
-    }
-    // Without project we can't check — be safe and don't split.
-    return false
+    // Without a project we cannot inspect the targets — be safe, don't split.
+    if (!project) return false
+    const featuresById = resolvedFeatureMap(project)
+    const hasLineTarget = operation.target.featureIds.some(
+      (id) => featuresById.get(id)?.operation === 'line',
+    )
+    if (hasLineTarget) return false
   }
   return true
 }
--- a/src/engine/toolpaths/multiFeature.ts
+++ b/src/engine/toolpaths/multiFeature.ts
@@ -57,10 +57,27 @@ export function perFeatureOperations(operation: Operation, project?: Project): O
   }))
 }
 
-export function isFeatureFirst(operation: Operation): boolean {
+export function isFeatureFirst(operation: Operation, project?: Project): boolean {
   if ((operation.machiningOrder ?? 'level_first') !== 'feature_first') return false
   if (operation.target.source !== 'features') return false
-  return operation.target.featureIds.length > 1
+  if (operation.target.featureIds.length <= 1) return false
+  // V-carve operations that target line features must be processed together
+  // — line-line even-odd fill cannot work when targets are split across
+  // separate per-feature sub-operations (issue #340).
+  if (operation.kind === 'v_carve' || operation.kind === 'v_carve_medial') {
+    if (project) {
+      const hasLineTarget = operation.target.featureIds.some((id) => {
+        const feature = project.features.find((f) => f.id === id)
+        if (!feature) return false
+        const definition = project.featureDefinitions[feature.definitionId]
+        return definition?.operation === 'line'
+      })
+      if (hasLineTarget) return false
+    }
+    // Without project we can't check — be safe and don't split.
+    return false
+  }
+  return true
 }
 
 function mergeBounds(a: ToolpathBounds | null, b: ToolpathBounds | null): ToolpathBounds | null {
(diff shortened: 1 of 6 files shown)


src/engine/toolpaths/multiFeature.ts
```diff
@@ -57,10 +57,27 @@ export function perFeatureOperations(operation: Operation, project?: Project): O
...
+      const hasLineTarget = operation.target.featureIds.some((id) => {
+        const feature = project.features.find((f) => f.id === id)
+        if (!feature) return false
+        const definition = project.featureDefinitions[feature.definitionId]
+        return definition?.operation === 'line'
+      })
+      if (hasLineTarget) return false
+    }
+    // Without project we can't check — be safe and don't split.
+    return false
```
zabooma: 🔴 **Dead code + broader-than-intended behavior change.**

`if (hasLineTarget) return false` is immediately followed by an unconditional `return false`, so `hasLineTarget` never affects the result — and **every** `v_carve`/`v_carve_medial` op with >1 target now returns `false`, even with **zero** line targets.

`machiningOrder: 'feature_first'` is the default for *all* operations (`operationDefaults.ts:320`), so this hits the common path. `testVCarveDisjointFeaturesAreMachiningOrderInvariant` (`toolpaths.test.ts:1587`) targets two disjoint **`subtract`** features (no lines): it used to split per-feature, now it never does. The test still passes but only *trivially* — both modes now take the identical non-split path — silently voiding what it was written to guard. This also contradicts the comment / PR description ("operations that include line targets").

Fix — gate the early return so it fires only with a line target (or no project), else fall through to `return true`:
```ts
if (operation.kind === 'v_carve' || operation.kind === 'v_carve_medial') {
  if (!project) return false // can't check — be safe, don't split
  const hasLineTarget = operation.target.featureIds.some((id) => {
    const f = project.features.find((x) => x.id === id)
    return f ? project.featureDefinitions[f.definitionId]?.operation === 'line' : false
  })
  if (hasLineTarget) return false
}
return true
```

If disabling per-feature splitting for *all* v-carve is actually intended, drop the dead `hasLineTarget` block and update the comment + PR description to say so.

Minor: `perFeatureOperations` above resolves features via `resolvedFeatureMap(project)`; consider the same here instead of `features.find` (O(1) lookup + consistent with how this module resolves features, including instanced ones).

---

src/engine/toolpaths/resolver.ts
```diff
@@ -404,9 +404,19 @@ export function resolvePocketRegions(authoritativeProject: Project, operation: O
...
         if (lineAreas.length > 0) {
-          lineAreas = differencePaths(lineAreas, [flattenFeatureToClipperPath(island.feature)])
+          const islandPath = flattenFeatureToClipperPath(island.feature)
+          const islandEnclosesAllLines = activeLineTargetsForBand.every(({ feature }) => {
+            const targetPath = flattenFeatureToClipperPath(feature)
+            return differencePaths([targetPath], [islandPath]).length === 0
+          })
+          if (!islandEnclosesAllLines) {
+            lineAreas = differencePaths(lineAreas, [islandPath])
+          }
```
zabooma: Two notes on this enclosure check:

1. 🟡 **All-or-nothing across multiple lines.** With more than one line target, if the add encloses *some* but not all of them, `.every()` is `false` and the add is subtracted from the **whole** `lineAreas` union — erasing the area of the line(s) it *does* enclose. (No worse than before for that specific combo, but the fix could decide per-line-target: subtract the add only from the lines it doesn't enclose.)

2. 🟡 **Redundant geometry recompute.** `linePaths` was already built a few lines above from this same `activeLineTargetsForBand`. Re-flattening each line target for every island here is O(islands × lines) `flattenFeatureToClipperPath` calls — zip with the precomputed `linePaths` instead.

---

src/engine/toolpaths/vcarveLineResolver.test.ts
```diff
@@ -154,47 +154,45 @@ const wtext = (w: { code: string; params?: Record<string, unknown> }): string =>
...
-  const bottomBand = result.bands.find((b) => b.topZ === 2 && b.bottomZ === 0)
-  assert(bottomBand !== undefined, 'bottom band 2→0 must exist')
-  assert(bottomBand.islandFeatureIds.includes('add1'),
-    `bottom band must discover add1 as island, got islands: ${bottomBand.islandFeatureIds.join(', ')}`)
-  // Machinable area = 10×10 inner minus 5×5 island = 100 - 25 = 75
-  const bottomArea = bottomBand.regions.reduce((sum, r) => sum + regionArea(r), 0)
-  assert(approx(bottomArea, 75),
-    `bottom band area should be ~75, got ${bottomArea}`)
-})
+	test('Z-span isolation: add inside line is a true island, add enclosing line is not (issue #340)', () => {
```
zabooma: 🟠 **The new behavior isn't actually tested.** The title and comments now advertise the enclosing case ("add enclosing line is not [an island]", "Contrast with an add that ENCLOSES the line → parent material"), but the body still only asserts the pre-existing "add **inside** line → island → subtracted" path — the same thing the old test checked (this is a rename/re-comment of the unchanged test). The new `islandEnclosesAllLines === true` branch in `resolver.ts` — the actual fix — is never exercised.

Please add a case where an `add` **fully encloses** the line target and assert the line area is preserved (add *not* subtracted, no `bandEmptySubject` warning, expected area).

Nit: this block is indented with **tabs** while the rest of the file uses 2-space indent (eslint has no indent rule, so it passes, but it's inconsistent — re-indent with spaces).

---

zabooma (commented):
## Review — closed-line v-carve enclosing-add fix (#340)

Well-targeted fix and the enclosure heuristic reads clearly. Two items worth resolving before merge, plus minor cleanups — details are in the inline comments.

- 🔴 **`isFeatureFirst` over-reach + dead code** (`multiFeature.ts`) — the v-carve branch returns `false` for *all* v-carve ops, not just line ones, and `hasLineTarget` is dead code (an `if (hasLineTarget) return false` directly above an unconditional `return false`). Since `feature_first` is the default machining order, this changes the common path and silently voids the existing `testVCarveDisjointFeaturesAreMachiningOrderInvariant` test (two disjoint `subtract` features, no lines — it no longer splits). This also contradicts the PR description's "operations that include line targets".
- 🟠 **Core fix is untested** (`vcarveLineResolver.test.ts`) — the enclosing-add → parent-material branch (`islandEnclosesAllLines === true`) has no assertion; the renamed test only re-covers the old inside-island case.
- 🟡 **`.every()` all-or-nothing + redundant recompute** (`resolver.ts`) — enclosure decision is global across all line targets, and `linePaths` is re-flattened per island instead of reusing the value computed a few lines above.

Build, existing unit tests, and the donut e2e all look good. From my side #1 and #2 are the blockers; #3 is a smaller correctness/perf follow-up.

_(Posted as a comment rather than "request changes" since GitHub doesn't allow requesting changes on one's own PR.)_

---

PR #355, opened 2026-07-23
Optimize linear toolpaths and add export-stage arc fitting

### LINKED ISSUE #354: Optimize linear toolpaths and add export-stage native arc fitting
## Summary

Reduce G-code size and controller motion overhead in two deliberately separate layers:

1. **Always-on straight-line optimization**: canonicalize generated linear toolpath moves so consecutive, collinear moves become one move.
2. **Per-operation native arc fitting**: when enabled (default **on**) and the selected machine explicitly supports it, recognize suitable constant-Z cutting polylines during G-code export and emit G2/G3. Otherwise, retain the existing linear G1 output.

This issue is the implementation plan. No code should be started until it is reviewed and explicitly approved.

## Analysis and architectural decision

The two optimizations have different owners.

| Concern | Owner | Why |
| --- | --- | --- |
| Consecutive straight-line reduction | Toolpath generation / finalization | It is controller-independent and should make the cached toolpath, 2D/3D preview, simulation, and every export smaller in the same way. |
| Arc fitting and G2/G3 emission | G-code export | It depends on the selected controller, coordinate transform, arc syntax, numeric formatting, and machine capability. The preview and simulation should keep consuming the canonical linear toolpath. |

Current code supports this split:

- `ToolpathMove` is linear-only (`rapid`, `plunge`, `cut`, `lead_in`, `lead_out`) and a `ToolpathResult` is the object shared by preview, simulation, and export.
- The postprocessor currently walks those moves as linear motion. Machine definitions already name clockwise/counter-clockwise arc commands and an `ij` or `r` format, but the model has no native arc move contract or explicit arc-support capability.
- `arcReconstruction.ts` reconstructs sketch/Clipper geometry, including a source-concentric Kasa/RDP path. It is useful reference material, but it must not be fed directly into export: generic G-code fitting needs independent residual, sweep, endpoint, and controller-capability checks.
- `Operation` already carries persistent per-operation settings with defaulting and normalization. An export-only arc-fitting setting must intentionally stay out of the toolpath-computation cache equality, so toggling it does not regenerate geometry.

The requested placement is therefore correct: line optimization belongs in toolpath generation; arc fitting belongs in export, with a persisted operation-level preference.

### 1. Add an always-on linear-move finalization pass

- Add a pure shared toolpath finalizer that removes zero-length duplicates and merges only adjacent moves that are:
  - contiguous;
  - the same move kind, feed-scale, and diagnostic/source boundary;
  - at the same Z;
  - direction-preserving and collinear within a conservative numeric epsilon.
- Never merge across rapid/plunge/cut/lead boundaries, feed changes, Z changes, corners, reversals, tabs, operation boundaries, or clamp/diagnostic boundaries.
- Run it once after a strategy has finished constructing its moves and at the point where any move-indexed clamp metadata is either not yet calculated or is recalculated against the finalized array.
- Preserve warnings, bounds, source tags, and collision reporting. Preview, simulation, export, and cached results then consume the same reduced linear path.
- Apply it unconditionally to all operation kinds; it introduces no user option and does not relax geometry tolerance.

### 2. Add the persisted operation preference

- Add `arcFittingEnabled?: boolean` to `Operation`, default it to `true` for new operations, and normalize missing legacy values to `true`.
- Surface an operation checkbox such as **Arc fitting (G2/G3 where supported)** in the CAM operation settings, with clear copy that it changes exported G-code, not the displayed/simulated path.
- Update typed localization, operation defaults/normalization, serialization fixtures, and the operation-comparison tests.
- Treat the setting as export-only: it must not invalidate the cached/generated toolpath.

### 3. Define explicit controller arc capability

- Extend the machine-definition schema with an explicit native-arc capability and its supported format/limits rather than inferring support merely because command strings exist.
- Verify each bundled definition before opting it in. Custom or legacy definitions without the explicit capability must safely retain linear output.
- Keep the existing `ij` / `r` format contract, adding any required constraints for unambiguous sweeps and full circles.
- When a selected operation requests arc fitting but the machine cannot use it, export the original linear moves and show an actionable non-blocking warning.

### 4. Fit and emit arcs only in the export pipeline

- Introduce a pure export-side fitting stage between project-to-machine-coordinate transformation and postprocessor motion formatting. It must not mutate `ToolpathResult`.
- Consider only contiguous, constant-Z cutting runs with one effective feed. Exclude rapids, plunges, Z-changing/helical moves, drilling cycles, and unsafe/ambiguous lead or transition moves in the first version.
- Fit a circle only when every candidate point meets a conservative chord/residual tolerance, its sweep is valid and direction-consistent, and the formatted endpoint/center data remain representable.
- Split full circles and long or ambiguous arcs into conservative sub-arcs appropriate for the selected `ij` or `r` controller format.
- Derive G2/G3 direction and I/J or R data after the export coordinate transform so the existing Y inversion, machine origin, and axis mapping cannot reverse an arc accidentally.
- Fall back per run to unchanged G1 moves when any validation fails; never approximate a questionable path as an arc.

### 5. Verification and documentation

- Add focused unit coverage for the linear finalizer: duplicate removal, collinear reduction, tolerance boundary, corner/reversal preservation, move-kind/feed/Z/source boundaries, and correct clamp-index handling.
- Add export fixtures for CW and CCW arcs, both supported arc formats, full-circle splitting, Y-down-to-machine conversion, legacy/unsupported-machine linear fallback, disabled-per-operation fallback, and unchanged rapid/plunge/Z-changing behavior.
- Cover defaulting/normalization and ensure changing `arcFittingEnabled` does not cause toolpath-cache invalidation.
- Add or extend a browser E2E check for the operation setting and export warning/preview wiring.
- Update `planning/G-code_Export_Design.md` with the canonical-linear-toolpath/native-arc-export boundary.
- Run focused tests, `npm run build`, and the relevant export UI E2E test.

## Acceptance criteria

- Every generated toolpath uses the same geometry but contains fewer moves whenever adjacent straight segments are safely reducible; no operation setting is required.
- No simplification crosses a semantic, feed, Z, corner, reversal, tab, or diagnostic/clamp boundary.
- New and legacy operations have arc fitting enabled by default, and the user can turn it off for an individual operation without regenerating its toolpath.
- On an explicitly arc-capable selected machine, a qualifying constant-Z cut emits correct G2/G3 output with deterministic formatting; its exported endpoint matches the original linear run.
- On unsupported/custom machines, disabled operations, failed fits, drilling, rapids, plunges, or Z-changing runs, export remains safe linear output and reports capability fallback where applicable.
- 2D preview, 3D preview, and simulation continue to use the canonical linear toolpath and do not depend on controller-specific arc behavior.

## Out of scope

- Helical G2/G3, G18/G19 planes, and fitting rapids/plunges/drilling cycles.
- Changing simulation or preview data to native arc primitives.
- A user-configurable fit/chord tolerance. The related backlog note in #196 remains a separate future UX decision; this issue uses a conservative tested internal tolerance.

## Risks

- A false-positive fit can change the machined path. Mitigation: residual/sweep/format checks, conservative splitting, and per-run linear fallback.
- Controllers differ on full-circle and R-format behavior. Mitigation: explicit capability, small unambiguous sub-arcs, and controller-specific fixtures.
- Move-indexed clamp diagnostics can become stale after merging. Mitigation: finalize before indexing or recompute the index metadata from the final move array.

## Approval checkpoint

After approval, move this item to **Ready** and implement on a dedicated issue branch.

zabooma:
## Summary

- Apply an always-on, exact linear-move optimizer to generated toolpaths.
- Add a default-on per-operation arc-fitting preference and explicit machine arc-interpolation capability.
- Fit only contiguous, constant-Z circular cut runs after machine-coordinate transformation; emit modal G2/G3 in I/J or R format, split arcs at 90 degrees, and retain linear fallback plus warning when unsupported.
- Add coverage for line optimization, operation defaults/migration, controller capability, split-arc I/J offsets, transformed direction, disabled/unsupported fallback, and move-count preservation.

## Verification

- `npm run build`
- No separate browser smoke: the one CAM checkbox uses the existing controlled operation-update path; its default, cache, and export behavior are covered by structural tests.

Closes #354


zabooma:
Added the safety correction from the fixture report.

The fitter now rejects a candidate when one angular chord step is disproportionately larger than the median step. This targets the false-positive shape in `work/tool-optimization-test.camj`: a long straight G1 followed by tiny corner fragments can match a huge-radius circle at its vertices, but it is not a real arc. Ordinary evenly flattened arcs remain eligible.

Verification:
- `npx tsx src/engine/gcode/arcFitting.test.ts`
- `npx tsx motionDebugFixtureCheck.ts` against the supplied fixture
- `npx playwright test e2e/motionDebug.smoke.spec.ts`
- `npm run build` (135 tests + production build)

Issue #359, opened 2026-07-24
Toolpath generator tessellation resolution degrades when circle is broken by island


zabooma:
## Problem

When a full-circle pocket is broken by an injected island (becoming a C-shape), the toolpath generator produces lower-resolution tessellation — fewer and longer linear segments approximating the curve. The arc fitting then works with this coarser input, producing arcs that deviate beyond the configured tolerance.

## Root cause

- **Full circle**: high-resolution tessellation (many short linear segments) → arc fitting fits tightly → small deviations within tolerance
- **Broken circle (C-shape)**: lower-resolution tessellation (fewer/longer linear segments) → arc fitting fits loosely → larger deviations exceeding tolerance

The generator's tessellation resolution is not consistent between full circles and broken circles. When an island breaks the circle, the resulting path segments are longer and coarser, which the arc fitting cannot compensate for.

## Observed behavior

The exported-motion debug view (issue #356) flags arc deviations of 0.0005–0.0014 inches at Z=0.3500 for the broken circle segments, while the full circle segments show no deviations. The tolerance is 0.0004 inches.

## Expected behavior

Tessellation resolution should be consistent regardless of whether the circle is full or broken by an island, so arc fitting deviations stay within tolerance across all path segments.

## Affected area

Toolpath generator tessellation for pocket operations with islands.

zabooma:
## Implementation plan

### Problem framing (confirmed from code)

The exported-motion debug diagnostic (`compareExportedVsOptimized` in `src/engine/gcode/motionDebug.ts`) measures the planar distance from each exported G2/G3 arc back to the **optimized linear toolpath** — i.e. the chord-to-arc sagitta of the polyline the arc replaced. So "arc deviation exceeds tolerance" is literally *the generator's chord sagitta*, not an arc-fit error.

- Full circle → generator emits a dense chord polyline → small sagitta → fitted arc stays within tolerance. ✅
- Broken circle (island injected) → generator emits **coarser** chords → larger sagitta → fitted arc deviates beyond tolerance. ❌

So the fix is purely a **generator tessellation-density** problem: the chord density on the curved wall portion must be the same whether or not an island breaks the circle, and it must be driven by the configured arc-fit/export tolerance, not by island presence or the offset delta.

### Root-cause candidates to confirm empirically

[...]

PR #360, opened 2026-07-24
feat: add exported-motion debug view (issue #356)

### LINKED ISSUE #356: Add a single-operation 2D exported-motion debug view for 2.5D toolpaths
# Add a single-operation 2D exported-motion debug view for 2.5D toolpaths

## Problem

The current preview and simulation consume the canonical generated toolpath, while G-code export can apply straight-line reduction, arc fitting, coordinate transformation, and postprocessor formatting afterward. That makes it difficult to verify that the motion written to the `.nc` file still represents the intended path without leaving PureCutCNC for a separate simulator.

Issue #354 established optimization at the generator/export boundaries. This follow-up provides a focused visual inspection tool for the resulting motion.

## First-version product contract

Add an **Inspect exported motion** action to the Export dialog when exactly one eligible operation is selected. It opens a dedicated debugger that shows a flat, planar view of that operation only.

The initial implementation supports generated operations whose cutting motion is organized into discrete, constant-Z passes. It is intended for conventional 2.5D paths such as pockets, inside/outside edge routes, and constant-depth follow-line work. It deliberately excludes operations with continuously varying cutting Z, including surface roughing/finishing and V-carve, as well as drilling where a plan-view trace would not be useful.

Eligibility is ultimately determined from the generated motion rather than only an operation-name allow-list: the operation must produce a non-empty planar cutting trace with discrete cutting Z levels. If it does not, the action explains why it is unavailable.

This is a diagnostic viewer, not a replacement for material-removal simulation or a general NC-file viewer.

## User experience

The debugger opens from the same export run and uses the exact generated G-code text for the selected operation. The primary view is an SVG plan view in project coordinates, initially fit to the visible trace. It offers fit/reset, pan, and zoom controls.

Controls:

- **Cutting level:** defaults to the first cutting Z encountered in machine order. Users can switch among discrete cutting levels or choose **All cutting levels**.
- **Layers:** independently toggle the following paths. They are all initially visible, with contrasting colors and opacity so differences remain legible.
  - **Generated:** raw toolpath before adjacent-collinear linear moves are merged.
  - **Optimized:** canonical toolpath after always-on straight-line optimization, before export arc fitting.
  - **Exported G-code:** the path reconstructed from the literal G-code that will be saved, including G0/G1/G2/G3 motion.
- **Non-cutting moves:** hidden by default; when enabled, show rapids/plunges/retracts as a subdued dashed layer.

The sidebar reports raw vs optimized linear-move counts and the number removed, plus emitted motion counts (linear, rapid, clockwise arc, counter-clockwise arc). This makes straight-line optimization observable without needing to infer it from a rendered overlap.

The bottom diagnostic status reports either **verified** or a clear warning. Warnings include an unsupported emitted motion construct, a parser failure, discontinuity, or a reconstructed exported path that exceeds the configured tolerance from its source geometry. The viewer must never silently present a partial parse as verified.

## Motion and data design

```mermaid
flowchart LR
  A[Toolpath generator] --> B[Raw generated moves]
  B --> C[Always-on line optimization]
  C --> D[Canonical optimized moves]
  D --> E[Postprocessor and arc fitting]
  E --> F[Literal G-code text]
  B --> G[Generated layer]
  D --> H[Optimized layer]
  F --> I[Read-only G-code motion parser]
  I --> J[Exported layer in project coordinates]
```

1. Introduce an ephemeral `ToolpathGenerationTrace` (name to be finalized during implementation) containing the raw generated moves and the optimized `ToolpathResult`. Existing generation, caching, preview, and simulation continue to consume only the optimized result. The trace is debug-only and is never serialized into `.camj`.
2. Reuse the normal export call for the selected operation; do not independently reimplement postprocessor formatting for this screen. The debugger receives the output G-code text and the associated generated/optimized trace from that same run.
3. Add a small, read-only motion parser for the emitted dialect used by the built-in postprocessor. It must honor modal coordinates and motion, comments/line numbers, X/Y/Z, G0/G1/G2/G3, and arc I/J or R values. It then applies the inverse machine-to-project coordinate transformation before rendering.
4. Keep parsed arcs analytic in the debug model and tessellate only for SVG display. This makes partial arcs visible as partial arcs and avoids hiding export issues behind a coarse polyline.
5. Compare the parsed exported trace with the postprocessor motion trace using endpoint continuity plus sampled planar deviation. Arc-fit deviation is assessed against the configured export tolerance. Unsupported custom syntax or an incomplete parse produces an explicit unsupported/failed state instead of a misleading overlay.

The debugger should be a narrow new component and pure engine helpers, rather than further growing `ExportDialog`. It will use the existing shared arc reconstruction/fitting and coordinate-transform utilities where they apply; it will not introduce another generic arc-fitting algorithm.

## Implementation plan

1. Define the pure trace, planar-layer, diagnostic, and eligibility types near the toolpath/G-code engine boundaries. Add a generator seam that captures pre-line-optimization moves while preserving existing production cache semantics.
2. Extract or expose the postprocessor motion trace needed to validate the literal output, then implement the strict read-only parser and inverse coordinate conversion. Cover modal coordinates, lines, partial/full G2/G3 arcs, and safe unsupported-input handling.
3. Build the single-operation Export-dialog entry point and a separately sized SVG debugger dialog with layer controls, Z-level filtering, non-cutting-move toggle, fit/pan/zoom, metrics, and diagnostic status.
4. Add engine fixtures that demonstrate merged straight segments, full/partial fitted arcs, multiple discrete Z passes, coordinate inversion, parser failure handling, and ineligible variable-Z operations. Add a browser smoke test for the one-operation entry point and layer toggles.
5. Update the durable G-code export design reference and the relevant directory indexes if new files or behavior require it. Run focused tests, `npm run build`, and the relevant browser smoke.

## Acceptance criteria

- With exactly one eligible operation selected for export, a user can open the debugger and see a 2D view for the actual G-code export.
- The Generated, Optimized, and Exported G-code layers can be toggled independently, and the view has a discrete cutting-Z selector plus an all-levels option.
- A user can identify line optimization through a displayed raw/optimized count delta; partial and full exported arcs are visibly reconstructed as arcs in the exported layer.
- The displayed export layer comes from parsing the literal generated G-code, mapped back into project coordinates, not from rendering an internal approximation in its place.
- The tool surfaces parse/continuity/tolerance problems explicitly and does not claim verification for a partial or unsupported trace.
- Variable-Z 3D operations are clearly unavailable in this release, with no attempt to flatten them into a deceptively simple 2D result.
- Focused unit/browser tests and `npm run build` pass.

## Deferred scope

- Multi-operation overlays and cross-operation ordering inspection.
- 3D path inspection, Z-height shading, material-removal simulation, stock/collision checks, and a 3D G-code viewer.
- Importing arbitrary external G-code or supporting every controller macro/dialect.
- Editing the generated program from the debug view.

## Size and board metadata

- Area: `export`
- Size: `L`
- Status: `Backlog` until this plan is explicitly approved.

## Approval checkpoint

Implementation starts only after explicit approval of this first-version scope. Any decision to include variable-Z work or multiple operations will be planned as a follow-up rather than folded into this diagnostic release.

zabooma:
## Summary

Add a single-operation 2D debug view that overlays Generated (raw), Optimized, and Exported G-code layers in project coordinates. Opens from the Export dialog when exactly one eligible operation is selected.

Closes #356

### Engine
- Read-only G-code parser (gcodeMotionParser.ts) reconstructs motion from literal emitted G-code, honoring modal coords/motion, comments/line numbers, G0-G3, I/J and R arcs. Arcs kept analytic; unsupported syntax yields explicit unsupported/failed states.
- machineToProjectPoint inverse transform (utils.ts) - exact inverse of projectToMachinePoint.
- Motion debug model (motionDebug.ts) - eligibility check (motion-derived, not name-allowlist), source-vs-exported diagnostic, exported-vs-optimized sampled deviation.
- Postprocessor trace exposure - captureMotionTrace option on PostProcessorOptions; when set, PostProcessorResult includes per-operation motionTraces. Zero cost when off.
- Raw toolpath capture seam in useToolpathGeneration.ts; ephemeral rawToolpathRef captures pre-optimization moves; getGenerationTrace hook forces fresh capture.

### UI
- ExportedMotionDebugDialog with SVG plan view, pan/zoom/fit, zoom-to-selection, layer toggles, Z-level selector, per-level stats, scrollable warnings panel, verified/warning diagnostic bar.
- Tablet-friendly: 44px touch targets, responsive grid, 95vw max-width.
- Entry point in Export dialog footer, gated on exactly one eligible operation.
- i18n keys in en/de/es/fr/zh-CN.

### Tests
- gcodeMotionParser.test.ts - modal coords, arcs, unsupported syntax, non-motion setup codes.
- motionDebug.test.ts - eligibility, inverse-transform round-trips, model builder, diagnostic cases.
- motionDebug.smoke.spec.ts - Playwright skeleton (not yet run in CI).

### Docs
- Updated engine/INDEX.md, components/INDEX.md, and G-code_Export_Design.md.

## Known issues
- Arc fitting deviations flagged by the debug view are real (confirmed by #359). Fix belongs in arc fitting code.
- E2e smoke test not yet run in CI.


zabooma (commented):
## Overview

This PR adds a polished, well-engineered 2D motion debug view that overlays Generated (raw), Optimized, and Exported G-code layers in project coordinates. At ~2,400 lines over 26 files, the feature is substantial but cleanly scoped. The architecture is sound: a read-only G-code parser reconstructs planar motion, a debug model builder constructs the three overlay layers with diagnostic comparisons, and a React dialog renders them in an SVG with pan/zoom/fit controls. The inverse transform `machineToProjectPoint` correctly inverts `projectToMachinePoint`, and the postprocessor trace capture is truly zero-cost when off. Test coverage is solid for the engine layer, and all 5 i18n locales are included.

**Verdict: Approve with minor comments.** The issues below are quality improvements, not blockers.

---

### 🔶 1. `sampleSegment` normalizes arc sweep to minor arc only (`motionDebug.ts:280-283`)

In `sampleSegment`, the directed sweep is computed then *re-normalized* to (-π, π], so the sampling always traces the minor arc even when `largeArc` is true. The SVG rendering is correct (the `A` command large-arc flag handles it), but `compareExportedVsOptimized` samples from these points — a large arc deviation could go undetected. In practice postprocessor arcs are ≤90° so this is theoretical, but consider adding a guard or comment noting the limitation.

### 🔶 2. "Fit" and "Reset" buttons are identical (`ExportedMotionDebugDialog.tsx:239-240`)

Both buttons call the same `handleFit()` function. Two differently-labeled buttons that do exactly the same thing is confusing UX. Consider removing the "Reset" button or having it do something distinct (e.g., reset layer toggles and Z-level to the first cutting level).

### 🔶 3. `getGenerationTrace` blocks the UI thread synchronously (`useToolpathGeneration.ts:190-199`)

`getGenerationTrace` calls `generateToolpathForOperation` synchronously inside the debug dialog `useEffect`. For complex pockets with many levels, this regenerates the entire toolpath synchronously and could cause a noticeable UI freeze when opening the debug view.

### 🔶 4. `captureMotionTrace` doubles arc-fitting cost when enabled (`postprocessor.ts:544-559`)

When `captureMotionTrace` is set, the postprocessor recomputes `transformMoves()` and `fitArcsInMachineMoves()` even though these were already computed in the hot path. Since it is only enabled for single-operation export, the blast radius is limited.

---

### 🔹 5. No test for `captureMotionTrace` output (`postprocessor.test.ts`)

The postprocessor tests extensively cover arc fitting but never verify the `captureMotionTrace` option. A test checking `result.motionTraces?.length` and `result.motionTraces[0].machineMoves` would prevent regressions.

### 🔹 6. German i18n: "Schnitttiefe" vs "Schnittebene" (`de/dialogs.ts`)

`dialogs.motionDebug.cuttingLevel: "Schnitttiefe"` means "depth of cut" (a single value), but the selector chooses between discrete Z-levels (planes). "Schnittebene" (cutting plane) would be more precise.

### 🔹 7. Tablet `<select>` element does not get 44px sizing (`dialog.css`)

The Z-level selector `<select>` has no `min-height` override in tablet mode, unlike the toolbar buttons and layer rows. Consistency would be good.

---

### ✅ What is excellent

- **G-code parser**: modal coordinates, arcs via I/J and R, large-arc detection, unsupported/failed states, robust comment stripping, non-motion setup codes correctly ignored
- **Eligibility gating**: motion-derived (drilling, noCutMoves, variableZ), not a name-allowlist
- **Two-tier diagnostic**: source-vs-exported (structural) PLUS exported-vs-optimized (sampled deviation)
- **Inverse transform**: true mathematical inverse, tested with identity and mirrored axes
- **Zero-cost trace capture**: `captureMotionTrace` defaults to false; hot path untouched
- **SVG rendering**: `sweep-flag = clockwise ? 1 : 0` — correct in SVG Y-down space
- **Tablet support**: 44px touch targets on layer rows and toolbar buttons, 95vw max-width
- **All 5 locales**: en/de/es/fr/zh-CN included
- **Unit tests**: modal coords, I/J arcs, R arcs, large arcs, unsupported canned cycles, missing centre fallback, comments, line numbers, setup codes, eligibility, inverse transform round-trips, no-positioning-rapid programs, initial-plunge exclusion, project-coord mapping, segment-count mismatch

src/components/export/ExportedMotionDebugDialog.tsx
```diff
@@ -0,0 +1,446 @@
...
+  const [zoomSelectMode, setZoomSelectMode] = useState(false)
+  const selectionRef = useRef<{ startX: number; startY: number } | null>(null)
+  const [selectionRect, setSelectionRect] = useState<{ x: number; y: number; w: number; h: number } | null>(null)
+  const didInitZRef = useRef(false)
+  // Default to the first cutting Z once the model loads (issue: defaults to the
+  // first cutting Z in machine order; "All cutting levels" is opt-in).
+  useEffect(() => {
+    if (model && model.zLevels.length > 0 && !didInitZRef.current) {
+      didInitZRef.current = true
+      setSelectedZ(model.zLevels[0])
```
zabooma, line 176: **Missing `onPointerCancel` handler**: `onPointerUp` and `onPointerLeave` clean up drag state, but there is no `onPointerCancel` binding. On touch devices, `pointercancel` can fire without a matching `pointerup` (system gesture, notification, etc.), leaving `dragRef.current` and `selectionRef.current` uncleared and the selection rect stuck on screen.

Fix: add `onPointerCancel={handlePointerUp}` to the `motion-debug__svg-wrap` div. `handlePointerUp` already handles both drag and zoom-select cleanup.

src/engine/gcode/motionDebug.ts
```diff
@@ -0,0 +1,526 @@
...
+
+  return { state: warnings.length > 0 ? 'warning' : 'verified', warnings }
+}
+
+// ── Exported-vs-optimized sampled deviation ──────────────────
+
+interface RefSeg { from: XYP; to: XYP }
+
+function buildRefPolyline(moves: ToolpathMove[]): RefSeg[] {
+  const segs: RefSeg[] = []
```
zabooma, line 325: **Index-mismatch in deviation comparison**: `sourceMachineMoves[i]?.kind` indexes into the pre-fit source moves array while the loop iterates `exportedSegments` by index. When arc fitting collapses N source moves into M arcs (N≠M) or rapid-splitting expands the count, the indices won't align. This means the guard meant to skip initial plunges (`sourceKind != null && sourceKind !== 'cut'`) can silently skip or include the wrong exported segments.

The `compareMotionTraces` diagnostic does catch structural mismatches (countMismatch/segmentKind), but because it runs independently on a different index space, it won't always detect that deviation warnings came from misaligned indices.

Suggested fix: since the postprocessor trace already has `machineMoves` and `descriptors`, consider building a position-preserving map from each exported segment back to its source move, or use a Z-based heuristic (skip exported segments whose Z differs from any cutting Z by more than the stepdown).

src/components/export/ExportedMotionDebugDialog.tsx
```diff
@@ -0,0 +1,446 @@
...
+    } else {
+      const drag = dragRef.current
+      if (!drag || !view) return
+      const dx = e.clientX - drag.startX
+      const dy = e.clientY - drag.startY
+      const fx = drag.vb.w / cssSize.w
+      const fy = drag.vb.h / cssSize.h
+      setView({ ...drag.vb, x: drag.vb.x - dx * fx, y: drag.vb.y - dy * fy })
+    }
+  }
```
zabooma, line 238: **`filterSegments` recomputes SVG path strings on every render, including pan/zoom**: During a pan drag, `setView` fires on every `onPointerMove`, triggering a re-render. Each render calls `filterSegments` for each visible layer, which iterates all segments, filters by Z/cutting, and calls `buildMotionLayerPathD` to concatenate the full `d` string. This is wasted work — the model, Z filter, and layer visibility don't change during a pan.

Consider wrapping in `useMemo` keyed on `[model, selectedZ, showNonCutting, visibleLayers]`:
```tsx
const layerPaths = useMemo(() => {
  if (!model) return null
  return (Object.keys(LAYER_STYLE) as LayerId[]).reduce((acc, id) => {
    acc[id] = filterSegments(id)
    return acc
  }, {} as Record<LayerId, { cutting: string; nonCutting: string }>)
}, [model, selectedZ, showNonCutting, visibleLayers])
```
Then use `layerPaths[id].cutting` / `layerPaths[id].nonCutting` in the render instead of calling `filterSegments` inline.

src/components/export/ExportDialog.tsx
```diff
@@ -148,6 +162,7 @@ export function ExportDialog({ onClose, generateToolpath, initialOperationIds }:
           emitToolChanges,
           emitCoolant,
           programName: project.meta.name,
+          captureMotionTrace: activeOperations.length === 1,
```
zabooma, line 165: **`captureMotionTrace` is set for every single-op export, not just when the debug view opens.** Setting `captureMotionTrace: activeOperations.length === 1` means every single-operation export pays the cost of re-running `transformMoves()` + `fitArcsInMachineMoves()` a second time in the trace-capture block (postprocessor.ts:544-559), even when the user never clicks "Inspect exported motion."

The PR description says "zero cost when off" — which is true when `captureMotionTrace` is not set. But for single-operation exports it's always on. The overhead is roughly 2× the arc-fitting + transform work per single-op export.

Two options:
1. Set `captureMotionTrace` only when the user clicks Inspect — re-run the postprocessor at that point. Adds latency at inspect-click time but avoids paying the cost speculatively.
2. Keep the current pre-compute approach but document the trade-off clearly in the option's JSDoc.

zabooma (commented):
## Review (independent pass)

Verified locally on `b18c1ea` (branch is up to date with main). Checks run:
- `npx tsc -b --force` → **reproduces the CI failure** (1 error, below)
- `npx tsx` on gcodeMotionParser / motionDebug / postprocessor tests → all pass (tsx doesn't type-check, which is why CI caught this and local test runs didn't)
- `npx eslint` on all new/changed source files → clean
- Confirmed bundled machine definitions use identity coordinateSystem; confirmed arc fitting caps emitted arcs at ≤90° and never emits negative R

### 🔴 Blocking: CI build fails on a type error in the new test

`src/engine/gcode/postprocessor.test.ts(944,5): error TS2322: Type '"outside-in"' is not assignable to type 'MachiningOrder | undefined'`

`MachiningOrder` is `'level_first' | 'feature_first'` (src/types/project.ts:458). "outside-in" is the offset-pocket *pattern* description, not a machining order — the captureMotionTrace test added in a395e66 uses it as the `machiningOrder` value. One-token fix: `machiningOrder: 'level_first'`.

### New findings (not in the earlier review)

**1. Exported-layer arc chirality is wrong for mirrored machine axes (`motionDebug.ts:161`)**
`segmentFromParsed` copies machine-space `clockwise` into the project-space segment unchanged, and `buildMotionLayerPathD` uses it directly as the SVG sweep flag. That's correct for orientation-preserving transforms (verified by hand for the identity mapping: machine-CW → sweep=1 ✓), but `machineToProjectPoint` is orientation-*reversing* for definitions like `xAxis: '-X'` or an X/Y axis swap — the inverse mirror flips arc direction, so the arc renders bulging to the wrong side (endpoints/centre stay correct). All bundled definitions are identity mappings, so this is a latent edge case for custom mirrored machines; debug-view only, no impact on emitted G-code. Suggest XOR-ing `clockwise` with the sign of the axis-permutation determinant, or documenting the right-handed assumption in `machineToProjectPoint`.

**2. `sourceMachineMoves[i]` index alignment is fragile (`motionDebug.ts:397`)**
`compareExportedVsOptimized` assumes the parsed move list is index-parallel with the postprocessor's `machineMoves`. The parser can drop moves (unsupported constructs `continue` without pushing), shifting every later index and misclassifying `sourceKind`. Impact is bounded — an unsupported parse already raises a warning — but the deviation comparison can then flag or suppress the wrong segments. Consider keying the source classification off endpoint matching instead of position.

**3. Trace capture is paid on every single-op preview, not just on inspect (`ExportDialog.tsx:165`)**
`captureMotionTrace: activeOperations.length === 1` means the doubled `transformMoves` + `fitArcsInMachineMoves` cost (earlier finding #4) runs on *every* single-operation export preview — which re-executes on each option/selection change — even when the user never opens the debug view. "Zero cost when off" holds, but it's effectively always on for single-op exports. Modest in practice; either gate it on intent-to-inspect or accept + adjust the PR description.

**4. `onWheel` + `e.preventDefault()` is a no-op (`ExportedMotionDebugDialog.tsx:188`)**
React attaches `wheel` listeners passively at the root, so preventDefault fails with a console warning and the page behind the modal can scroll while zooming. Attach a non-passive native listener via the wrap ref instead.

**5. Dead check in eligibility (`motionDebug.ts:68`)**
`zSet.size === 0` is unreachable — `cutMoves.length > 0` already guarantees a non-empty set. Harmless; drop it or the redundant `noCutMoves` branch.

### Earlier-review status check
- #1 (sampleSegment minor arc): real but unreachable from this dialect — arc fitting splits at ≤90° and R-format emission never goes negative. Comment addition is sufficient.
- #2 (Fit/Reset duplicate), #5 (captureMotionTrace test), #7 (44px select): fixed in a395e66 ✓ (though #5's new test is what now breaks `tsc -b`)
- #3 (synchronous recompute): still synchronous; fine for one op, will freeze on very large pocket toolpaths — acceptable for a debug view.
- **#6 (German `Schnitttiefe` → `Schnittebene`): still unaddressed** — de/dialogs.ts:204-205.

### Verified good
- Parser: modal coords/motion, I/J + R arcs, unsupported-vs-failed states, comment/line-number stripping — solid and well-tested.
- `machineToProjectPoint` is a true inverse; round-trip tests cover mirrored axes at the *point* level.
- SVG sweep mapping for the identity case is correct (machine-CW → sweep=1 in Y-down project space).
- Eligibility is motion-derived; two-tier diagnostic (structural + sampled deviation) is the right design.
- INDEX files and G-code_Export_Design.md updated per the maintenance rule; all 5 locales present; temp script from a395e66 correctly removed in b18c1ea.
- Note: untracked `cleanup.sh` still sits in the issue-356 worktree root — keep it out of future commits.

**Verdict: one blocking type error (trivial fix), everything else is minor/edge-case. Fix `'outside-in'` → `'level_first'` and this is good to merge.**

PR #377, opened 2026-07-27
Bound arc-fitting candidate search

### LINKED ISSUE #369: Weekly Design Check — arc-fitting hot path has worst-case O(n³) behavior, no dedicated test
## Weekly Design Check

Routine review of this week's changes (2026-07-20 → 2026-07-24) against `ARCHITECTURE.md`/`AGENTS.md`. Most of the week's work (theme tokenization cleanup, i18n, tablet dialogs, view-preset menu) checked out cleanly. One finding rises to a real performance risk worth a dedicated look, surfaced while reviewing issue #354 (toolpath optimization / native arc fitting).

### 1. `findArcRunsInPoints` worst case is O(n³) on the input polyline length

`src/engine/toolpaths/arcReconstruction.ts:625-666`, added/unified in `c9c57fe` and `1ba9270` as the **shared** greedy arc-search core for both the editor offset path and the export path:

```ts
while (i < n - opts.minArcPoints + 1) {
  let best: FittedArcRun | null = null
  for (let end = n - 1; end >= i + opts.minArcPoints - 1; end -= 1) {
    const sub = points.slice(i, end + 1)          // O(run length)
    const fit = fitCircleLeastSquares(sub)         // O(run length)
    if (!fit) continue
    if (!validateArcFitPoints(sub, fit, opts)) continue   // O(run length)
    ...
    break // longest wins; stop trying shorter runs
  }
  if (best) { runs.push(best); i = best.endIndex }
  else { i += 1 }   // <-- no fit found: advance by exactly one point
}
```

For each `i`, the inner loop tries every candidate `end` from `n-1` down to `i + minArcPoints - 1`, doing `O(run length)` work per candidate. When a run genuinely doesn't fit any arc (noisy or organic profiles — e.g. gear teeth, rest-machining islands, hand-drawn imports), `i` only advances by 1, so the outer loop runs close to `n` times. Worst case that's `O(n)` positions × `O(n)` candidate ends × `O(n)` per-candidate work ≈ **O(n³)** on polyline length, with no length cap, no candidate-window limit, and no early-exit heuristic.

This sits on two paths both called out in `ARCHITECTURE.md` §8 as performance-sensitive:

- **Editor/offset path** — `simplifyOffsetContour` (`arcReconstruction.ts:1134`) → `fitArcsInLineRuns` → `findArcRunsInPoints`, called from `src/store/helpers/derivedFeatures.ts:421` on every offset-feature recompute (i.e., on ordinary edits, not just export).
- **Export path** — `fitArcsInMachineMoves` (`src/engine/gcode/arcFitting.ts:208`) → same `findArcRunsInPoints`, called 3× from `src/engine/gcode/postprocessor.ts:496,519,551` (export preview, final export, and the exported-motion debug trace from issue #356 — so it now also runs every time a user opens the new debug view).

None of this is hypothetical-only: gear/pocket rest-region geometry (the same feature area fixed in #352 this week) is exactly the kind of irregular polyline that produces long non-fitting runs.

### 2. The shared core has no dedicated unit test

`1ba9270` ("unify arc-fitting candidate search in editor and export paths") deleted the old editor-only `tryFitArcRun`/`ArcFitOutcome` loop and replaced it with a call into `findArcRunsInPoints`, touching only `arcReconstruction.ts` — no test file was added or changed in that commit, and there is no `arcReconstruction.test.ts` in the repo. Coverage is indirect only, via `src/store/helpers/offsetSimplify.test.ts` (editor) and `src/engine/gcode/arcFitting.test.ts` (export) — neither exercises the shared function's worst-case/large-input behavior. `AGENTS.md` requires unit tests for engine bug fixes/features; a refactor that deletes and replaces the core search algorithm should have direct coverage of the algorithm it introduced, including a large/non-fitting-input case that would catch a perf regression here in CI.

### 3. Related, lower-severity note: two stacked heuristic patches in the same week

`src/engine/toolpaths/restRegions.ts` — `13480be` ("preserve gear island rest wedges") and the very next commit `a697820` ("overlap gear rest regions") each add a new magic-number special case for the same gear-island rest-region code path (`maxIslandWedgeArea = toolRadius * toolRadius * 4`, `residualOverlap = toolRadius * 0.5`, a new `touchesIsland` area/intersection branch). Both are commented and tested (`toolpaths.test.ts` +62/+14 lines), so this isn't untested, but two consecutive area-specific patches in as many days is worth a follow-up look to see whether a more general geometric rule would replace the accumulating special cases — not blocking, just flagging so it doesn't become a third patch next week.

## Suggested fix (for item 1, the main ask)

1. Add a length cap / candidate-window limit to `findArcRunsInPoints` (e.g. cap the candidate `end` search window instead of scanning back to `n-1` every time, or memoize per-`i` fit failures) so a long non-fitting run is bounded, not cubic.
2. Add `src/engine/toolpaths/arcReconstruction.test.ts` covering `findArcRunsInPoints` directly, including a large (500+ point) non-arc-fitting polyline to guard against this regression and document the intended complexity bound.
3. Re-verify both call sites (`derivedFeatures.ts:421`, `postprocessor.ts:496/519/551`) still return identical fit results after any change — this must stay a pure performance fix, not a tolerance change.

## Out of scope

- Changing arc-fit tolerance, validation gates, or output G2/G3 geometry.
- The restRegions.ts heuristic consolidation (item 3) — noted for awareness, not proposed for immediate action.

cc @zabooma — flagging per the weekly design-conformance check.

zabooma:
## Summary

- Bound each `findArcRunsInPoints` greedy candidate search to 128 points, preventing long non-fitting polylines from triggering cubic work.
- Add direct shared-core tests for longest-run selection, sweep direction, exact-circle splitting, production editor validation, minimum-window handling, and non-fitting search scaling.
- Document the new test coverage in the toolpaths index.

## Why

The shared arc fitter re-tested nearly every remaining endpoint from every point on non-fitting input. That stalled both offset recomputation in the editor and export arc fitting.

## Fit contract

Fit tolerances and validation gates are unchanged. Exact circular input longer than the window is split into runs with the same center, radius, and direction. Slightly non-circular input can receive tighter local, within-tolerance fits, which can change run boundaries and fitted circle parameters. The revised, explicitly approved acceptance contract is recorded on issue #369.

## Verification

- `npx tsx src/engine/toolpaths/arcReconstruction.test.ts`
- `npx tsx src/store/helpers/offsetSimplify.test.ts`
- `npx tsx src/engine/gcode/arcFitting.test.ts`
- `npm run build`

No E2E test: this is a pure engine search change with direct unit coverage and existing caller-level tests.

Closes #369

### DIFF
--- a/src/engine/toolpaths/INDEX.md
+++ b/src/engine/toolpaths/INDEX.md
@@ -43,7 +43,8 @@ Toolpath generators. Each file owns one strategy. `index.ts` re-exports everythi
 	- `camOperationSmoke.test.ts` --- per-operation-kind smoke: pocket parallel/waterline patterns, drill-type differentiation (simple/peck/dwell/chip_breaking), post smoke for thin ops (v_carve, surface_clean, follow_line, v_carve_medial; closed-Line V-carve smoke); also documents the stock-target resolver gap
 - `roughSurface.test.ts` / `finishSurface.test.ts` / `finishSurfaceCleanup.test.ts` / `meshSlicing.test.ts` / `vcarveMedial/vcarveMedial.test.ts` — strategy-specific
 - `pocketTessellationConsistency.test.ts` — regression for circle/arc sampling consistency (issue #359): full-circle and broken-circle pockets must have identical chord sagitta
-- arc-reconstruction coverage lives with its store-level callers: `store/helpers/offsetSimplify.test.ts` (offset simplification) and `store/second_cut_test.ts` (segment-preserving boolean reconstruction)
+- `arcReconstruction.test.ts` — direct partial-run arc-search coverage: greedy longest valid runs, sweep direction, and the bounded large non-fitting path (issue #369)
+- additional arc-reconstruction coverage lives with its store-level callers: `store/helpers/offsetSimplify.test.ts` (offset simplification) and `store/second_cut_test.ts` (segment-preserving boolean reconstruction)
 
 ## Adding a new strategy
 1. New file `myStrategy.ts` exporting a generator function.
--- a/src/engine/toolpaths/arcReconstruction.test.ts
+++ b/src/engine/toolpaths/arcReconstruction.test.ts
@@ -1,0 +1,111 @@
+/**
+ * Copyright 2026 Franja (Frank) Povazanj
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Direct regression coverage for the shared partial-run arc search.
+ *
+ * Run with: npx tsx src/engine/toolpaths/arcReconstruction.test.ts
+ */
+
+import { findArcRunsInPoints } from './arcReconstruction'
+import type { PartialArcFitOptions } from './arcReconstruction'
+import type { Point } from '../../types/project'
+
+function assert(condition: boolean, message: string): void {
+  if (!condition) throw new Error(`Assertion failed: ${message}`)
+}
+
+function approxEq(actual: number, expected: number, tolerance = 1e-6): boolean {
+  return Math.abs(actual - expected) <= tolerance
+}
+
+function sampledArc(
+  radius: number,
+  startAngle: number,
+  endAngle: number,
+  segments: number,
+): Point[] {
+  const points: Point[] = []
+  for (let i = 0; i <= segments; i += 1) {
+    const angle = startAngle + ((endAngle - startAngle) * i) / segments
+    points.push({ x: radius * Math.cos(angle), y: radius * Math.sin(angle) })
+  }
+  return points
+}
+
+const FIT_OPTIONS: PartialArcFitOptions = {
+  minArcPoints: 4,
+  maxResidual: 1e-6,
+  maxSegmentAngleDeg: 90,
+  minChordRatio: 0.15,
+  minTotalSweepRad: Math.PI / 180,
+  maxAngularStepRatio: 4,
+}
+
+function testFindsLongestValidArcRun(): void {
+  console.log('Testing longest valid arc run...')
+
+  const points = [
+    ...sampledArc(10, 0, Math.PI / 2, 12),
+    { x: 0, y: 11 },
+    { x: 1, y: 12 },
+    { x: 2, y: 11 },
+  ]
+  const runs = findArcRunsInPoints(points, FIT_OPTIONS)
+
+  assert(runs.length === 1, `expected one arc run, got ${runs.length}`)
+  const run = runs[0]
+  assert(run.startIndex === 0, `expected start index 0, got ${run.startIndex}`)
+  assert(run.endIndex === 12, `expected longest valid end index 12, got ${run.endIndex}`)
+  assert(approxEq(run.center.x, 0), `expected center.x 0, got ${run.center.x}`)
+  assert(approxEq(run.center.y, 0), `expected center.y 0, got ${run.center.y}`)
+  assert(approxEq(run.radius, 10), `expected radius 10, got ${run.radius}`)
+  assert(run.clockwise === false, 'expected counter-clockwise arc')
+}
+
+function testDetectsClockwiseArcRun(): void {
+  console.log('Testing clockwise arc run...')
+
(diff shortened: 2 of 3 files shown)


zabooma:
# Code review — PR #377

Reviewed by measuring `main` and this branch side by side in isolated worktrees, rather than reasoning from the diff. All timings are same-machine, same run.

**Verdict: approach approved; two changes requested before merge** — the geometry claim is incorrect for non-circular input, and the behavioral contract this PR introduces has no test.

*(Posting as a comment rather than a formal review — GitHub does not accept request-changes on your own PR.)*

---

## The performance fix works, and it is a large win

Base vs this branch, measured on both the test's option set and the production editor-offset option set (results were within noise of each other, so one table):

| n (non-fitting input) | `main` | PR #377 |
|---|---|---|
| 400 | 113 ms | 31 ms |
| 800 | 890 ms | 67 ms |
| 1200 | 3,038 ms | 102 ms |
| 2400 | 25,496 ms | **215 ms** |

Scaling on this branch is cleanly linear — doubling n doubles the time (800→67, 1200→104, 2400→213). The cubic is genuinely gone, and the worst case goes from a 25-second freeze to a fifth of a second.

Every verification claim in the PR description checks out independently: all three named test files pass, and `npm run build` is green on the branch.

---

## Blocking: the geometry claim is false for non-circular input

```ts
// A fitted arc can be split downstream without changing its geometry, while
```

**For a true circular arc this holds.** Verified — a 720-point circle splits into 6 runs, and every one recovers an identical circle (`r=10.000000`, `c=(0,0)`). Splitting there is genuinely lossless.

**For anything not exactly circular it does not.** A 0.5%-eccentricity ellipse at n=360, using the real editor-path options:

| | runs | fitted radii |
|---|---|---|
| `main` | 2 | `10.049747`, `10.078884` |
| PR #377 | 3 | `10.063125`, **`9.903509`**, `10.032537` |

Different run count, different centers, different radii — the middle run's radius is ~0.15 mm away from anything base emitted. On a G2/G3-emitting path in CNC software, a change like that should not ship under a comment asserting it didn't happen.

**Two mitigating facts, both measured** — this is why the ask is *fix the claim*, not *abandon the approach*:

1. Both old and new stay inside the fitter's existing tolerance gate. This is **not** a tolerance change.
2. The new output is measurably **closer** to the source polyline — max deviation `0.0198 mm` vs base's `0.0497 mm`. Shorter runs fit their local points better.

So the behavior change is defensible and arguably an improvement. But #369 set an explicit acceptance criterion:

> Re-verify both call sites still return identical fit results after any change — this must stay a pure performance fix, not a tolerance change.

That criterion is not met, and the PR description's *"which downstream arc emission can split without changing geometry"* obscures rather than surfaces it.

**Requested:** correct the comment and PR description to state the real contract — *identical for circular arcs; a within-tolerance refinement otherwise* — and renegotiate the "identical fit results" criterion explicitly in #369 rather than dropping it quietly. Given the output is tighter to source, I'd expect that to be granted on the merits.

---

## Blocking: nothing tests the actual change

The split contract — a >128-point valid arc becomes N runs **on the same circle** — is the entire behavioral effect of this PR and is untested. Asserting `circle(360)` → 3 runs with matching center/radius would document the intent and catch a regression in the greedy/split logic.

## Non-blocking: test fidelity

- **The fit options match neither production caller.** The test uses `maxResidual: 1e-6`, `maxSegmentAngleDeg: 90`, `minTotalSweepRad`, `maxAngularStepRatio`, and no `sourceCenters`. The editor path (`arcReconstruction.ts:939`) uses `radiusToleranceFraction: 0.01`, `maxSegmentAngleDeg: 20`, populated `sourceCenters`, and neither optional gate. The new direct tests therefore never exercise the `radiusToleranceFraction` branch at `:678-681` that the editor actually takes. I confirmed by benchmark that it behaves equivalently — but the tests don't.
- **The timing assertion won't catch a partial regression.** 750 ms against a measured 101 ms is 7× headroom, so no flakiness risk, and a full revert (3,038 ms) trips it. But raising the cap to 512 would sail through. A scaling-shaped assertion (`t(2n)/t(n) < 3`) is machine-independent and catches more.

## Minor

- **`128` is unexplained.** Worth recording *why* it is safe: a full circle at `DEFAULT_FLATTEN_ARC_STEP` (π/36 = 5°) is 72 points, comfortably under the cap, so ordinary flattened circles are unaffected. That is the margin justifying the number.
- **Latent dead branch.** If `minArcPoints` ever exceeded 128, `minEnd > maxEnd` and the loop silently never executes — no arcs, no warning. Unreachable today (`minArcSegments` hardcoded to 6 at `:781`); a one-line guard or comment closes it.
- **Residual cost, for later.** 213 ms at n=2400 still runs on ordinary edits via the offset recompute. Vastly better than 25 s, but not free — follow-up territory, not this PR.

## Conventions

Clean throughout: license header present, `INDEX.md` updated in the same commit per the AGENTS.md maintenance rule, test style matches the existing `toolpaths.test.ts` idiom, and the e2e omission is explicitly justified as AGENTS.md requires.

---

Estimated work to clear the two blocking items: ~30 minutes.

PR #384, opened 2026-07-28
perf: convert PropertiesPanel to per-field Zustand selectors (#233)

### LINKED ISSUE #233: perf: PropertiesPanel subscribes to the entire store — re-renders on every state change
## Problem

`src/components/feature-tree/PropertiesPanel.tsx` calls `useProjectStore()` with no selector and destructures the entire store object (lines 154–199):

```ts
const {
  project, selection, updateFeature, expandTextFeature,
  setMachineDefinitions, ...many more...
} = useProjectStore()
```

This means `PropertiesPanel` — a 1570-line component — re-renders on **every single store state change**, including transient tool state (hover, pending-add anchors, snap events, viewport panning). Any interaction that touches Zustand state triggers a full re-render of this component.

The established pattern is granular subscriptions:
```ts
const project = useProjectStore((s) => s.project)
const selection = useProjectStore((s) => s.selection)
// etc.
```

This was a pre-existing issue, but the recent `feat/issue-221-panel-expand` and `feat/217-text-expansion` additions grew the destructured set without addressing it.

## Impact

- Every canvas interaction (mouse move for snapping, pending-add) triggers an unnecessary re-render of the properties panel
- Panel has disclosure sections, format helpers, and conditional render blocks — all recalculate on transient state changes that are irrelevant to the panel's displayed content

## Fix

Replace the single `useProjectStore()` call with targeted per-field selectors, or group logically related fields into a shallow-equality selector (`useShallow`). At minimum, split `project` from the action callbacks (actions are stable refs and don't need to trigger re-renders).

## Detected by

Weekly design conformance check (2026-06-29).

zabooma:
Closes #233

Replaced the single selector-less `useProjectStore()` call (42-field destructure) with 42 per-field selectors, **and memoised the component** — the second half turned out to be load-bearing.

Of the 42 names, only `project`, `selection`, and `backdropImageLoading` are reactive; the 39 actions are stable refs.

### Why `memo` is required, not cosmetic

The selector conversion **on its own changed nothing measurable.** `src/App.tsx:115` subscribes to the whole store with no selector and renders `<PropertiesPanel />` inline at `:448`, so every store write re-rendered the panel as a child regardless of what it subscribed to. Per-field selectors stop a component re-rendering on its own account; they cannot stop a parent re-rendering it.

The component takes no props, so `memo` is unconditionally safe — the comparison is free and can never go stale.

### Measured

Render counter compiled into each build, Vite dev servers on isolated ports, real browser.

**Real canvas drag, 30 `pointermove` events:**

| build | panel renders |
|---|---|
| `main` | 30 / 30 |
| per-field selectors only | 30 / 30 |
| **selectors + `memo`** | **0 / 30** |

**Store actions, one per macrotask so React commits each, ×40:**

| driver | `main` | this PR |
|---|---|---|
| `hoverFeature`, id unchanged | 40 | **1** |
| `setConstraintAnchor` (top-level, unsubscribed) | 40 | **0** |
| `hoverFeature`, id changes | 40 | 39 |
| `setActiveControl` | 40 | 40 |

The bottom two rows are unchanged by design. Both actions rebuild the `selection` object (`selectionSlice.ts:621` and `:811` — the latter with no equality guard), and the panel legitimately subscribes to `selection` for `selectedFeatureIds`. It never displays `hoveredFeatureId` or `activeControl`, so those re-renders are still waste — but fixing that means narrowing the selector or lifting the transient fields out of `SelectionState`, which is separate work and the larger remaining win.

### Verification

- `npm run build` — green
- No unadorned `useProjectStore()` remains in the file
- Functional check with `memo` applied: panel still updates on `project`, `selection`, and `backdropImageLoading` changes; 37 interactive controls present; no console errors
- All instrumentation was stripped before committing — the diff is the `memo` wrap and the import only

### DIFF
--- a/src/components/feature-tree/PropertiesPanel.tsx
+++ b/src/components/feature-tree/PropertiesPanel.tsx
@@ -159,50 +159,48 @@ function DraftNumberInput({
 }
 
 export function PropertiesPanel() {
-  const {
-    project,
-    selection,
-    addFeatureFolder,
-    startAddTabPlacement,
-    startAddClampPlacement,
-    assignFeaturesToFolder,
-    deleteTab,
-    deleteClamp,
-    deleteFeatureFolder,
-    toggleFolderGrouped,
-    setProjectName,
-    setShowFeatureInfo,
-    setProjectClearances,
-    setSelectedMachineId,
-    refreshMachineDefinitions,
-    setOrigin,
-    startPlaceOrigin,
-    loadBackdropImage,
-    backdropImageLoading,
-    setBackdropImageLoading,
-    updateBackdrop,
-    deleteBackdrop,
-    startMoveBackdrop,
-    startResizeBackdrop,
-    startRotateBackdrop,
-    setGrid,
-    setStock,
-    setStockSourceFeature,
-    updateTab,
-    updateClamp,
-    updateFeatureFolder,
-    updateFeature,
-    updateFeatures,
-    deleteFeature,
-    deleteFeatures,
-    enterSketchEdit,
-    enterStockSketchEdit,
-    enterTabEdit,
-    enterClampEdit,
-    deleteConstraint,
-    makeUnique,
-    expandTextFeature,
-  } = useProjectStore()
+  const project = useProjectStore((s) => s.project)
+  const selection = useProjectStore((s) => s.selection)
+  const addFeatureFolder = useProjectStore((s) => s.addFeatureFolder)
+  const startAddTabPlacement = useProjectStore((s) => s.startAddTabPlacement)
+  const startAddClampPlacement = useProjectStore((s) => s.startAddClampPlacement)
+  const assignFeaturesToFolder = useProjectStore((s) => s.assignFeaturesToFolder)
+  const deleteTab = useProjectStore((s) => s.deleteTab)
+  const deleteClamp = useProjectStore((s) => s.deleteClamp)
+  const deleteFeatureFolder = useProjectStore((s) => s.deleteFeatureFolder)
+  const toggleFolderGrouped = useProjectStore((s) => s.toggleFolderGrouped)
+  const setProjectName = useProjectStore((s) => s.setProjectName)
+  const setShowFeatureInfo = useProjectStore((s) => s.setShowFeatureInfo)
+  const setProjectClearances = useProjectStore((s) => s.setProjectClearances)
+  const setSelectedMachineId = useProjectStore((s) => s.setSelectedMachineId)
+  const refreshMachineDefinitions = useProjectStore((s) => s.refreshMachineDefinitions)
+  const setOrigin = useProjectStore((s) => s.setOrigin)
+  const startPlaceOrigin = useProjectStore((s) => s.startPlaceOrigin)
+  const loadBackdropImage = useProjectStore((s) => s.loadBackdropImage)
+  const backdropImageLoading = useProjectStore((s) => s.backdropImageLoading)
+  const setBackdropImageLoading = useProjectStore((s) => s.setBackdropImageLoading)
+  const updateBackdrop = useProjectStore((s) => s.updateBackdrop)
+  const deleteBackdrop = useProjectStore((s) => s.deleteBackdrop)
+  const startMoveBackdrop = useProjectStore((s) => s.startMoveBackdrop)
+  const startResizeBackdrop = useProjectStore((s) => s.startResizeBackdrop)
+  const startRotateBackdrop = useProjectStore((s) => s.startRotateBackdrop)
+  const setGrid = useProjectStore((s) => s.setGrid)
+  const setStock = useProjectStore((s) => s.setStock)
+  const setStockSourceFeature = useProjectStore((s) => s.setStockSourceFeature)
+  const updateTab = useProjectStore((s) => s.updateTab)
+  const updateClamp = useProjectStore((s) => s.updateClamp)
+  const updateFeatureFolder = useProjectStore((s) => s.updateFeatureFolder)
+  const updateFeature = useProjectStore((s) => s.updateFeature)
+  const updateFeatures = useProjectStore((s) => s.updateFeatures)
+  const deleteFeature = useProjectStore((s) => s.deleteFeature)
+  const deleteFeatures = useProjectStore((s) => s.deleteFeatures)
+  const enterSketchEdit = useProjectStore((s) => s.enterSketchEdit)
+  const enterStockSketchEdit = useProjectStore((s) => s.enterStockSketchEdit)
+  const enterTabEdit = useProjectStore((s) => s.enterTabEdit)
+  const enterClampEdit = useProjectStore((s) => s.enterClampEdit)
+  const deleteConstraint = useProjectStore((s) => s.deleteConstraint)
+  const makeUnique = useProjectStore((s) => s.makeUnique)
+  const expandTextFeature = useProjectStore((s) => s.expandTextFeature)
   const features = useMemo(() => resolvedProjectFeatures(project), [project])
   const { t } = useI18n()
   const backdropFileInputRef = useRef<HTMLInputElement>(null)
--- a/src/components/feature-tree/PropertiesPanel.tsx
+++ b/src/components/feature-tree/PropertiesPanel.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { useCallback, useContext, useMemo, useRef, useState } from 'react'
+import { memo, useCallback, useContext, useMemo, useRef, useState } from 'react'
 import type { ChangeEvent } from 'react'
 import { Icon } from '../Icon'
 import { ExpandedPanelContext } from '../layout/expandedPanelContext'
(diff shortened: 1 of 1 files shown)


src/components/feature-tree/PropertiesPanel.tsx
```diff
@@ -159,50 +159,48 @@ function DraftNumberInput({
...
+  const toggleFolderGrouped = useProjectStore((s) => s.toggleFolderGrouped)
+  const setProjectName = useProjectStore((s) => s.setProjectName)
+  const setShowFeatureInfo = useProjectStore((s) => s.setShowFeatureInfo)
+  const setProjectClearances = useProjectStore((s) => s.setProjectClearances)
+  const setSelectedMachineId = useProjectStore((s) => s.setSelectedMachineId)
+  const refreshMachineDefinitions = useProjectStore((s) => s.refreshMachineDefinitions)
+  const setOrigin = useProjectStore((s) => s.setOrigin)
+  const startPlaceOrigin = useProjectStore((s) => s.startPlaceOrigin)
+  const loadBackdropImage = useProjectStore((s) => s.loadBackdropImage)
+  const backdropImageLoading = useProjectStore((s) => s.backdropImageLoading)
```
zabooma, line 189: **The PR description's reasoning is off by one here.** It says *"only `project` and `selection` are genuinely reactive; the 40 actions are stable refs"*.

`backdropImageLoading` is `boolean` state, not an action — I checked all 42 against `store/types.ts`:

- **reactive state: 3** — `project`, `selection`, `backdropImageLoading`
- **actions: 39**

The code is right: this field drives real render output (`:861-877` — button `disabled` states and the loading label), so it must stay subscribed. Only the description needs correcting.

It does change what the Profiler trace should show: **3** re-render triggers remain, not 2. Worth knowing so a commit fired by a backdrop load is not misread as the conversion having failed.

zabooma (commented):
**Verified mechanically — the conversion is clean.** I diffed the two name lists rather than eyeballing them:

- 42 destructured → 42 selectors, **exact set equality**: nothing missing, nothing extra, no local name diverging from its store key.
- No selector-less `useProjectStore()` remains in the file.
- No `getState()` in the file, so there is no snapshot-consistency concern from splitting the subscription.
- Every selector returns `s.X` directly — no object construction, so no zustand v5 re-render loop risk.
- Branch is up to date with `main`.

Two actionable items below, plus one inline. Neither is a correctness problem.

### 1. Run the Profiler trace before merging, or renegotiate the criterion

The approved plan in #233 made this acceptance criterion 3:

> A React Profiler before/after is included in the PR description, with the actual numbers.

and specifically:

> If the profile does **not** show that, say so in the PR rather than merging on the theory.

The description defers it. That is honest, and the theory is sound — but the whole issue was opened on a *performance* claim that the audit deliberately did not measure, so merging on theory reproduces the gap that created the issue. A canvas mouse-drag trace is a few minutes.

**What the trace should show, so it is not over-read:** the win is in re-render *count*, not per-render cost. `resolvedProjectFeatures` is already memoised on `project` (`:204`), so the heavy derivation never re-ran on hover even before this change. The ~8 unmemoised array passes at `:237-264` and the full JSX reconcile are what stop happening. Expect commits-per-drag to drop toward zero; do **not** expect per-commit duration to change.

### 2. `npm run build` is not in the verification list

The description lists `tsc -b`, `eslint`, and `npm test` run separately. `AGENTS.md` requires `npm run build`, which is a wider gate — it also runs `docs:check` and `sync-icons`. CI covers it here, so this is a note rather than a blocker; worth running locally so a failure is not first discovered in CI.

src/components/feature-tree/PropertiesPanel.tsx
```diff
@@ -159,50 +159,48 @@ function DraftNumberInput({
...
+  const deleteFeature = useProjectStore((s) => s.deleteFeature)
+  const deleteFeatures = useProjectStore((s) => s.deleteFeatures)
+  const enterSketchEdit = useProjectStore((s) => s.enterSketchEdit)
+  const enterStockSketchEdit = useProjectStore((s) => s.enterStockSketchEdit)
+  const enterTabEdit = useProjectStore((s) => s.enterTabEdit)
+  const enterClampEdit = useProjectStore((s) => s.enterClampEdit)
+  const deleteConstraint = useProjectStore((s) => s.deleteConstraint)
+  const makeUnique = useProjectStore((s) => s.makeUnique)
+  const expandTextFeature = useProjectStore((s) => s.expandTextFeature)
   const features = useMemo(() => resolvedProjectFeatures(project), [project])
```
zabooma, line 213: Context for whoever runs the Profiler: this memo is why the win here is **re-render count**, not per-render cost.

`project` is unchanged by a hover, so `resolvedProjectFeatures` never re-ran on pointer-move even before this PR. What this change stops is the ~8 unmemoised array passes at `:237-264` plus the full JSX reconcile of a 1592-line component.

Not a change request — flagging so the before/after numbers get read correctly.

src/components/feature-tree/PropertiesPanel.tsx
```diff
@@ -159,50 +159,48 @@ function DraftNumberInput({
 }
 
 export function PropertiesPanel() {
```
zabooma: **Add `React.memo` here — without it this PR measurably does nothing.**

```suggestion
export const PropertiesPanel = memo(function PropertiesPanel() {
```
(plus `memo` in the `react` import, and close with `})` at the end of the component)

Measured with a render counter compiled into both builds: a 30-event canvas drag produces **30 panel renders on `main` and 30 on this branch**. Adding `memo` takes it to **0**.

The cause is `App.tsx:115` — it subscribes to the whole store with no selector and renders `<PropertiesPanel />` inline at `:448`. Since the panel is not memoized, every store write re-renders `App`, which re-renders the panel as a child no matter what its own selectors say.

The component takes no props, so `memo` is unconditionally safe here — there is no props-comparison cost or staleness risk.

zabooma (commented):
## Profiler measurement — ran it, and the result blocks this PR

The two doc items from my last review are fixed (off-by-one corrected, `npm run build` added). This is the outstanding one, and it did not go the way the description predicts.

**Measured by instrumenting a render counter into `PropertiesPanel` in two throwaway worktrees — `main` and this branch — and driving identical interactions in a real browser.** Counts are actual component renders, not estimates.

### Result: this PR alone changes nothing

Store actions driven one-per-macrotask so React commits each, 40 iterations:

| driver | `main` | **this PR** | this PR + `React.memo` |
|---|---|---|---|
| `hoverFeature`, id changes each event | 40 | **40** | 39 |
| `hoverFeature`, id unchanged | 40 | **40** | **1** |
| `setActiveControl` | 40 | **40** | 40 |
| `setConstraintAnchor` (top-level, unsubscribed) | 40 | **40** | **0** |

Real pointer-event drag across the canvas, 30 `pointermove` events:

| build | renders |
|---|---|
| `main` | 30 / 30 |
| **this PR** | **30 / 30** |
| this PR + `React.memo` | **0 / 30** |

The conversion is correct in isolation, but the panel re-renders exactly as often as before. The description's *"Collapses ~42 re-render triggers down to 3"* does not hold as measured.

### Why

`src/App.tsx:115` subscribes to the **whole store with no selector**, and renders `<PropertiesPanel />` inline at `:448`. `PropertiesPanel` is not memoized, so every store write re-renders `App`, which recreates the element, which re-renders the panel — regardless of what the panel itself subscribes to. `AppShell.tsx:325` does the same thing one level down.

Per-field selectors can only prevent a component from re-rendering *on its own account*. They cannot stop a parent from re-rendering it.

### The fix is one line, and I verified it works

`PropertiesPanel` takes no props, so `React.memo` is unconditionally safe:

```ts
export const PropertiesPanel = memo(function PropertiesPanel() { … })
```

With it, the selectors in this PR start doing their job: unchanged-hover drops 40 → 1, an unsubscribed top-level field drops 40 → 0, and a real canvas drag drops 30 → 0.

**Recommend adding `memo` to this PR.** Without it this merges as a no-op that reads like a perf fix — which is worse than not doing it, because the issue closes and the panel still re-renders on every mouse move.

### Two residual re-render sources, worth a follow-up not a fix here

Even with `memo`, rows 1 and 3 stay high, and both are legitimate given the current state shape:

- `hoverFeature` rebuilds `selection` when the id changes (`selectionSlice.ts:621`)
- `setActiveControl` rebuilds `selection` **unconditionally**, with no equality guard (`selectionSlice.ts:811`)

The panel subscribes to `selection` because it needs `selectedFeatureIds` — but it never displays `hoveredFeatureId` or `activeControl`. Those two transient fields live in the same object as durable selection state, so touching them re-renders every `selection` subscriber.

Fixing that means either narrowing the subscription (`useProjectStore(s => s.selection.selectedFeatureIds)`) or lifting the transient fields out of `SelectionState`. Both are beyond this PR — worth its own issue, and it is the larger remaining win.

### Correction to my earlier review

I wrote *"expect commits-per-drag to drop toward zero."* That was wrong for this PR as it stands — it only becomes true once `memo` is added. Apologies for the misdirection.

---

*Method: render counter compiled into each build, Vite dev servers on isolated ports, store exposed for deterministic action dispatch, plus real `pointermove` dispatch on the canvas. Browser and both servers shut down; instrumented worktrees removed.*

zabooma:
Pushed `a949bb4` adding the `memo` wrap, and re-ran the measurement end to end against a freshly-instrumented control at `origin/main`.

**Real 30-event canvas drag: 30 renders on `main` → 30 with selectors alone → 0 with selectors + `memo`.**

Functional check passed with `memo` applied — the panel still updates on `project`, `selection`, and `backdropImageLoading` changes, 37 interactive controls present, no console errors. Instrumentation stripped before committing; the diff is the import plus the wrap.

PR description updated with the full before/after tables. Waiting on CI.

Issue #381, opened 2026-07-28
ci: require branches to be up to date before merging (a conflicting PR silently gets zero CI)


zabooma:
## Problem

PR #380 was branched from `267d9be`, five merged PRs behind `main`. That produced a merge conflict landing exactly on the #376 drilling mutation fix — but the conflict was **not** the real damage.

**A conflicting PR receives zero CI.** `pull_request` workflows run against a merge ref that GitHub cannot compute when the branch conflicts, so `pr-check.yml` never ran:

```
$ gh pr checks 380
no checks reported on the 'refactor/issue-128-nearest-neighbor-consolidation' branch

$ gh pr view 380 --json mergeable,mergeStateStatus,statusCheckRollup
mergeable=CONFLICTING  state=DIRTY  checks=[]
```

For contrast, #377 (mergeable) ran both jobs and passed.

The failure mode is that the PR page showed **no red X — just nothing**, which reads as "hasn't run yet" rather than "will never run". A branch that does not compile (`geometry.ts:94`, two `TS18049` errors) sat there looking fine. A human review caught it; nothing automated would have.

So staleness does not merely create merge work — **it silently disables the entire quality gate.**

## What is already in place

Ruleset `main-requires-pr` (id `18129403`, active) on `main`:

| rule | state |
|---|---|
| `deletion`, `non_fast_forward` | on |
| `pull_request` | on, `required_approving_review_count: 0` |
| `required_status_checks` | **`build` only** |
| `strict_required_status_checks_policy` | **`false`** ← the gap |

`pr-check.yml` defines two jobs, `build` and `e2e`, both on `pull_request`.

### 1. Enable strict status checks (the main ask)

`strict_required_status_checks_policy: true` is GitHub's *"Require branches to be up to date before merging."* It is exactly the rule this issue asks for, and it already exists — it is just switched off.

```bash
gh api -X PUT repos/PureCutCNC/purecutcnc/rulesets/18129403 --input ruleset.json
```
(full ruleset body with the flag flipped; GitHub's rulesets API replaces rather than patches)

**Enforce at merge time, not PR-create time.** A PR created from a fresh base goes stale while it sits — which is how #380 would have slipped through a create-time check anyway, since it was opened before several of those merges landed. Merge-time enforcement is strictly stronger and needs no new code.

### 2. Add `e2e` to the required checks

[...]

PR #387, opened 2026-07-28
refactor: extract shared effectiveFeed helper + fix O(n²) in areaPathsToDrafts

### LINKED ISSUE #256: Duplicated effective-feed logic across 4 modules + O(n²) clipper rebuild in areaPathsToDrafts
> [!IMPORTANT]
> **Scope reduced 2026-07-27 (audit #374).** The original body below is preserved as filed and is now partly out of date — do not action it directly.
>
> - **Item 1** (stale `REGION_FEATURE_SEMANTICS.md`) — **fixed** by `4edb321`, 2026-07-15.
> - **Item 3** (inline coffee SVG) — **moved to #184**.
> - **Item 4** (shared badge primitive) — **dropped, won't fix.**
> - **Items 2 and 5 are the live scope of this issue.** Both re-verified against `main` @ `b00d016`; line numbers in the body below have moved, and `report.ts` / `playback.ts` have both changed directory. See the audit comments for current locations and measurements.

---

## Goal

Automated weekly design-compliance check covering `origin/main` commits `dbbcadf..32fe418` (2026-06-29 → 2026-07-05, ~48 commits / 100 files / +6746 −572). Reviewed the week's four main feature areas against `ARCHITECTURE.md`, `AGENTS.md`, and `planning/` docs for design drift, repeated code, and UI/UX inconsistency. Construction geometry (#199, PR #253) came back clean — implementation matches `ARCHITECTURE.md` §3 and routes exclusively through `isMachinable`/`modelFeatures()`, no ad-hoc bypasses found. The findings below are the ones that didn't come back clean; recommend triaging into separate `<type>/issue-NN` follow-ups per the normal workflow rather than fixing in this issue.

## Findings

1. **`planning/REGION_FEATURE_SEMANTICS.md` is stale and now contradicts shipped behavior.** PR #249 (`fix/issue-168-region-mask-holes`, 11 commits 07-02→07-04) made region-mask evaluation explicitly **order-dependent** — masks are sorted by project order and the base include/exclude mode comes from the *first* region (`src/engine/toolpaths/regions.ts`, `buildRegionMask`/`baseIncludesSubject`; commit titles "Honor ordered region masks across toolpaths", "Honor project order for region masks", "Honor leading exclude region nesting"). But `planning/REGION_FEATURE_SEMANTICS.md` still states "Region features should be order-independent" and its "Current Status" section has no mention of exclude-only masks or the new base-solid rule (`8b92b1f`). The doc was last touched 2026-06-03 (`abdd0ba`), before this whole window. `AGENTS.md` and `planning/INDEX.md` both require durable design docs to be updated in the same commit as the behavior they describe — that didn't happen here, and the doc is now actively wrong rather than just incomplete.
   - Fix: update `REGION_FEATURE_SEMANTICS.md`'s ordering/base-mode section and "Current Status" to match `regions.ts` as shipped.

2. **Effective-feed-scale application is duplicated across 3 modules, already causing 2 reactive bugs.** `resolveSlotFeedScale` computes the scale once and stamps it onto `ToolpathMove.feedScale`, but *applying* it (`feed * (feedScale ?? 1)`) is hand-duplicated in `src/engine/gcode/postprocessor.ts:398`, `src/engine/toolpaths/operationBooklet/report.ts:158` (which is missing the `tool.defaultFeed` fallback the postprocessor has), and `src/components/simulation/playback.ts:175`. This exact drift is why `47d0eac` ("preserve feedScale through move subdivision so the sim feed readout is correct") and `5de9dc2` ("price slot-feed fragments at their scaled feed in the booklet time estimate") had to land as follow-up fixes this same week.
   - Fix: extract a single `effectiveFeed(feed, move)` (or similar) helper and use it in all three call sites so a future change to the scaling formula can't drift again.

3. **New inline SVG bypasses the icon sprite system.** `src/components/about/AboutDialog.tsx` (added in `7d99af3`, "Buy Me a Coffee support link") hand-rolls an inline `<svg>` coffee-cup icon instead of adding `src/assets/icons/coffee.svg` + `npm run sync-icons` + `<Icon id="coffee">`, per `ARCHITECTURE.md` §6 / `src/assets/icons/README.md`. This is a fresh instance of the pattern already tracked in #184 ("Finish migrating panels to the Icon component") — linking rather than duplicating that issue, but flagging since it's new code, not legacy debt.

4. **Two badges shipped the same week with no shared styling primitive.** `CreationTargetBadge` (`src/components/canvas/CreationTargetBadge.tsx`, `.creation-target-badge` in `src/styles/layout.css:1254-1289`) and the tree region-mask badge (`.tree-region-badge`, `src/styles/layout.css:3290-3311`, from `fd124e1`) were each hand-rolled independently and diverge on every axis (pill vs rectangular radius, 11px vs 9px text, solid vs dashed border, icon+label vs text-only). Some divergence is justified by context (floating canvas overlay vs. inline tree chip), but there's no shared base for the common properties (font-weight, letter-spacing, sizing scale).
   - Fix: extract a minimal shared `.badge-base` (or a `Badge` component) for the common typographic/sizing rules; keep context-specific shape/position as modifiers.

5. **Minor: new O(n²) pattern in `restRegions.ts`.** `areaPathsToDrafts` (added in `3e587aa`) computes hole-nesting depth via a nested `contours.reduce` calling `pointInContour` for every `(index, otherIndex)` pair, and each call re-runs `toClipperPath(other, ...)` redundantly instead of precomputing clipper paths once per contour. Not currently a hot path, but easy to fix while it's fresh: precompute the clipper paths once outside the loop.

## Not flagged (checked, no issue found)

- Region-mask clip logic is well-centralized through `src/engine/toolpaths/regions.ts` (`buildRegionMask`/`applyRegionMaskToPaths`/`clipToolpathResultToRegionMask`) — the 07-02→07-04 fix burst *increased* centralization rather than duplicating logic per toolpath.
- The `1f7fa27` perf fix (build the offset ring tree once per band instead of per Z level) is durable — confirmed still hoisted in both rough and finish-floor paths in `pocket.ts` after all subsequent commits in the window.
- The slot-feed engagement scan (`PriorCutIndex` in `pocket.ts`) is a proper O(1)-amortized spatial grid hash, not an O(n²) scan.
- Construction geometry (#199) fully matches its documented design; `constructionExclusion.test.ts` guards the hard-exclusion boundary; the two same-day follow-on fixes (`058ad75`, `fc2d0ba`) route through the new predicates rather than hardcoding.
- `AboutDialog.tsx` correctly reuses the shared `.dialog`/`.dialog-header` shell rather than reimplementing dialog chrome.
- WebGL context-loss handling (`af8bc5f`) degrades gracefully with a user-visible overlay and pauses playback rather than failing silently.

## Tests

No new tests needed for items 1, 3, 4 (docs/CSS only). Item 2 should get a test asserting the shared `effectiveFeed` helper is used by all three call sites (or at minimum a test that would fail if postprocessor/report/playback drift on a scaled move). Item 5 has no behavior change if fixed as a pure precompute.

## Open questions / risks

- Should items 3–4 be folded into #184's scope instead of tracked separately here? Leaving that to triage.
- None of these are urgent/user-facing breakages — recommend `Low`/`Medium` priority, split into per-item follow-up issues before implementation per the normal issue→plan→approve workflow.

## Out of scope

- Re-auditing older/pre-existing debt not touched this week (e.g. `FeatureTree.tsx`/`PropertiesPanel.tsx` size, `#184`'s remaining inline-SVG panels) beyond what's newly introduced.

zabooma:
Closes #256

## Changes

- Add `src/engine/toolpaths/feed.ts` — `effectiveFeed(moveKind, feedScale, cutFeed, plungeFeed)` takes raw feed numbers so all **four** callers (postprocessor, booklet report, playback ratio, live-feed readout) share one entry point. Plunge returns `plungeFeed`; cut/lead-in/lead-out return `cutFeed × feedScale` (feedScale absent → 1).
- **Migrate all four sites:**
  - postprocessor: resolves `tool.defaultFeed` / `tool.defaultPlungeFeed` fallback before calling
  - booklet report: threads `tool` through, now gets the default feed fallback it was missing
  - `SimulationPlayback.feedRatioForMove`: computes ratio as `effectiveFeed(...) / referenceFeedPerSecond`
  - `SimulationViewport.currentFeedPerSecond`: delegates to shared helper
- Precompute clipper paths and first-point probes once in `areaPathsToDrafts()` so the nested point-in-contour test is O(N) per comparison instead of rebuilding clipper paths O(N×M) times.
- Add `feed.test.ts` covering every move kind, feedScale present/absent, and plunge-ignores-feedScale invariant.
- Update `src/engine/toolpaths/INDEX.md`.

## Verification

- **Build**: `npm run build` passes (lint + tsc + 141 structural test files + vite)
- **Postprocessor**: `feedForMove` delegates with identical semantics; G-code output should be byte-identical on existing projects
- **Booklet report**: `tool` fallback wired in — `operation.feed` is always populated in practice, so the effective feed is unchanged for valid projects; the fallback only kicks in defensively when feed is zero/undefined
- **Playback ratio**: preserves existing behaviour including the `plungeFeedPerSecond <= 0` → ratio=1 guard and the 1e-3 floor clamp
- **Viewport readout**: preserves null-guard for unknown/missing feeds
- **restRegions**: drafts are identical (same values, different performance profile)


src/engine/operationBooklet/report.ts
```diff
@@ -166,25 +169,24 @@ function feedControlledTimeSeconds(
       case 'lead_in':
       case 'lead_out': {
         feedDistance += distance
-        // Slot-feed pocket fragments carry a feedScale multiplier and run
-        // slower than the operation feed — price them at the effective feed
-        // the postprocessor emits.
-        const effectiveFeed = operation.feed * (move.feedScale ?? 1)
-        if (effectiveFeed > 0) {
-          seconds += (distance / effectiveFeed) * 60
+        const feed = effectiveFeed(move.kind, move.feedScale, operation, toolDefaults)
```
zabooma: **This changes booklet output, and the change is unguarded and untested.**

Before: `operation.feed * (move.feedScale ?? 1)`, and `feed <= 0` set `hasInvalidFeed`, which makes the report return `seconds: null`.

After: `(operation.feed || tool.defaultFeed) * (feedScale ?? 1)`.

So when a user explicitly sets `operation.feed` to `0` with a tool default of, say, 1000, the booklet used to **flag the feed as invalid** and now **reports a confident time estimate derived from a feed the operation does not use**. The same applies to the plunge branch below.

The #256 plan called this out as open decision 1 and recommended keeping `report.ts` as-is:

> `Operation.feed` is a required `number` defaulting to `tool.defaultFeed` (`operationDefaults.ts:306`), so this only fires when a user explicitly sets 0 — where arguably the booklet's current `hasInvalidFeed` flag is the *more* correct behaviour.

Making the change anyway is a legitimate call — it does align the booklet with what the postprocessor actually emits. But two things should follow:

1. **Say so in the description.** "gets the fallback it was missing" reads as a bug fix; this is a deliberate change to reported estimates.
2. **Add a test.** I checked: no test anywhere sets `feed: 0`, and `operationBooklet.test.ts` only exercises `feed: 800`. The branch you changed has no coverage in either direction, so nothing pins the intended behaviour.

src/engine/toolpaths/feed.ts
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Plunge moves use the plunge feed; all other motion kinds use the cut feed
+ * multiplied by the move's feedScale (defaulting to 1 when absent).
+ * Falls back to the tool's default feed/plunge feed when the operation's
+ * value is zero or undefined.
+ */
+export function effectiveFeed(
+  moveKind: ToolpathMove['kind'],
+  feedScale: number | undefined,
+  operation: { feed?: number; plungeFeed?: number },
```
zabooma: `operation: { feed?: number; plungeFeed?: number }` is looser than the domain — `Operation.feed` and `Operation.plungeFeed` are both required `number` in `types/project.ts`. Optional here means a caller can omit them and silently land on the tool default rather than getting a type error.

Tightening to `{ feed: number; plungeFeed: number }` costs nothing at the two current call sites and keeps the fallback meaning "the value is 0" rather than "the value might be missing".

Related, for whichever signature you land on: taking `cutFeed`/`plungeFeed` as plain numbers instead of the `operation`/`tool` pair is what would let `playback.ts` and `SimulationViewport.tsx` use this too — see the main review comment.

src/engine/toolpaths/restRegions.ts
```diff
@@ -414,11 +403,27 @@ function areaPathsToDrafts(
...
+  // Precompute clipper representations once so the nested point-in-contour
+  // test below is O(N) per comparison instead of O(N×M) per contour rebuild.
+  const clipperPaths: ClipperPath[] = contours.map(
+    (c) => toClipperPath(c, DEFAULT_CLIPPER_SCALE),
+  )
+  const firstClipperPoints: ClipperPoint[] = contours.map((c) => ({
+    X: Math.round(c[0].x * DEFAULT_CLIPPER_SCALE),
+    Y: Math.round(c[0].y * DEFAULT_CLIPPER_SCALE),
+  }))
+  const pointIn = (ClipperLib.Clipper as unknown as {
```
zabooma, line 415: Non-blocking, but worth a note: this detaches `PointInPolygon` from its receiver. That is safe **today** — I checked the clipper-lib source and verified it empirically; it is a pure static function with no `this` reference, and detached calls return identical results (inside `1`, outside `0`).

It is fragile across upgrades, though: if clipper ever makes that method use `this`, this call site breaks and the failure mode is a wrong nesting depth, which silently flips a rest region between `include` and `exclude` rather than throwing.

Hoisting `const clipper = ClipperLib.Clipper` and calling `clipper.PointInPolygon(...)` keeps the whole optimisation — the win came from precomputing the paths, not from detaching the method.

zabooma (commented):
Built the branch and checked each claim. **`npm run build` passes.** The postprocessor and restRegions changes are both clean; the gaps are in scope and verification, plus one behaviour change that needs surfacing rather than reverting.

### Verified good

- **Postprocessor is behaviour-preserving.** `feedForMove` delegates with identical semantics, including `rapid` still falling into the cut branch via the same non-plunge path. That was open decision 2 in the #256 plan — resolved correctly by preservation.
- **`areaPathsToDrafts` precompute is correct.** Point/path index mapping is preserved (`firstClipperPoints[index]` against `clipperPaths[otherIndex]`), and both derive from the same values the old `pointInContour` built, so drafts are identical.
- `ClipperPoint` is properly exported from `./types`; `feed.ts` carries the licence header.

### 1. Only 2 of the 4 sites were migrated

#256's acceptance criterion 2 was *"All four sites use it; no hand-rolled `* (feedScale ?? 1)` remains outside it."* Two remain untouched:

- `src/engine/simulation/playback.ts:175` — `ratio = move.feedScale ?? 1`
- `src/components/simulation/SimulationViewport.tsx:226` — `feedPerSecond * (pose.feedScale ?? 1)`

Both encode the same rule this helper now owns (scale applies to cut/lead-in/lead-out, never to plunge). Leaving them is exactly the drift the issue was opened to stop, and the sim readout is the surface where a mismatch is user-visible.

They do not fit `effectiveFeed`'s current signature — playback wants a ratio, the viewport works in units-per-second from pre-resolved feeds. The #256 plan proposed `effectiveMoveFeed(moveKind, feedScale, cutFeed, plungeFeed)` for exactly this reason: taking the two base feeds as numbers rather than an `operation`/`tool` pair covers all four callers. **Either widen the signature and migrate all four, or state in the PR that 2 of 4 is the intended scope and why** — but "Closes #256" should not land while half the duplication stands.

### 2. `feed.ts` has no tests

`AGENTS.md`: *"New engine features or bug fixes must include unit tests."* #256's acceptance criterion 1 asked for direct coverage of each move kind, `feedScale` present/absent, and that plunge ignores `feedScale`. That last one is the property most likely to regress silently and it is currently unasserted anywhere.

### 3. `src/engine/toolpaths/INDEX.md` was not updated

A new file was added. `AGENTS.md` maintenance rule: *"when you add, rename, remove, or significantly change the purpose of a file, update the nearest `INDEX.md` in the same commit."* `index.ts` (the barrel) was updated; `INDEX.md` was not.

### 4. No verification section, on a PR that deliberately changes output

The description has no Verification section at all. #256's acceptance criterion 4 was **G-code byte-identical on the example projects**. That should still hold — the postprocessor is semantically unchanged — but it is the check that proves it, and it is the one worth actually running here.

### 5. The report.ts fallback is a real behaviour change, described as a fix

See the inline comment. Short version: the #256 plan recommended *against* this and asked for an explicit decision. Making the change is defensible, but the bullet *"report now gets the `tool.defaultFeed` / `tool.defaultPlungeFeed` fallback it was missing"* reads as a bug fix when it changes what the booklet reports.

---

Also: the branch is **BEHIND** `main` and will need a rebase before it can merge, per the rule added in #381.

src/engine/toolpaths/feed.ts
```diff
@@ -0,0 +1,41 @@
...
+ * and the simulation viewport live-feed readout.
+ */
+export function effectiveFeed(
+  moveKind: ToolpathMove['kind'],
+  feedScale: number | undefined,
+  cutFeed: number,
+  plungeFeed: number,
+): number {
+  if (moveKind === 'plunge') {
+    return plungeFeed
```
zabooma, line 44: Minor: `effectiveFeed('rapid', scale, cutFeed, plungeFeed)` falls through to the cut branch and returns `cutFeed × scale` — a real cutting feed for a rapid move. The doc comment says *"Rapid moves are not fed — callers handle them separately"*, and both callers do guard before calling, so nothing is broken today.

But the guarantee lives in prose and in two call sites rather than in the function, and `feed.test.ts` does not cover `'rapid'` — so nothing would catch a third caller that forgets the guard. In a module that decides feed rates for machine motion, that is worth closing.

Either return `0` for `'rapid'` and let callers treat 0 as "no feed", or narrow the parameter to `Exclude<ToolpathMoveKind, 'rapid'>` so the compiler enforces what the comment currently only requests. The second is free and makes the guard unforgettable.

Either way, a `'rapid'` case in `feed.test.ts` would document the intent.

zabooma (commented):
## Recheck — four of five items closed, and I verified the G-code claim rather than taking it

Every item from my previous review has been addressed except one, and the reworked design is better than what #256 proposed.

### Closed

| item | status |
|---|---|
| Only 2 of 4 sites migrated | **fixed** — all four now share the helper |
| No tests for `feed.ts` | **fixed** — `feed.test.ts`, 116 lines |
| `INDEX.md` not updated | **fixed** |
| Branch BEHIND `main` | **fixed** — merged, now `CLEAN` |
| Verification section absent | **added** |

Reworking the signature to `effectiveFeed(moveKind, feedScale, cutFeed, plungeFeed)` — raw numbers instead of an `operation`/`tool` pair — is what made all four callers fit. That is the right call.

### G-code: verified byte-identical, not just asserted

The description says output *"should be byte-identical"*. I measured it. Generated G-code from all three bundled example projects on `main` and on this branch with an identical harness:

| project | ops | moves | with `feedScale` | result |
|---|---|---|---|---|
| `badge` | 4 | 3,631 | 0 | **identical** — 115,030 B, `c5bd2df25ded` |
| `purecutcnc` | 6 | 6,197 | **131** | **identical** — 132,732 B, `aadd5f8d6643` |
| `t-style-body` | 6 | 26,617 | **630** | **identical** — 643,491 B, `0d894cef3da7` |

`cmp` clean on all three. This is a real exercise of the changed path, not a vacuous pass: 761 `feedScale`-carrying moves across the set, and the emitted G-code contains multiple distinct feed values (`F8`, `F12`, `F18`, `F30`), so both scaled and unscaled feeds are actually being written. **Acceptance criterion 4 of #256 is met.**

### Behaviour review of the two new migrations

Both are behaviour-preserving. `ToolpathMoveKind` is a closed union of exactly five (`rapid | plunge | cut | lead_in | lead_out`), and I checked each against the old code:

- **`playback.feedRatioForMove`** — cut/lead-in/lead-out give `refFps × scale ÷ refFps = scale`, matching the old `ratio = feedScale ?? 1`. Plunge with `plungeFps > 0` gives `plungeFps / refFps`, unchanged; `plungeFps <= 0` early-returns 1, as the old guard did. Rapid early-returns 1, matching the old `default`. The `1e-3` floor is preserved.
- **`SimulationViewport.currentFeedPerSecond`** — null-guards preserved for rapid, unknown kind, and missing feeds.

### Still open: the booklet change is real, and untested

I measured this too, same operation on both branches:

| case | `main` | this PR |
|---|---|---|
| `feed = 30` (normal project) | `50 s` | `50 s` — **unchanged** |
| `feed = 0`, `tool.defaultFeed = 30` | **`Unavailable (invalid feed)`** | **`50 s`** |

So the description's *"effective feed is unchanged for valid projects"* is **confirmed correct**. But the second row is a genuine user-visible change: where the booklet used to tell the user their feed was invalid, it now reports a confident 50-second estimate derived from a feed the operation does not use.

That is defensible — it matches what the postprocessor actually emits — but **no test covers it in either direction.** `grep "feed: 0"` across all `*.test.ts` still returns nothing, and `operationBooklet.test.ts` only exercises `feed: 800`. One test asserting the new behaviour would pin it; without it, the next person to touch this has nothing telling them the old flagging behaviour was traded away deliberately.

**This is the only thing I would still ask for before merge.**

### One minor note

See the inline comment on `feed.ts` about `'rapid'`.

---

*Method: identical harness on both branches building `{operation, tool, toolpath}` and calling `runPostProcessor`, plus `buildOperationBookletReport` for the booklet rows. Differential, so any output delta is attributable to the branch. Worktrees removed.*

zabooma:
Pushed `2500553` closing both remaining items.

### 1. `'rapid'` is now excluded at the type level

`effectiveFeed` takes `FedMoveKind` (`ToolpathMoveKind` minus `'rapid'`), so passing a rapid is a compile error rather than a convention documented in prose.

**Correction to my last review:** I said `feed.test.ts` did not cover `'rapid'`. That was wrong — my grep matched other files in the diff. It *did* cover it, at `:93-97`, asserting `effectiveFeed('rapid', undefined, 600, 200) === 600` with the comment *"rapid should return cutFeed by default (caller responsibility)"*. So the footgun was codified as intended behaviour, which is a bit worse than it being untested. That assertion is replaced with a comment documenting the compile error, since the guarantee is now static.

The two postprocessor call sites already discarded the feed for rapids — they now branch on rapid *before* computing it, which is what lets the narrowing type-check. No behaviour change; the value was never used.

### 2. The booklet fallback is pinned

`testFeedTimeFallsBackToToolDefaultFeed` asserts that `operation.feed: 0` with `tool.defaultFeed: 800` yields `1.5 s` rather than "Unavailable (invalid feed)".

**Verified it actually pins something:** swapped in `main`'s `report.ts` and the test fails —

```
Assertion failed: feed time should use tool.defaultFeed when operation.feed is 0,
not report an invalid feed
```

— then passes again with this branch's version. The docblock records why the trade was made, so it reads as a decision rather than an accident.

### Re-verified after the postprocessor reorder

Since I moved code in `postprocessor.ts`, I regenerated G-code from all three example projects and re-compared against `main`:

| project | bytes | result |
|---|---|---|
| `badge` | 115,030 | identical |
| `purecutcnc` | 132,732 | identical |
| `t-style-body` | 643,491 | identical |

761 `feedScale`-carrying moves across the set. `npm run build` green (`feed.ts tests: 10 passed`).

From my side this is ready to merge.

PR #388, opened 2026-07-28
refactor: share profile geometry helpers via a layer-neutral module

### LINKED ISSUE #234: tech-debt: text/index.ts duplicates canonical geometry helpers (clonePoint, translateProfile, transformProfile)
> [!IMPORTANT]
> **Scope reduced 2026-07-27 (audit #374).** The original body below is preserved as filed and is now partly out of date — do not action it directly.
>
> - **Item 1** (`featureSlice.ts` at the `max-lines` ratchet) — **no longer valid.** The file is at **1103 / 1200** effective lines (ESLint's own counter), 97 lines of headroom, reduced by `3ec140d` / PR #287. The premise "any further addition will definitively breach the guard" is false. Extracting `expandTextFeature` may still be good design, but it needs a design argument, not this one.
> - **Item 2** (`text/index.ts` duplicates canonical geometry helpers) — **live scope of this issue.** Re-verified, and the copies have since *diverged* from canonical (text versions handle `'arc'` only where canonical handles `'arc' || 'circle'`). Latent today. Line numbers in the body below have moved.

---

### 1. `featureSlice.ts` at the max-lines ESLint ratchet

`src/store/slices/featureSlice.ts` has 1338 total lines. After skipping blank and comment-only lines, it sits at **exactly 1200 code lines** — the ESLint `max-lines` cap for `src/store/**` (`eslint.config.js` line 63). The `expandTextFeature` action added in `feat/217-text-expansion` (78 lines) pushed it to the limit.

From AGENTS.md:
> A file approaching its `max-lines` cap is a design smell — split before adding, don't bump the cap.

Any further addition to this slice will definitively breach the guard. The text-expansion action block (`expandTextFeature`, ~78 lines, lines 1259–1336) is a natural candidate to extract into its own slice or helper-backed action, consistent with the existing `expandTextFeature` helper in `store/helpers/textExpansion.ts`.

### 2. `text/index.ts` duplicates canonical geometry helpers

`src/text/index.ts` defines private copies of three functions that already exist in canonical helpers:

| Duplicate in `text/index.ts` | Canonical location |
|---|---|
| `clonePoint` (line ~235) | `src/store/helpers/geometry.ts` |
| `translateProfile` (line ~494) | `src/components/canvas/previewPrimitives.ts` / `src/store/helpers/transform.ts` |
| `transformProfile` (line ~514) | `src/store/helpers/transform.ts` |

Three independent implementations of the same pure geometric primitives is a DRY violation. Bug fixes in the canonical helpers won't propagate to the text engine.

## Fix

1. Extract the `expandTextFeature` store action out of `featureSlice.ts` into a dedicated slice (`textExpansionSlice.ts`) or fold the store wiring directly into the existing `store/helpers/textExpansion.ts`.
2. Remove the private copies of `clonePoint`, `translateProfile`, and `transformProfile` from `text/index.ts` and import from the canonical helpers.

## Detected by

Weekly design conformance check (2026-06-29).

zabooma:
`src/text/index.ts` carried private copies of `clonePoint`,
`translateProfile`, `transformProfile` (plus a `cloneProfile` built from
them) that had diverged from the canonical versions in
`src/store/helpers/`: the text copies branched on `segment.type === 'arc'`
only, where canonical handles `'arc' || 'circle'`. A circle segment passed
through the text copies would keep an untranslated `center`.

Importing the store helpers directly would have created a `text -> store`
edge, inverting the layering — `src/text/` produces profiles that the store
consumes, and it imports nothing from `src/store/` today. So the shared
primitives move to a new neutral `src/geometry/profile.ts` that both sides
import.

- Add `src/geometry/profile.ts` with `clonePoint`, `transformProfile`,
  `translateProfile` (now expressed via `transformProfile`), and
  `cloneProfile`, all handling `circle` alongside `arc`.
- Add direct unit tests, including the `circle`-segment cases that were
  untested in either original location.
- Delete the four private copies from `src/text/index.ts`.
- Remove `clonePoint` from `store/helpers/geometry.ts` and
  `translateProfile`/`transformProfile` from `store/helpers/transform.ts`,
  re-pointing the 10 store importers at the new module. No shims left
  behind. `translatePoint` and the store-only transforms
  (`mirrorProfile`, `transformProfileAffine`, `transformStlFeatureData`)
  stay in `store/helpers/transform.ts`.

Behaviour change is limited to `circle` segments, which go from
"centre not translated" to "centre translated" in the text path. That path
is currently unreachable — text builds profiles from `lineProfile` and
outline-font contours and never constructs `circle` or `arc` segments.
Verified: generating text shapes across 2 styles x 3 fonts x 3 strings x 2
anchors yields `line` segments only, and the full output is byte-identical
before and after this change.

Closes #234


zabooma:
## Review — approve

Reviewed against the branch checkout at `bb40a37`, with the behaviour claim verified independently rather than taken on trust.

### Overview

Extracts `clonePoint` / `transformProfile` / `translateProfile` / `cloneProfile` into a new `src/geometry/profile.ts`, deletes the diverged private copies in `src/text/index.ts`, removes the originals from `store/helpers/{geometry,transform}.ts`, and re-points 10 store importers. Adds `src/geometry/INDEX.md` and a direct unit test.

This is option **B** from the plan in #234, which was the right call — the layering argument holds up. `text/` produces profiles, `store/` consumes them, and a neutral third module is the correct fix rather than a `text -> store` edge. Going without re-export shims (plan stages S2/S4 left this open, "decide in review with the diff in hand") is the better outcome: shims would have left a fourth place for the next reader to find.

### Verification

| check | result |
|---|---|
| `npx tsc -b` | clean |
| `npm run lint` | clean |
| `npm run docs:check` | OK (49 active docs) |
| `npm test` | **142/142 test files pass**, license headers OK |
| `npx tsx src/geometry/profile.test.ts` | passes |

The `circle` behaviour change was re-checked rather than assumed. Generating text shapes plus frame profiles over 4 fonts x 3 strings x 2 anchors on `origin/main` and on this branch gives **byte-identical output** (1,650,996 bytes both), and the emitted segments are `line` x 33,008 with zero `arc` / `circle` / `bezier`. The change is genuinely unreachable from the text path today. That satisfies acceptance criteria 1-5.

One subtlety that could have broken the claim, recorded for the archive: text points originate as THREE `Vector2` instances from `shape.extractPoints(20)`, and the shared `clonePoint` uses `{ ...point }` where the deleted text copy used `{ x, y }`. `Object.keys(new Vector2())` is `['x','y']` (`isVector2` lives on the prototype), so the spread is equivalent and no stray fields can leak into saved `.camj` data. This confirms the correction already made in the issue's plan comment.

### Code quality

- `translateProfile` expressed via `transformProfile` removes the last duplicated segment-kind switch. The extra closure per call is irrelevant at these sizes.
- `transformProfile` spreads `...profile` before overwriting `start` / `segments`, so `closed` and any future field survive — the test pins this.
- The doc comments and `INDEX.md` state the real invariant ("every segment kind that carries points beyond `to`"), which is what actually prevents recurrence. Given the documented history of this bug class in this repo (`transformProfileAffine` circle-centre, `resolveProfile` arcs — `planning/archive/FEATURE_REFERENCES_Ledger.md`), that note earns its place.

### Suggestions — none blocking

**1. `translatePoint` left behind.** `store/helpers/transform.ts:21` still owns `translatePoint`, which is as layer-neutral as anything moved and is logically re-implemented inside `translateProfile`'s closure. The INDEX explains why `mirrorProfile` / `transformProfileAffine` / `transformStlFeatureData` stay (store semantics), but `translatePoint` has no such story — it just splits the primitive set across two homes.

**2. Test doubles now diverge from the canonical implementation.** `src/sketch/constraintSolver.test.ts:46` and `src/store/linkedConstraintResolve.test.ts:66` define local `transformProfile`s described as "matching the store's". Neither transforms bezier `control1` / `control2`. They are injected as the solver's `transformProfile` option, so a future bezier-carrying fixture would exercise something production does not do. Now that the real implementation has zero store dependency, these can just import it.

**3. `cloneProfile` is documented as a deep copy but is shallow per segment.** `transformProfile(profile, clonePoint)` deep-copies Points only; `...segment` shares any other nested value. True enough today, but the test header says "deep copy, no shared point objects" while nothing constrains a future segment field. A one-line caveat on the `cloneProfile` doc comment would keep the next reader honest.

**4. Nits.** The new `'../../geometry/profile'` import is appended *after* the `'./...'` group in `copyFeatures.ts`, `derivedFeatures.ts`, `profileEdit.ts`, `referenceTransforms.ts` and `pendingAddSlice.ts`, where surrounding files put `../../` first — cosmetic, ESLint does not enforce it. The test file uses a single `run()` where the first failed assertion aborts every later case, rather than the `test(name, fn)` pattern most store tests use (`editInPlace.test.ts:186`); matching it would report the full failure set instead of one. And `cloneProfile`'s aliasing assertions only check `segments[2]`/`[3]`, not `line.to` / `arc.center`.

### On the copies left behind

The remaining `translateProfile` / `clonePoint` copies at `components/canvas/previewPrimitives.ts:90`, `store/helpers/featureDefinitions.ts:358` and `import/dxf.ts:460` are correctly out of scope — #234 names them by line number under **Out of scope** and says a blanket sweep needs its own issue. Two notes for whoever picks that up:

- `featureDefinitions.ts` is not a like-for-like swap. Its `cloneProfile` builds `{start, segments, closed}` explicitly where the shared one spreads `...profile`, and its `cloneSegment` clones by reflective `'x' in val` scan rather than by segment kind. Same output today, different contracts. Its "inline to avoid circular deps" comment is now obsolete, though — `geometry/profile.ts` imports nothing but types.
- **The `dxf.ts` copy is not a latent bug, contrary to how it reads.** Its `circle` branch translates `center` and leaves `to` alone, and `convertProfileToTarget:392` does the same under unit conversion. Both are correct: `segmentEndPoint` (`types/project.ts:1060`) documents that a circle's traversal endpoint is `profile.start`, not `to`, and every consumer honours it — radius from `profile.start` vs `center` (`project.ts:1112`, `:1171`, `dimensions.ts:87`), `dimensions.ts:90` branching to a ±2π sweep without reading `to`, and `resolveFeatures.ts:292` rebuilding `to: newStart` rather than trusting the incoming value. Both functions do transform `profile.start`. The stale `to` is inert. `dxf.ts`'s `translateProfile` cannot see a circle regardless — it is reached only from `mergeOpenProfiles`, whose candidates are filtered on `!profile.closed` (`dxf.ts:651`), and dxf circles come from `circleProfile` with `closed: true`. Worth at most a comment noting the convention, not a fix.

### Risk

Low. The behaviour change is confined to `circle` segments in the text path, verified unreachable and byte-identical in output. The 10 re-pointed call sites are covered by type checking plus the existing 142-file suite. No security surface — pure functions over plain data, no I/O, no new dependencies.

---

zabooma:
All four suggestions applied in `05fbb0c`. None changed behaviour — text output re-verified byte-identical, `npm run build` green (142 test files).

| # | suggestion | disposition |
|---|---|---|
| 1 | `translatePoint` left behind | **Applied.** Moved to `geometry/profile.ts`; `translateProfile` expresses its closure through it again. You're right that it had no story for staying — the three that remain in `store/helpers/transform.ts` carry store semantics, and it doesn't. |
| 2 | test doubles diverge | **Applied.** Both `constraintSolver.test.ts` and `linkedConstraintResolve.test.ts` now import the real `transformProfile`. Their locals genuinely didn't transform bezier controls, so the doubles were weaker than what they stood in for. |
| 3 | `cloneProfile` documented as deep | **Applied.** Doc comment now says it copies Points but shallow-spreads segments. |
| 4 | nits | **All applied.** `../../geometry/profile` imports moved ahead of the `./` group; `profile.test.ts` adopts the `test(name, fn)` harness so one failure no longer aborts the rest; aliasing assertions extended to `line.to` and `arc.center`. Also added direct coverage for `translatePoint` now that it lives here. |

On #2 — worth recording that this is the same failure mode the issue predicted, one layer out: a copy described as "matching the store's" that had quietly stopped matching it. It was only fixable this cheaply because the real implementation now has zero store dependency, which is a concrete payoff from picking option B over A.

Not acting on the out-of-scope copies, per #234's **Out of scope** section. Your two notes on them are useful enough that they should outlive this PR — particularly that the `dxf.ts` copy is *correct* rather than latently buggy (circle traversal endpoint is `profile.start`, `to` is inert), and that `featureDefinitions.ts` is not a like-for-like swap and its "inline to avoid circular deps" comment is now obsolete. Happy to open the sweep issue and carry both notes across if you want that tracked.

CI re-running on `05fbb0c`.

---

PR #440, opened 2026-08-03
fix: resolve worktree target dir in block-default-branch-commit hook

### LINKED ISSUE #439: block-default-branch-commit hook denies legitimate worktree commits
## Problem

`.claude/hooks/block-default-branch-commit.sh` blocks legitimate commits made to
a **git worktree** whose branch is not `main`.

The guard resolves the branch from the hook process's own working directory:

```bash
branch=$(git symbolic-ref --short -q HEAD || git rev-parse --abbrev-ref HEAD 2>/dev/null)
```

It never inspects the `-C <path>` or `cd <path> &&` in the command it is
judging. The hook is wired as a `PreToolUse(Bash)` guard and runs with
`${CLAUDE_PROJECT_DIR}` semantics, so for an agent session driving a worktree it
resolves against the primary checkout — which sits on `main` — while the commit
is actually aimed at a feature branch somewhere else entirely.

Result: the guard denies the commit and prints "Refusing to commit directly on
the default branch (main)" for a command that would have committed to
`process/issue-432-fast-lane-discoverability`.

## Reproduction

From a session whose commands target a secondary worktree:

```bash
git -C /path/to/worktrees/purecutcnc/some-slice branch --show-current
# -> process/issue-432-fast-lane-discoverability   (not main)

git -C /path/to/worktrees/purecutcnc/some-slice commit -m "docs: ..."
# -> denied: "Refusing to commit directly on the default branch (main)"
```

Both the `git -C <path>` and the `cd <path> && git commit` forms are denied.

## Why it went unnoticed

The manager-delegate flow's worker commits run under `bypassPermissions`, so
hooks never fire for them. The guard is only reached when a session commits to a
worktree itself, which is exactly the case it gets wrong. Found while committing
a review fix to PR #437.

## Impact

The failure is safe (it denies rather than allows) but it is a false positive on
the repository's own documented workflow — `AGENTS.md` §"Git & Branching" and the
manager-delegate skill both put work in `$PURECUT_WORKTREE_BASE/<slug>`
worktrees. It pushes toward `--no-verify`, which defeats the guard entirely.

## Proposed change

Parse the target directory out of the command and resolve the branch there,
falling back to the current behaviour when no directory is given:

1. Extract a `-C <path>` argument or a leading `cd <path> &&` from
   `.tool_input.command`.
2. Run the `symbolic-ref`/`rev-parse` probe with `git -C "$target"`.
3. Keep denying only when *that* branch is `main`/`master`.

Worth handling `git -C a -C b` (git applies them cumulatively, last-relative-wins)
by simply taking the resolved final path, and leaving the guard fail-closed: if
the target cannot be resolved, check the cwd as it does today.

## Acceptance criteria

- [ ] A commit targeting a non-default branch in another worktree is allowed,
      via both `git -C <path> commit` and `cd <path> && git commit`.
- [ ] A commit targeting a worktree that *is* on `main`/`master` is still denied.
- [ ] A plain `git commit` on `main` in the primary checkout is still denied.
- [ ] A plain `git commit` on a feature branch in the primary checkout is still
      allowed.
- [ ] Non-commit commands (`git commit-tree`, `git log --grep=commit`) still pass
      through, as the existing regex intends.

## Verification

Exercise each of the five acceptance cases by hand against a real worktree and
record the observed decision. A guard whose negative cases have not been
re-checked after a rewrite is not known to still block anything — the two
"still denied" cases matter as much as the fix.

## Notes

Touches `.claude/**`, so this is full lane.

zabooma:
## Problem
`block-default-branch-commit.sh` denied legitimate commits aimed at a feature branch in a worktree. The hook probed its own cwd (the primary checkout, on `main` in a worktree-driven session) and never inspected the `git -C <path>` or `cd <path> &&` in the command it was judging.

## Fix
Parse the target directory out of the command and probe the branch there:
- apply a leading `cd <path> &&`/`;` (single- or double-quoted paths supported);
- apply each `git -C <path>` global option cumulatively (absolute overrides, relative stacks) — matching git's own semantics; handles both `-C <path>` and `-C<path>`;
- scan only the pre-`commit` segment: `-C` *after* `commit` is `git commit -C <commit-ish>` (message reuse), not a directory;
- when no directory is given, fall back to cwd (the pre-fix behaviour); when a parsed target is unresolvable, fall back to cwd too — fail-closed, since the primary sits on `main`.

## Acceptance verification (real git worktrees)
Each case run by hand against real worktrees; observed decision recorded:

| # | Case | Expected | Observed |
| - | --- | --- | --- |
| 1a | `git -C <worktree-on-feature> commit` | ALLOW | ALLOW ✓ |
| 1b | `cd <worktree-on-feature> && git commit` | ALLOW | ALLOW ✓ |
| 2 | `git -C <worktree-on-master> commit` | DENY | DENY ✓ |
| 3 | plain `git commit`, primary on `main` | DENY | DENY ✓ |
| 4 | plain `git commit`, primary on feature | ALLOW | ALLOW ✓ |
| 5a | `git commit-tree` | passthrough | ALLOW ✓ |
| 5b | `git log --grep=commit` | passthrough | ALLOW ✓ |

### Edge cases (extra confidence)
- `git -C <wt> -C . commit` → ALLOW (cumulative resolves to `<wt>`). 
- `git -C <wt-feature> log && git commit` → DENY: the `-C` on `log` is not applied to the plain `git commit`; the committing git is the primary on `main`. Correct.
- `git commit -C HEAD~1` → DENY (primary is `main`): the post-`commit` `-C` is message-reuse, not a directory, so target stays cwd.
- `git add -A && git -C <wt-feature> commit` → ALLOW.
- `git -C /nonexistent commit` → DENY: unresolvable target falls back to cwd (`main`). Fail-closed.

## Gates
- `npm run build` — green (158 test files pass; vite build OK).
- `npm run check:fast-lane` — NOT ELIGIBLE (full lane): protected path `.claude/**`, 76 lines > 25. Plan + explicit approval captured in issue #439 before branching. No `fast-lane` label.

## Files
- `.claude/hooks/block-default-branch-commit.sh` — rewrite (the only change).

### DIFF
--- a/.claude/hooks/block-default-branch-commit.sh
+++ b/.claude/hooks/block-default-branch-commit.sh
@@ -12,87 +12,114 @@
 # when no directory is given we fall back to the cwd (the primary checkout).
 # That fallback is fail-closed: in a worktree-driven session the primary sits
 # on main, so an unresolvable target denies rather than slips through.
+#
+# Compound commands are checked in full: the command is split on chain
+# operators (&&, ||, ;, |, &) and every `git commit` segment is resolved. A
+# `cd <path>` segment updates the running directory for later segments (as it
+# would in a real shell); `git -C <path>` applies only within its own segment
+# (git never changes the shell cwd). If ANY commit segment resolves to the
+# default branch the whole command is denied — `git -C /feature commit && git
+# commit` does not get to slip the second commit onto main.
 
 input=$(cat)
 command=$(printf '%s' "$input" | jq -r '.tool_input.command // ""')
 
-# Only act on an actual `git commit` invocation (also catches it inside a
-# compound command like `git add -A && git commit`). Does not match
-# `git commit-tree`, `git log --grep=commit`, etc.
-#
-# Capture the segment between `git` and `commit` (the global-options segment)
-# so we can parse `-C <path>` from it. `-C` appearing after `commit` is
-# `git commit -C <commit-ish>` (reuse message), not a directory, so it is
-# deliberately excluded from that segment.
+# Tokenize the command into segments by chain operators so every `git commit`
+# in a compound command is checked, not just the first. Operators are replaced
+# with newlines (length-descending: && before &, || before |) so multi-char
+# operators are not split into single chars.
+nl=$'\n'
+split="$command"
+split="${split//&&/$nl}"
+split="${split//||/$nl}"
+split="${split//;/$nl}"
+split="${split//|/$nl}"
+split="${split//&/$nl}"
+
+# Matches a `git commit` invocation and captures the global-options segment
+# between `git` and `commit`. `-C` after `commit` is `git commit -C <commit>`
+# (reuse message), not a directory, so only the pre-`commit` segment is
+# scanned for -C. Does not match `git commit-tree`, `git log --grep=commit`.
 detect_re='(^|[^[:alnum:]])git[[:space:]]+([^|;&]*[[:space:]])?commit([[:space:]]|$)'
-if ! [[ $command =~ $detect_re ]]; then
-  exit 0
-fi
-global_opts="${BASH_REMATCH[2]}"
-
-# Resolve the directory the git command targets. Start from the hook's cwd
-# (the primary checkout) and apply a leading `cd <path>` and any `git -C
-# <path>` global options cumulatively — matching git's own semantics.
-target="$(pwd)"
-
-# A leading `cd <path> &&` (or `;`) redirects the working directory for the
-# rest of the command. The path may be single- or double-quoted.
-cd_re="^[[:space:]]*cd[[:space:]]+(\"([^\"]+)\"|'([^']+)'|([^[:space:]&;|]+))[[:space:]]*(&&|;)[[:space:]]*(.*)$"
-if [[ $command =~ $cd_re ]]; then
-  cd_target="${BASH_REMATCH[2]:-${BASH_REMATCH[3]:-${BASH_REMATCH[4]}}}"
-  if [[ $cd_target == /* ]]; then
-    target="$cd_target"
-  else
-    target="$target/$cd_target"
-  fi
-fi
-
-# `git -C <path>` global options, applied in order. Handles both `-C <path>`
-# (separated) and `-C<path>` (joined); paths may be quoted. Other global
-# options are skipped one token at a time so a later -C is still picked up.
+# A `cd <path>` segment (own segment after splitting) updates the running cwd.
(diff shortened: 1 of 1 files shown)


zabooma:
### P1 — compound command bypass: fixed (c781f34)

Confirmed the bypass. The segment resolver only matched the first `git … commit`, so `git -C <feature-worktree> commit && git commit` from a `main` checkout let the second commit land on `main`.

**Fix:** tokenize the command on chain operators (`&&`, `||`, `;`, `|`, `&`) and resolve **every** `git commit` segment:
- A `cd <path>` segment updates a running directory that persists across later segments (as `cd` does in a real shell).
- `git -C <path>` applies only within its own segment (git never changes the shell cwd).
- If **any** segment resolves to `main`/`master`, the whole command is denied.

This mirrors real shell semantics: `cd /a && git commit && git commit` → both commits resolve to `/a`; `git -C /a commit && git commit` → only the first resolves to `/a`, the second to the shell cwd.

### Regression cases (real worktrees; observed decision)
| # | Command (cwd) | Expected | Observed |
| - | --- | --- | --- |
| R1 | `git -C <wt-feature> commit && git commit` (main) | DENY | DENY ✓ |
| R2 | `git commit && git -C <wt-feature> commit` (main) | DENY | DENY ✓ |
| R3 | `cd <wt-feature> && git commit && git commit` | ALLOW | ALLOW ✓ |
| R4 | `git -C <wt-feature> commit && git -C <wt-feature> commit` | ALLOW | ALLOW ✓ |
| R5 | `git -C <wt-feature> commit && cd <wt-master> && git commit` | DENY | DENY ✓ |
| R6 | `git add -A && git -C <wt-feature> commit && git commit` (main) | DENY | DENY ✓ |
| R7 | `git -C <wt-feature> commit ; git commit` (main) | DENY | DENY ✓ |
| R8 | `git -C <wt-feature> commit \|\| git commit` (main) | DENY | DENY ✓ |
| R9 | `git commit \| cat` (main) | DENY | DENY ✓ |
| R10 | `git -C <wt-feature> commit -C HEAD~1` (reuse-msg, on feature) | ALLOW | ALLOW ✓ |
| S1 | `cd <wt-feature> && git -C subdir commit` | ALLOW | ALLOW ✓ |
| S2 | `git -C <wt-feature> commit && cd <wt-feature> && git -C subdir commit` | ALLOW | ALLOW ✓ |

The original 5 acceptance cases from #439 also still pass (1a/1b ALLOW, 2/3 DENY, 4 ALLOW, 5a/5b passthrough). `npm run build` green.

### Known limitation (pre-existing, out of scope)
Subshell grouping like `(git commit) && git commit` is not parsed — `commit)` does not match the `commit([[:space:]]|$)` boundary, same as the original hook. The realistic threat (`&&`/`;`/`||`/`|` chaining) is covered.

Issue #467, opened 2026-08-08
fix(cam): operation start plunges diagonally across the stock at plunge feed (gouge)


zerg32:
## Summary

The first plunge of a non-trochoidal contour operation is emitted as a single `G1 X Y Z F<plungeFeed>` that travels diagonally from the previous operation's end position across the workpiece while descending. For full-depth passes this descends below the stock top mid-travel, cutting the stock at the wrong location.

## Reproduction

Project: drill op + "Edge route outside Finish" (full-depth contour, 4 mm stock). Export with `optimizeLinearMoves` + `runPostProcessor` (the app pipeline):

```
G0 Z5.000
X195.289 Y7.250              ; end of drill op (safeZ, last hole)
; Operation 2: Edge route outside Finish
; Tool 4: 3mm SF long
G1 X54.433 Y41.499 Z-4.000 F100.000     ; diagonal + full depth at plunge feed
X54.377 Y41.500 Z-4.000 F600.000
```

The G1 interpolates from (195.289, 7.250, 5.000) to (54.433, 41.499, -4.000). With stock top at Z0 it crosses the surface at 55.6% of the travel, so the tool is **cutting the stock, off the intended contour, for ~44% of the move at plunge feed**.

## Root cause

Three layers line up:

1. Contour toolpaths start each operation with a **zero-length positioning rapid** followed by the plunge — `pushRapidAndPlunge` with `from === null` (`pocket.ts:195`, `if (!from ...)` emits a rapid whose `from == to`). This is a documented, load-bearing marker: `entry.ts` — *"A zero-length rapid tells it to position XY and Z at safeZ before the next rapid"*. Used by pocket, edge-route inside/outside, v-carve, surface ops.
2. `optimizeLinearMoves` (`linearMoveOptimization.ts:39`) deletes **every** zero-length move, silently removing the marker.
3. The postprocessor emits each non-rapid move as one `G1 X Y Z` from `move.to` only, ignoring `move.from` (`postprocessor.ts`), so the travel is done at the move's own feed (plunge) — diagonal, through the stock.

## Fix

In `optimizeLinearMoves`, preserve zero-length **`rapid`** moves; continue removing zero-length `cut`/`plunge`/`lead_in` moves. A zero-length rapid is a positioning directive, never noise: `emitRapid` in the postprocessor compares against its tracked machine position and emits only the needed `G0` axes (or nothing when already there), so preserved markers are harmless.

Expected output after fix:

```
[...]

PR #480, opened 2026-08-11
fix(cam): clamp the drilling retract plane to the material surface

### LINKED ISSUE #479: drilling: retractHeight interpreted as absolute project Z — value below stock top emits rapid plunge through material
## Summary

The drilling operation's `retractHeight` parameter is interpreted as an **absolute project-space Z coordinate**, but the UI presents it as a plain "Retract Height". Any value below the top of the stock (project Z `stock.thickness`) is silently accepted and emitted as a **rapid** descent and, on the first hole, a **rapid** XY traverse through the material. There is no clamping and no warning.

This is a machine-safety defect: the spindle turns on and the drill rapid-plunges through the part.

## Reproduction

1. Stock: 4 mm thick (project Z `0..4`), machine origin `z = 4` → stock top = machine Z `0`, stock bottom = machine Z `-4`. `operationClearanceZ = 5`.
2. Create a drilling op over circle features spanning the full stock thickness (`z_top = stock.thickness`, `z_bottom = 0`).
3. Set **Retract Height = 2**.
4. Export.

### Observed G-code (first hole)

```gcode
M3 S18000.000
G0 Z-2.000        ; rapid +5 → -2: descends 2 mm INTO the stock (top is Z0)
X35.289 Y32.750   ; rapid traverse to first hole at Z-2, through material
G1 X35.289 Y32.750 Z-3.000 F150.000
G0 Z5.000         ; peck retract to safe plane (machine +5)
Z-2.500           ; peck re-entry
G1 X35.289 Y32.750 Z-4.000
G0 Z5.000
```

Every subsequent hole repeats the `Z5.000` → `Z-2.000` rapid descent (which crosses the stock top at Z0) before the fed plunge. The top 2 mm of each hole is cut in rapid, never at plunge feed; the very first hole also does its travel move at Z-2.

### Expected

Retract plane must be at or above the stock top. For the intended "2 mm above material" the retract plane should be machine `+2`:

```gcode
G0 Z2.000         ; rapid to 2 mm above the stock, safe
X35.289 Y32.750   ; traverse above the material
G1 X35.289 Y32.750 Z-3.000 F150.000
```

## Root cause

`src/engine/toolpaths/drilling.ts:384-389` computes the retract plane as a raw project-Z value and only caps it from above:

```ts
const retractZ = operation.retractHeight !== undefined
  ? Math.min(safeZ, operation.retractHeight)
  : Math.min(safeZ, highestTop + defaultRetractOffset)
```

- `safeZ` (clearance plane) = `max(stock.thickness, highestFeatureTop) + operationClearanceZ` (`geometry.ts:395`).
- `retractHeight` is never clamped to the stock top, so `retractHeight: 2` with a 4 mm stock (top at Z4) yields `retractZ = 2` — 2 mm inside the material.
- `emitDrillCycle` (`drilling.ts:246-250`) then emits a **rapid** from `safeZ` down to `retractZ`, and when `current` is null (first hole) the XY traverse to the hole happens at `retractZ` too (`drilling.ts:242` skips the "rapid to safe" positioning move). `emitSimplePlunge` (`drilling.ts:218-220`) has the same unguarded rapid descent.

The value is confusingly presented: `src/components/cam/CAMPanel.tsx:1667` and the default in `src/store/helpers/operationDefaults.ts:347` both use `stock.thickness + 1`, i.e. they treat the field as an **absolute project Z** whose "1 mm above material" default only works because the stock's top sits at Z `= thickness`. A user entering `2` reasonably means "2 mm above the part".

## Scope

- `src/engine/toolpaths/drilling.ts` — `retractZ` computation, `emitDrillCycle`, `emitSimplePlunge`.
- `src/engine/toolpaths/entry.ts:725-729` — helical bores make the same rapid descent to `retractZ`; must be covered by the same guard.
- `src/engine/toolpaths/warningCodes.ts` + `src/i18n/locales/*/warnings.ts` — new warning code.
- Possibly `src/components/cam/CAMPanel.tsx` — surface the warning / validate the field.

## Proposed fix

The field's meaning itself is the defect. Two candidate directions:

**Option A — reinterpret the value as relative to the machine origin (recommended).** `retractHeight` is stored and edited as a distance above the machine origin (`project.origin.z`, which defaults to the stock top). A value of `2` then means "2 mm above the origin / above the part" — matching what the field name implies and what users enter. Generation resolves it to `origin.z + retractHeight` (clamped at `safeZ`), so the retract plane can never be inside the stock. Note this changes the stored meaning of existing `retractHeight` values (`stock.thickness + 1` defaults would need a `.camj` format migration so old projects aren't reinterpreted as ~4× higher).

**Option B — keep absolute-Z semantics, clamp and warn.** Clamp the effective retract plane to at least the stock top (`Math.max(operation.retractHeight, stock.thickness)`), emit a new warning (e.g. `drillRetractBelowStockTop`, with requested/clamped values as params) when a clamp fires, apply the same guard to the helical path in `entry.ts`, and add a live `min = stock.thickness` on the UI field.

Either way, the acceptance criteria below must hold.

## Acceptance criteria

- With `retractHeight < stock.thickness`, exported G-code never contains a rapid whose Z goes below the stock top; the first move of each hole is a rapid to a plane ≥ stock top, the traverse is above the material, and only the fed plunge crosses into it.
- A warning (not a silent clamp) is raised when the requested retract height had to be raised, or the value is stored and understood relative to the origin so this can't happen.
- Helical drilling obeys the same invariant.
- Unit tests cover: value below stock top (resolved/clamped + warned), value at/above stock top (unchanged), value above safeZ (unchanged, still capped at safeZ).

## Tests to add

- Engine test in `src/engine/toolpaths/` asserting no rapid `to.z < stock.thickness` for a fixture with a hostile retract height, plus the new warning code.
- Extend the existing entry-safety fixtures (`assertEntrySafe` pattern, #467/#470) to drilling with a hostile `retractHeight`.
- If Option A: format-migration test for existing saved projects carrying absolute-Z `retractHeight` values.

## Workaround (immediate)

Treat the field as an absolute project Z: set `retractHeight` to `stock.thickness + 2` (= `6` in the example) to get a 2 mm clearance above the part.

zabooma:
Closes #479.

`retractHeight` is an absolute project-space Z but was only capped from above, so any value below the top of the stock was accepted verbatim. The generator emitted a rapid descent into the part and — on the first hole, where the postprocessor has no known machine position and so emits Z before XY — a rapid traverse through it. The top of every hole was cut at rapid rather than plunge feed. Canned-cycle machines got the same plane as the `R` word.

Reported externally with a correct root-cause analysis; reproduced at `3a8a3ac` before changing anything.

## The fix

Clamp the retract plane up to `max(stock.thickness, highest feature top)` before capping it at the clearance plane, and warn when an operator-supplied value had to be raised. The two bounds cannot conflict: `getOperationSafeZ` is that same material top plus the operation clearance, so `safeZ >= materialTopZ` always holds — including at `operationClearanceZ: 0`, where the retract plane lands on the surface and no rapid goes below it.

Two things the original report did not cover:

- **Canned cycles were affected too.** `drillCycles[].retractZ` carried the unclamped value into the `R` word, so a controller with canned-cycle support got `G81 X20 Y60 Z-4 R-2 F150` — an R plane 2 mm inside the part. No postprocessor change was needed; it inherits the clamped plane.
- **First-hole positioning is an independent defect.** `emitDrillCycle` skipped the positioning rapid when `current` was null, so hole 1 traversed at the retract plane while holes 2+ traversed at `safeZ`. Both drill emitters now use the zero-length safe-Z entry marker `emitCenterLockedCircularBore` already used — the marker `linearMoveOptimization` deliberately preserves.

Helical boring needed no separate guard: `emitCenterLockedCircularBore` has exactly one production caller and receives the clamped value, which covers both its descent and the centre-to-helix-start rapid. Adding a clamp inside `entry.ts` would have required project context that helper deliberately does not take.

### Not the approach the report recommended

The report's preferred option was to reinterpret the field as relative to the machine origin. That is unsound on its own: `origin.z` is a free numeric field and the origin properties panel has a **"bottom left" preset that sets it to `0`**. Z-zero on the spoilboard is a normal setup, and "2 mm above the origin" would then land 2 mm above the table — back inside a 4 mm part. A clamp against the material surface is required whatever the field means, and the relative reading would add a `.camj` migration on a protected path without removing the need for it.

The reporter's reading is still right, and the UI agrees with them — the parameter-reference icon dimensions the arrow from the stock top up to the retract plane, i.e. it depicts a relative distance. Tracked as a follow-up rather than folded in here.

## Deviations from the approved plan

All three narrow scope rather than widen it:

- The warning fires only for explicit operator values. Clamping our own default would be reporting our arithmetic back to the user.
- The default path was unsafe too — it derived from the highest feature top alone, so it sat inside the stock whenever features were below the top. Same defect through a different door. It now keeps its 1 mm offset above the material surface rather than resolving to the surface exactly.
- The canned-cycle `R` assertion lives in the new test file rather than `postprocessor.test.ts`, keeping one home for this coverage.

## UI

The retract field's `min` becomes the stock thickness. `DraftLengthInput`'s `min` is a commit-time guard, not a DOM attribute: a below-minimum entry is refused and the field snaps back rather than being clamped. The engine warning remains the backstop for projects already saved with an unsafe value.

## Tests

`drillingRetract.test.ts` asserts one invariant over both the toolpath and the posted program: **every rapid endpoint below the material surface must sit where a fed move already cut.** That is deliberately narrower than "no rapid goes below the stock top", which is false — G83 re-entry into the hole the cycle just drilled, and a helical bore retracting across the floor it just cleared, are both correct. Both move only through material that is already gone.

Verified the tests fail against the unfixed generator, and that the geometric invariant fails rather than only the warning assertion:

```
✗ rapid reached (20, 20, 2) — below the material surface (4) at a location no fed move has opened
✗ R plane should be at or above the stock top (machine Z 0), got R-2
```

Coverage: below stock top (clamped + warned), at/above (untouched), above `safeZ` (still capped), helical, peck re-entry, canned-cycle `R`, and the first-hole traverse. `camOperationSmoke.test.ts` gains a hostile-`retractHeight` fixture so the shared `assertEntrySafe` check (#467/#470) sees it too.

## Verification

`npm run build` green on 174 test files, re-run after rebasing onto current `main`.

**`npm run test:e2e` was not run.** The sandbox's network policy denies `gc.zgo.at`, which `src/utils/analytics.ts` loads at runtime, and `e2e/fixtures.ts` fails any test with a console error — so every spec fails there, including ones this change cannot touch. Environmental, not a product defect; CI should run the gate normally. In its place the UI change was driven directly in a real browser with that request stubbed: with 2″ stock, entering `1` is refused and the field snaps back, entering `5` commits.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QH3gMNPcd395k4NGaYT3Bi)_


zabooma:
## Before / after, captured from the running app

Fixture: 12 mm stock, three through-holes, 4 mm drill, `operationClearanceZ: 6`, machine origin on the stock top, **Retract Height = 3** — i.e. 9 mm inside the part. Both programs are the export dialog's G-code preview, taken from the same project on either side of the fix.

**Before** — the rapid descends 9 mm into the stock, and the first hole's traverse happens down there:

```gcode
M3 S18000.000
G0 Z-9.000                              ; rapid 9 mm INTO a 12 mm part
X12.000 Y60.000                         ; traverse at Z-9, through material
G1 X12.000 Y60.000 Z-12.000 F150.000
G0 Z6.000
X30.000 Y60.000
Z-9.000                                 ; and again for every subsequent hole
G1 X30.000 Y60.000 Z-12.000
```

**After** — the retract plane is raised to the stock top, and the traverse moves to the clearance plane:

```gcode
M3 S18000.000
G0 Z6.000                               ; clearance plane first
X12.000 Y60.000                         ; traverse above the material
Z0.000                                  ; descend only to the stock top
G1 X12.000 Y60.000 Z-12.000 F150.000
G0 Z6.000
X30.000 Y60.000
Z0.000
G1 X30.000 Y60.000 Z-12.000
```

Both halves of the defect are visible in that diff: the descent stops at the surface instead of 9 mm inside it, and the first hole gains the `G0 Z6.000` → XY ordering the other holes already had (the extra line is the safe-Z entry marker, hence 13 moves vs 12).

The CAM panel's **Toolpath warnings** field shows the new message, where before there was none:

> Retract height (3) is below the top of the material; raised to 12 so the tool does not enter the part at rapid. Retract height is an absolute Z in project coordinates, not a distance above the stock.

## Incidental finding — worth its own issue

I first tried to show this in the 3D preview and could not, for a reason that is itself a defect: **a descending rapid is never drawn.** `buildToolpathOverlayLayers` splits rapid moves into two layers, and `Viewport3D` filters each one:

- `rapids` — `horizontalOnly`, i.e. `|from.z - to.z| < 1e-9`
- `retractions` — `retractOnly`, i.e. `to.z > from.z`

A rapid that goes *down* matches neither predicate, so it is dropped from the overlay entirely. Before and after this fix, the 3D viewport renders pixel-identical frames for the program above — the plunge-through-material move is invisible in the preview that exists to catch exactly that.

That reads against PROJECT.md's "machine output must be inspectable", and it is out of scope here. Happy to file it separately.

---

zabooma:
IMG_9355
IMG_9354
IMG_9356

PR #483, opened 2026-08-11
fix(viewport): draw descending rapids in both toolpath previews

### LINKED ISSUE #482: toolpath preview: descending rapids are drawn by neither renderer, so a rapid into material is invisible
Found while producing before/after screenshots for #480. Verified against `e19a1ff`.

## The gap

Both toolpath renderers split `rapid` moves into two mutually exclusive layers, and a rapid that **descends** matches neither predicate, so it is filtered out of the drawn geometry entirely.

`buildToolpathOverlayLayers` (`src/components/viewport3d/toolpathOverlay.ts`) declares:

- `rapids` — `horizontalOnly`, applied in `Viewport3D` as `Math.abs(move.from.z - move.to.z) < 1e-9`
- `retractions` — `retractOnly`, applied as `move.to.z > move.from.z + 1e-9`

A move with `to.z < from.z` satisfies neither, and there is no third layer. The 2D sketch canvas carries the same split independently, in `previewPrimitives.ts` — same two flags, same two predicates, same omission.

The direction-arrow code in *both* files classifies the same move differently: it computes `isRetraction = move.to.z > move.from.z` and treats everything else as a rapid gated by `visibility.rapids`. So within one file a descending rapid is a "rapid" for arrow purposes but belongs to no line layer. That inconsistency is the clearest statement of the defect — one of the two classifications is wrong, and it is the layer one.

## Why it matters

This is the preview that exists to catch unsafe motion, and the motion it cannot draw is a rapid going *into* the work.

Concretely, from the #480 fixture (12 mm stock, `retractHeight: 3`): before the fix the program opened `G0 Z-9.000` — a rapid 9 mm into the part — and after the fix it opened `G0 Z6.000` → XY → `Z0.000`. The 3D viewport renders **pixel-identical frames** for those two programs. The 9 mm rapid plunge is simply not in the scene.

In drilling the omission is partly masked, because the fed plunge that follows is collinear with the dropped rapid, so a vertical line appears either way — the operator just cannot tell which part of it was rapid. In plan view a purely vertical move projects to a point, so 2D is moot for that case; but a rapid that descends *while* travelling in XY is a visible diagonal in plan view and is dropped there too.

This reads against `PROJECT.md`'s "Machine output must be inspectable" and against the safety contract's "verify ... safe-Z behavior ... with focused tests".

## Suggested direction

Give descending rapids a home rather than adding a fourth toggle. Options, roughly in order of preference:

1. Widen `rapids` from `horizontalOnly` to "every rapid that is not a retraction". The visibility toggles stay as they are (`Rapids` / `Retractions`), the layer set stays at five, and the split matches the arrow code's existing `isRetraction` classification — which is already the correct one.
2. Keep three-way and add a `plungeRapids` layer with its own colour, so a rapid into material is visually distinct from a horizontal one. More faithful, but adds a toggle and a palette entry.

Option 1 is a small change to a shared declaration plus the two filter sites. Both renderers must move together, or 2D and 3D disagree.

## Tests

`toolpathOverlay.test.ts` already covers layer membership and visibility gating; it needs a case asserting a descending rapid lands in exactly one visible layer. The 2D side has `previewPrimitives.test.ts`. Worth asserting the property directly in both: **every `rapid` move belongs to exactly one layer, for all sign combinations of `to.z - from.z`** — the current code fails that for the negative case, which is how this stayed hidden.

## Scope

- `src/components/viewport3d/toolpathOverlay.ts` — the layer declaration
- `src/components/viewport3d/Viewport3D.tsx` — the 3D filter site
- `src/components/canvas/previewPrimitives.ts` — the 2D filter site
- `src/components/toolpathVisibility.ts` — only if option 2 is chosen

`src/components/**` is not a protected path, but the file set and likely diff size put this past the fast-lane budget, so: plan in the issue, approval, then implementation.

zabooma:
Closes #482.

Both toolpath renderers split `rapid` moves into a `rapids` layer taking only level travel (`horizontalOnly`) and a `retractions` layer taking only ascents (`retractOnly`). A rapid that **descends** satisfied neither predicate, so it belonged to no layer and was dropped from the drawn geometry — in the 3D overlay and, independently, in the 2D canvas.

The motion that could not be drawn is a rapid going into the work. Rendering the #480 fixture on either side of that fix produced **pixel-identical frames**: a 9 mm rapid plunge into a 12 mm part was simply absent from the preview whose job is to catch it.

## The change

One `zFilter` discriminant replaces the two independent booleans, evaluated through a single shared predicate:

```ts
export type ToolpathLayerZFilter = 'nonRetract' | 'retract'
```

Two booleans admit the state that caused this — a move matching neither. `'nonRetract' | 'retract'` partitions every rapid, making "lands in exactly one layer" structural rather than incidental. `nonRetract` now takes level travel and descents alike, so a descending rapid follows the **Rapids** toggle — which is what the direction-arrow code in both files already assumed. That disagreement between the arrow classification and the layer classification was the tell.

Beyond the predicate, two sources of drift are removed:

- **`previewPrimitives.ts` re-declared the same five layers inline** instead of consuming `buildToolpathOverlayLayers`. That duplication is the mechanism by which the two views diverged, and fixing only the predicate would leave the next change free to reintroduce it. The 2D renderer now consumes the shared declaration and maps layer key → its own stroke and dash, so styling stays per-renderer while there is one definition of what a layer *contains*. The shared list's order already matched the inline one exactly, so paint order is unchanged.
- **The arrow paths in both files carried private copies of the retraction predicate**, epsilon and all. They now call the shared one.

No new toggle and no palette change: the layer set stays at five and `ToolpathVisibility` is untouched. This was option 1 from the issue; option 2 (a distinct `plungeRapids` layer with its own colour) is more faithful but costs a toggle and a palette entry, and can be layered on later if a descending rapid deserves its own colour.

## Verification

`npm run build` green on 174 test files.

Tests state the **partition property** over all three signs of `to.z - from.z` rather than adding a case for the descending one — the old pair passed for level and ascending rapids, so only the exhaustive form catches a predicate set that stops covering the space. Verified they fail against the old predicate:

```
Assertion failed: descending rapid (plunge into material): expected exactly 1 layer to draw it, got 0 (none)
```

Coverage: the partition over all three signs plus sub-tolerance drift; a descending rapid following the Rapids toggle and not the Retractions one; `moveMatchesZFilter` never placing a move in both halves or neither; and, on the 2D side, a descending rapid that travels in XY — the diagonal shape, which is the case plan view can actually show — rendering under `rapids` and not under `retractions`.

Confirmed in the running app as well, with every layer except **Rapids** switched off, on a drilling program whose safe-Z → retract-plane descents are exactly this motion: before, only the horizontal traverse draws; after, the three vertical descents appear at the hole centres. A pixel diff of the default-visibility frames isolates the change to narrow vertical bands at those three centres, and nowhere else.

**`npm run test:e2e` was not run** — same sandbox limitation as #480: the environment's network policy denies `gc.zgo.at`, which `src/utils/analytics.ts` loads at runtime, and `e2e/fixtures.ts` fails any test with a console error, so every spec fails there regardless of the change. CI should run the gate normally. The in-app check above was driven with that one request stubbed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QH3gMNPcd395k4NGaYT3Bi)_


zabooma:
Superseded by #484 — identical commit, branch renamed from `claude/issue-479-71mx2r` to `fix/issue-482-descending-rapids` so the name matches the issue and the repo's `<type>/issue-<NN>-<slug>` convention. A PR's head branch cannot be changed after creation, hence the replacement rather than an edit.

Closing this one; review #484 instead.

---

PR #486, opened 2026-08-12
feat(cam): corner relief generator — dogbone / T-bone / longest edge

### LINKED ISSUE #203: Corner relief generator (dogbone / T-bone / longest-edge)
**Roadmap:** Phase 3 / P2 · _Extracted from `work/cad-enhancement-roadmap.md`._

A selected-corner relief generator sized from the **actual selected tool**.

- Dogbone, T-bone, and longest-edge relief styles.
- Operates on a selected inside corner; reuse the selection-driven + property-panel workflow.

zabooma:
Implements the approved plan on #203 (v5 as amended by v6). Closes #203

## What it is

A new operation-level parameter, `cornerRelief` — `none` (default) / `dogbone` / `t_bone` / `longest_edge` — on **Pocket, Edge Route (inside) and Edge Route (outside)**. Relief is cut as a **dedicated pass appended after the operation's main path**, stepping down at its own levels, on every convex corner of whatever region the operation clears.

One optional field with a `'none'` default: no `.camj` version bump, no migration, and a legacy operation is byte-identical to today.

## Why a separate pass

Both alternatives were tried on paper and both are wrong:

- **Region union silently no-ops.** Eroding a radius-`r` lobe by a radius-`r` cutter collapses it to a zero-area point Clipper discards — a dogbone unioned into the region produces a tool path byte-identical to no relief at all.
- **An inline wall detour breaks the standard workflow.** Rough big with radial stock and relief off, then finish small with relief on — the correct way to work, since the lobe must be sized to the tool that defines the final wall. Under that, the lobe is virgin material at full depth when the finish pass arrives, so an inline excursion is a full-width slot at full pocket depth with a small cutter.

## Geometry — touch the corner, no deeper

| style | excursion ends at |
| --- | --- |
| Dogbone | `C + b·r` |
| T-bone | `C + u·r` on the chosen edge |
| Longest edge | T-bone on the longer adjacent edge |

`r` is the exact limit, not a conservative choice. Against driving the centre to the corner, ending at touch costs **0.785 r²** of virgin material per corner per level instead of 2.787 r², an excursion of `r(1/sin(θ/2) − 1)` instead of `r/sin(θ/2)`, and a notch `r(1 − sin(θ/2))` deep past each wall instead of `r`.

There is no per-corner selection (regions already cover that) and, per v6, no clearance parameter — so the T-bone's edge choice is automatic and documented: plain `t_bone` takes the more X-parallel edge, `longest_edge` takes the longer one, where a `2r × r` slot does less damage.

## Which corners qualify

One sign test against which side of each loop the cleared material lies on, so `cornerRelief.ts` never learns which operation kind called it:

| operation | relieved corners |
| --- | --- |
| Pocket, Edge Route (inside) | convex corners of the boundary, plus reflex corners of islands |
| Edge Route (outside) | concave corners of the part outline |

Convex island corners are excluded — the cutter wraps them. Detection runs on the resolved Clipper region (what actually traps a cutter), so arcs are already flattened; two tests reject tessellation, and the turn threshold derives from `DEFAULT_FLATTEN_ARC_STEP` / `DEFAULT_FLATTEN_CURVE_SAMPLES` rather than being picked by eye.

## The guard

One general guard replaces the enumerated hazards: **the pass descends at a corner only when the emitted moves prove the operation cut there at or below the deepest relief level.** Descending into uncut material is the only way this feature hurts anyone, so it is checked against what the operation really emitted rather than inferred from settings. Without naming them it catches a `finishWalls`-off pocket, an outside corner under a tab, and trochoidal roughing whose channel never reaches the contour. Skipped corners warn by coordinate; corners whose adjacent edges cannot hold the notch get their own warning.

Tab footprints are additionally passed as keep-outs, because `applyEdgeRouteTabs` runs on the result *after* generation — at generation time the contour still reads as cut at full depth through a tab.

Levels come from `min(tool.defaultStepdown, tool.maxCutDepth)`, never `operation.stepdown`, which the panel hides and `generateFinishBandMoves` ignores on a finish pass and so can hold a stale value. A tool with no stepdown fails the relief closed rather than cutting one full-depth slot per corner.

Relief and `roundOutsideCorners` compose: on an outside route the latter rounds convex part corners, relief acts on concave ones. Disjoint sets, both may be on.

## Verification

`npm run build` green (docs check, lint, colour check, `tsc -b`, 175 test files, vite build).

New `cornerRelief.test.ts` — 44 tests. The central assertion is **coverage of the emitted swept envelope**, not "the region changed", which would pass on the silent no-op above. The same coverage model is run with the excursion ending at `1.2r` and must *fail* — that is what makes the touch-the-corner rule falsifiable rather than merely satisfied. Also covered: the qualification sets, arcs and fillets producing nothing, `'none'`/legacy identity, tool-derived stepdown and `maxCutDepth` clamping, every guard path, the rough-big-then-finish-small workflow, `optimizeLinearMoves` preserving the out-and-back (it survives only because the collinear merge requires `dot > 0`), and mm/inch plus a tool recorded in the other unit system. The suite was mutation-checked: reversing the corner sign test kills 24, moving the endpoint kills 14, disabling the guard kills 4, lowering the turn threshold below the arc step kills the arc criterion.

Four cases added to `camOperationSmoke.test.ts` so relief also goes through `postToolpath`'s #467 entry-invariant check on real emitted G-code.

CAM e2e (`camOperations.smoke.spec.ts`, 7 tests) run on an isolated port to confirm the panel renders with the new select.

**No new e2e spec**, stated rather than skipped silently: the change is an engine pass plus one CAM-panel select.

## Two findings recorded rather than smoothed over

- A T-bone's "confined to one face" guarantee is **exact only at 90°** — v6 measured it there. Away from the right angle the endpoint sits `r sin θ` from the other wall, so the guarantee degrades as `r(1 − sin θ)`. The panel text and booklet say the T-bone costs a slot rather than promising an untouched neighbouring face.
- Above about **105°** a T-bone's excursion runs more across the bisector than along it, so its capsule body sweeps over the corner whatever the endpoint distance — the endpoint distance stops being what determines coverage. The 1.2r falsification probe is therefore split by style, with that crossover asserted explicitly so it is documented rather than hidden.

## Not verified locally

`npm run check:gcode` exports the corpus but no controller interpreter is installed on this machine, so it passed without validating. Relief emits ordinary G0/G1 XYZ moves, nothing novel for a parser.


zabooma:
Followed up on the one gap the PR description named as unverified.

Built `gvalidate` (GRBL's own `gcode.c`) in this worktree and ran `npm run check:gcode` for real. The existing 15 export cases still pass with corner relief in the tree, and I added two relief cases — a relieved pocket with arc fitting on and off. Both pass, so 17/18 now verified against the strictest measured controller rules. The remaining gap is the Mach3 dialect, which no available interpreter speaks; that predates this PR.

Worth a case rather than an assertion: a relief excursion is an *exactly antiparallel* out-and-back repeated per level, which is the run shape an arc fitter can degenerate on.

Issue #524, opened 2026-08-15
e2e: feedColours slots-only test fails on main, and its vacuous-pass guard is what's failing


zabooma:
## Symptom

`e2e/feedColours.smoke.spec.ts:289` — *"slots-only pocket renders cut segments in
exactly one colour with the toggle on"* — fails on `main`.

Reproduced on a clean detached worktree at `origin/main` (`6aae5fc`), no other
changes:

```
1 failed
  e2e/feedColours.smoke.spec.ts:289:3 › slots-only pocket renders cut segments in exactly one colour with the toggle on
2 passed (1.2m)
```

The other two tests in the file pass. Run with:

```bash
PURECUT_E2E_PORT=1443 PURECUT_E2E_ISOLATED=1 npx playwright test e2e/feedColours.smoke.spec.ts --workers=1
```

## What actually fails

Not the colour assertion — the **vacuous-pass guard** at line 308. The test:

1. samples the canvas with the feed-colour toggle off,
2. toggles it on and asserts the pixel groups are unchanged (slots-only cuts carry
   no feed scale, so nothing may change),
3. then hides the cuts layer and polls for the canvas to differ from the off
   sample — proving step 2's sample contained real cut pixels rather than an empty
   canvas.

Step 3 is what times out. So the guard is doing its job: it is reporting that the
equality asserted in step 2 is currently passing **vacuously**, because the sampled
canvas does not contain the cut pixels the test believes it does.

That makes this more than a flake — while it is red, the two assertions it guards
are not actually constraining anything.

## Likely causes, not yet investigated

- the cuts layer toggle being clicked is not the one the sample depends on
  (`sketchItems(page).first()` assumes an ordering);
- the canvas not having settled before `sampleCanvas` runs, so both samples are
  of an empty canvas;
- the slots-only fixture rendering no cut segments at all in the sampled region.

## Not caused by, but found during, #518

Found while running the toolpath-surface e2e suite for
[#518](https://github.com/PureCutCNC/purecutcnc/issues/518). Confirmed identical
failure — same test, same assertion line — on the #518 branch and on pristine
`origin/main`, so it is filed separately rather than folded into that PR.

`npm run build` does not run e2e, so this does not gate the build.

zabooma:
## Why CI never caught this

CI (`.github/workflows/pr-check.yml`) runs e2e through three hand-maintained lanes
that enumerate spec files by name:

- `test:e2e:settings`
- `test:e2e:project-input`
[...]

PR #525, opened 2026-08-15
perf: invalidate toolpath caches per operation instead of on every sketch edit

### LINKED ISSUE #518: Perf: any sketch edit regenerates every operation's toolpath, including unaffected ones
## Problem

Every sketch edit regenerates **every** operation's toolpath, whether or not the edit
can possibly affect that operation. Adding a shape in an empty corner of the design,
deleting an unrelated feature, or even toggling a feature's visibility in the tree
throws away the whole toolpath cache and recomputes from scratch.

The user-visible result is lag while sketching: the sketch view stutters during node
drags, and visible toolpaths blank out and re-appear one per frame after every edit.

## Root cause

`isCacheHit` in [`src/app/useToolpathGeneration.ts`](src/app/useToolpathGeneration.ts) (line 112)
compares whole-array **identity** for every project collection:

```ts
export function isCacheHit(entry: ToolpathCacheEntry, operation: Operation, project: Project): boolean {
  return (
    operationComputationEquals(entry.operation, operation)
    && entry.stock === project.stock
    && entry.features === project.features   // ← any feature change at all
    && entry.tools === project.tools
    && entry.tabs === project.tabs
    && entry.clamps === project.clamps
  )
}
```

Every store action that touches features replaces the array
(`{ ...s.project, features: nextFeatures }` — `featureSlice.ts`, `featureGeometrySlice.ts`,
`treeVisibilitySlice.ts`, …), so `entry.features === project.features` fails for **all**
operations after **any** feature mutation. The operation's own `target.featureIds` is
already recorded on the operation and is simply not consulted for invalidation.

The comparison is deliberately coarse — `src/store/tabShape.test.ts:279` even asserts the
array identity must change so the cache invalidates — so this is a conservative design to
be replaced carefully, not an oversight to be patched.

## Evidence

**Provably-pointless invalidation.** Toolpath generation never reads `feature.visible` —
`grep -rn "\.visible" src/engine/toolpaths` returns only tab and clamp visibility
(`modelProtection.ts`, `clamps.ts`); the CSG *preview* filters on it, the toolpath engine
does not. Yet `setAllRegionsVisible` / folder and feature visibility toggles in
`treeVisibilitySlice.ts:45,68,95,121,150` all rebuild `project.features`, invalidating
every cached toolpath for zero change in output. Renames, colour changes and folder moves
are in the same category.

**Cost per invalidation.** Generating the operations of two committed engine fixtures, one
pass each (`npx tsx`, this machine):

| fixture | features | operations | generation |
| --- | --- | --- | --- |
| `issue-401-cone-finish.camj` | 1 | 2 × `finish_surface` | 217 ms + 298 ms = **515 ms** |
| `v-carve-noise-test.camj` | 3 | 2 × `v_carve_medial` | 184 ms + 94 ms = **278 ms** |

Those are 1–3 feature fixtures. Real projects have more features and more visible
operations, and the whole set is recomputed on every edit that touches any feature.

**Per-frame invalidation during drags.** Node drags call `moveFeatureControl` on every
`pointermove` (`usePointerGestures.ts:498,562,755`), so the array is replaced every frame
and the pipeline restarts every frame. Generation itself is synchronous and uncancellable
once entered — `startToolpathGenerationPipeline`'s `cancelled` flag only prevents the
*next* operation from starting — so a drag on a project with a 300 ms operation stutters
for the length of the gesture.

**Visible blanking.** `startToolpathGenerationPipeline` calls
`setToolpathMap(immediateResults)` (`useToolpathGeneration.ts:154`) with a map containing
only the cache *hits*, so every operation awaiting recompute is dropped from the map
immediately. Its toolpath disappears from the canvas and 3D view and returns one frame
later. This is what reads as "flicker" during editing.

## Why the obvious fix is wrong

"Only invalidate when a target feature changed" would produce wrong toolpaths. An
operation legitimately reads geometry beyond its target list:

- **Islands** — `resolver.ts:324` treats *any* `add` feature whose path intersects the
  target union as an island, with its own Z span.
- **Feature order** — `resolver.ts:351` (`expandedFeaturesInOrder`) and `regions.ts:106`
  make per-band topology order-dependent, so a pure reorder is a real change.
- **Tabs and clamps** — tabs intersecting the target union become obstacles
  (`resolver.ts:331`); clamps feed collision warnings (`clamps.ts`).
- **Stock-targeted surface operations** (`rough_surface`, `finish_surface`,
  `finish_surface_cleanup` with `target.source === 'stock'`) cut the whole model, so they
  genuinely depend on every solid feature. No narrowing is possible there and none should
  be attempted.

So the fix is a **dependency model**, not a target-id shortcut. A stale toolpath is
G-code that does not match the design — the invalidation predicate has to be conservative
by construction, with "unknown → invalidate" as the default.

## Proposed approach

**1. Diff features per row instead of per array.**
Keep `entry.features === project.features` as the fast path. When it fails, compute the
set of changed feature ids by row identity (`O(n)`, no deep compare — store mutations are
immutable per row) and decide per operation from that set.

**2. Ignore display-only feature fields.**
Mirror the existing `operationComputationEquals` pattern with a
`featureComputationEquals(a, b)` that compares only what the generators read, excluding
`visible`, `name`, `color`, and folder membership. A visibility toggle then produces an
empty changed set and zero regeneration. As with `operationComputationEquals`, the
exclusion list is a maintenance obligation: a new geometry-relevant field that is not
listed is a correctness bug, so it gets the same "any new field must be listed here"
comment and a guard test.

**3. Record each operation's read footprint at generation time.**
The generator already visits exactly the features it depends on. Have the resolver report
the feature ids it consulted, and store on the cache entry:
- `readFeatureIds` — the exact set consulted, and
- a **spatial footprint** (target-union bbox grown by tool radius, or stock bounds for
  stock-targeted operations).

Invalidate operation `O` when a changed feature is in `O.readFeatureIds`, **or** its bbox
(before *or* after the change) intersects `O`'s footprint — the second clause is what
catches a newly-added or newly-moved feature that `O` never saw but now overlaps. Feature
reordering, and any change that cannot be located spatially, invalidates everything.

**4. Coalesce during gestures.**
The store already signals an in-flight gesture via `history.transactionStart`. While a
transaction is open, hold heavy regeneration (optionally keeping the selected operation
live) and run it on commit, so a drag costs one regeneration instead of one per frame.

**5. Stop blanking the map.**
Keep the previous result in `toolpathMap` while a recompute is pending rather than
dropping it, so a stale path stays drawn until its replacement lands. The existing
`generatingOperationIds` spinner already communicates that it is being recomputed.

## Acceptance criteria

- Toggling a feature's visibility, renaming a feature, recolouring it, or moving it
  between folders regenerates **no** toolpaths.
- Editing, moving, or deleting a feature that is outside operation `O`'s read footprint
  does not regenerate `O`.
- Regeneration **does** still happen for: a change to a target feature; a change to an
  overlapping island/obstacle feature; a new feature that moves into the footprint; a
  feature reorder; any tab/clamp/tool/stock change affecting the operation; and any change
  to a solid feature when the operation targets stock.
- `src/store/tabShape.test.ts:279` (tab-array identity ⇒ invalidation) is re-expressed
  against the new predicate rather than deleted — the behaviour it protects must survive.
- **G-code byte-identity**: for each committed engine fixture, generated output before and
  after this change is identical.
- A node drag on a multi-operation project runs at most one regeneration per gesture, not
  one per `pointermove`.
- Visible toolpaths do not disappear from the canvas or 3D view while a recompute is
  pending.

## Non-goals

- Making individual generators faster — that is #263 (`buildBooleanModel`) and #517
  (engagement feed reduction).
- The CSG preview rebuild path (already decoupled from selection in #261/#262).
- Moving generation to a Web Worker. Worth considering separately; it addresses *where*
  the work runs, while this issue is about not doing the work at all.

## Risk

This is a safety-adjacent cache: a missed invalidation means the toolpath on screen — and
the G-code exported from it — no longer matches the design. Every widening of the cache
must land with the guard test that proves the corresponding invalidation still fires. When
the dependency of a change cannot be determined, invalidate.

Touches `src/engine/toolpaths/**` (protected path) ⇒ full lane: plan approval required
before implementation.

zabooma:
Closes #518

Operations regenerated their toolpath on **every** sketch edit, because
`isCacheHit` compared whole-array identity (`entry.features === project.features`)
and every store mutation replaces that array. Adding a shape in an empty corner,
toggling a feature's visibility, or dragging a node re-ran generation for every
visible operation — measured at 217 ms + 298 ms for the two operations in
`issue-401-cone-finish.camj`, a **one-feature** fixture.

Invalidation is now decided per operation, from what that operation actually reads.

## What changed

`src/engine/toolpaths/toolpathDependencies.ts` (new) answers two questions purely:

- **what changed** — `diffToolpathInputs` diffs two project snapshots by feature
  row, by referenced `FeatureDefinition` (geometry lives on the definition, so an
  instance row can be byte-identical while its profile changed), and by the global
  inputs every operation reads;
- **who cares** — `operationFootprint` / `operationAffectedByChange` give the world
  XY region a change must reach to matter to one operation.

`useToolpathGeneration` records the footprint on each cache entry at generation
time and consults it. `buildToolpathCacheEntry` is the single definition of an
entry, consumed by both the hook and the tests.

Scheduling (no effect on generated geometry): a pending recompute keeps its
previous result in `toolpathMap` instead of blanking the canvas, and
`deferGeneration` — driven by the store's open history transaction — coalesces a
drag gesture into one regeneration instead of one per `pointermove`.

## Four missed invalidations this also fixes

The old predicate was coarse on `features` but silently omitted real inputs, so
these could already leave a stale toolpath on screen before this PR:

| input | read by |
| --- | --- |
| `project.dimensions` | `resolveDimensionRef` → every feature's Z span |
| `project.featureDefinitions` | all feature geometry |
| `project.modelAssets` | imported-model features |
| `meta.maxTravelZ`, `operationClearanceZ`, `clampClearanceXY/Z` | `clamps.ts`, `geometry.ts` |

The clamp-clearance ones are safety-adjacent — they change collision warnings.

`meta` is compared through an explicit `MACHINING_META_FIELDS` allowlist, never
wholesale: `meta.modified` is rewritten by nearly every store action, so a wholesale
compare would invalidate on every mutation and silently undo the entire change.

## Deliberately still invalidating

`name` is treated as computation-relevant, not display-only, because it is embedded
in user-visible warnings (`drilling.ts`, `carving.ts` put it in
`warnings[].params.name`) — a cached result would otherwise show a renamed feature's
old name. Warnings could instead carry feature ids resolved at display time, which
would let renames stop invalidating; that touches every warning site and its i18n
params, so it is noted in the ledger as out of scope.

`visible`, `locked`, and `folderId` are exempt: `locked` and `folderId` have zero
reads under `src/engine/`, and `visible` is read only for tabs and clamps.

## Why the footprint is sound

Each verified against the generators, and recorded in the module docstring:

- an island or obstacle only counts if it intersects the target union
  (`resolver.ts`), so it lies inside the target bbox;
- safe/travel Z derives from the operation's **target** spans plus stock and
  `operationClearanceZ` — no distant feature raises a retract;
- protected-footprint paths only subtract where they meet the operation's own coverage;
- rest machining is materialized as region features at creation time, so there is no
  live operation-to-operation dependency.

Every unknown returns `bounds: null`, which invalidates: empty target list,
unresolvable row or definition, no profiles, missing tool, non-positive diameter.

## Verification

- **G-code byte-identity**: toolpath output dumped at the merge base and at the
  branch tip across the engine fixtures — 4 operations, 36,916 moves, identical. No
  generator file is modified; the only change under `src/engine/toolpaths/` is the
  new module plus one export line.
- **`npm run build`** green (185 test files).
- **e2e** `toolpathVisibility`, `camOperations`, `gcodeExport` — 13/13.
- **Mutation testing, 23 mutations across 7 slices.** Every guard was broken on
  purpose and confirmed to fail, in both directions — over-invalidation *and*
  under-invalidation, since only the latter ships a stale toolpath.

One mutation initially passed: replacing the recorded footprint with `bounds: null`
changed nothing, because the test file re-spelled the hook's entry construction and
so never constrained the write path. That produced the `buildToolpathCacheEntry`
extraction; the mutation now fails, as do variants that shrink the bounds or empty
the target set.

`e2e/feedColours.smoke.spec.ts:289` fails on this branch **and** on pristine
`origin/main` — same test, same assertion line — so it is filed separately as #524
and is not a regression here.


zabooma:
## S5 — user testing found the footprint margin was swallowing the benefit

Testing on a real project (`work/feed-reduction-test4.camj`, a pocket on a 4×3"
stock with a 0.25" tool) showed the operation still regenerating when a feature was
drawn outside the stock. Two causes, both measured.

### The margin was picked, not derived

`operationFootprint` grew the target bbox by `4 · toolDiameter + stepover`, justified
in the spec as "being generous costs only extra invalidation". That reasoning is
backwards for a performance fix — extra invalidation *is* the defect:

| | |
| --- | --- |
| pocket target | X 0.50 .. 3.50 |
| stock | X 0.00 .. 4.00 |
| margin | 1.320" |
| footprint | X **−0.82 .. 4.82** — 0.82" past the stock edge |

Nothing a user could draw near the part avoided regenerating, while every gate
stayed green.

The margin is now derived from what actually reaches past the target: an outside
edge route offsets by one radius and the cutter body extends another (1 diameter),
and `buildProtectedFootprintPaths` expands features by about a radius when clipping
coverage (half a diameter more). `stepover` is spacing between passes *inside* the
region and reaches nothing, so it is dropped.

**`2 · toolDiameter + trochoidalCutWidth + stockToLeaveRadial`** — ~1.33× the real
geometric reach, asserted by a test on the literal numbers so a future change fails
loudly.

### An unrelated tool import regenerated everything

`isCacheHit` still compared `entry.tools !== project.tools`. Every engine read is
`project.tools.find(t => t.id === operation.toolRef)` — ten call sites, no
exceptions — so it now compares only the operation's own tool row.

`tabs` and `clamps` deliberately stay whole-array identity: tab reads are not all
spatially filtered (`modelProtection.ts` iterates every tab, `edge.ts` passes
`project.tabs` wholesale for trochoidal), so narrowing them needs its own footprint
argument and is out of scope here.

### Verified on the reported project, in the reported scenario

| | before | after |
| --- | --- | --- |
| footprint | X −0.82 .. 4.82 | X 0.00 .. 4.00 |
| rect drawn just outside the stock | regenerates | **no regeneration** |
| unrelated tool imported | regenerates | **no regeneration** |
| island drawn inside the pocket | regenerates | regenerates ✓ |

Four mutations, all caught: restoring the old margin, dropping
`stockToLeaveRadial`, reverting tools to whole-array identity, and ignoring tool
changes entirely. Full `npm run build` green (187 test files).

The lesson worth keeping: a tolerance chosen "to be safe" in a performance fix needs
deriving and a test asserting its actual number, and the result needs checking
against a real project in the user's own scenario — unit tests, build gate, e2e and
G-code byte-identity were all green while the feature delivered nothing.

zabooma:
## S6 + S6b — non-target regions, and a latent construction bug

User testing: drawing a **region** over the operation area regenerated the toolpath
without changing it.

Correct — a region is a machining *filter*, and it can only filter an operation it
is a target of. Verified exhaustively rather than assumed: `splitFeatureTargets`
resolves only the ids handed to it and never scans `project.features`, and all 13
call sites pass `operation.target.featureIds`. `resolver.ts` filters
`selectedTargetFeatures`, `multiFeature` filters the target list, and
`isEdgeRouteTargetFeature` runs on already-resolved targets. Regions cannot enter as
islands (island discovery is `add`-only) or as protected footprints (`add`/`model`
only). The remaining region reads are the 3D preview and print/export — not
generation.

### The exemption did not fire, and construction had the same bug

S6 shipped the region skip guarded on the role holding in **both** snapshots. A
newly *drawn* feature does not exist in `previous`, so its status there is `null`,
the both-sides test fails, and it is never exempt — which is exactly the reported
case. Measured on the user's project: `isCacheHit = false`.

**Construction geometry carried the identical bug since S3a.** Its spec said
"construction geometry in both snapshots", the implementation matched, and its tests
covered only a feature present on both sides. A newly drawn construction feature
regenerated everything.

S6b replaces both with one rule — the role must hold on every side where the feature
*exists*, absence raising no objection:

```ts
function roleExemptEverywhere(previousStatus: boolean | null, nextStatus: boolean | null): boolean {
  return previousStatus !== false && nextStatus !== false
    && (previousStatus === true || nextStatus === true)
}
```

Added or removed while holding the role → exempt. Converted to or from the role →
still invalidates, both directions. Absent on both sides → falls through to the bbox
check, which is safe.

### Verified on the reported project

| drawn over the pocket area | before | after |
| --- | --- | --- |
| region | regenerates | **no regeneration** |
| construction | regenerates | **no regeneration** |
| `subtract` | regenerates | regenerates ✓ |
| `add` (island) | regenerates | regenerates ✓ |

Mutations caught: needing only the `next` side, removing the skip entirely, dropping
it from the whole-model branch, reverting to "role on both sides", and dropping the
definite-`false` guard. Full `npm run build` green (187 test files).

Clamps and tabs remain whole-array identity checks by decision; narrowing tabs needs
its own footprint argument, since `modelProtection.ts` iterates every tab and
`edge.ts` passes `project.tabs` wholesale for trochoidal.

### Process note

Three defects in this PR reached a merge with every gate green — the footprint
margin, the region exemption, and the construction exemption. All three were
specification defects: the workers built exactly what was specified. Mutation
testing validates the rule that was written, not the rule that was needed. Each was
caught only by replaying the user's scenario against their own project file, which
belongs in the definition of done for a behavioural slice ahead of the unit tests,
not after them.

zabooma:
## S7 — non-target line features

Drawing a line regenerated every operation. A line reaches an operation only through
its target list, verified the same way as regions:

- `resolver.ts:266/270/282` derive every line from the target list.
- `resolver.ts:380` iterates **every** feature, but the loop is gated three lines
  earlier by `activeBandFeatureIds` — active targets plus active islands — and island
  discovery filters `operation === 'add'`. A non-target line is never in that set, so
  the loop `continue`s before any line handling.
- `multiFeature.ts:75` tests `operation.target.featureIds.some(...)`.
- Lines are excluded from protected footprints (`add`/`model` only) and from the CSG
  solid model — `isSolid` says so outright — so a stock-targeted surfacing operation
  cannot see one either.
- `isEdgeRouteTargetFeature` never matches `'line'`. The `'line'` hits in
  `arcReconstruction.ts` are segment types, not feature roles.

Region and line now share one `TARGET_GATED_EXEMPT_ROLES` list rather than a third
near-identical branch. Construction stays out of it: its exemption is unconditional
(issue #199), while these two are target-gated — a targeted region or line **is** the
operation's input and must still invalidate.

### Verified on the reported project

Drawn over the pocket area, inside its Z band:

| | result |
| --- | --- |
| line | no regeneration |
| region | no regeneration |
| construction | no regeneration |
| `subtract` | regenerates ✓ |
| `add` (island) | regenerates ✓ |

Mutations caught: removing line from the role list, making the line predicate always
true (so the exemption stops being role-specific), and reverting `roleExemptEverywhere`
to "both sides". Full `npm run build` green (187 test files).

### Out of scope, recorded for later

A non-target `subtract` that eats part of an island does not change the pocket
toolpath, although it does change the CSG model — pocket + full island and pocket +
subtract-eating-half-the-island are byte-identical (198,377 bytes), where the
equivalent half-width island gives 217,844. Reproduces byte-identically on pristine
`origin/main` (`1f3043a`), so it predates this branch and is not a regression from it.
Cause: `resolver.ts`'s band loop gates on active targets plus `add`-only islands, and
its subtract branch additionally requires `targetIdSet.has(feature.id)`.

Left for separate investigation at the author's request; details and a probe caution
are in `planning/ISSUE_518_INTEGRATION_HANDOFF.md`.

Issue #497, opened 2026-08-14
Hybrid adaptive pocket clearing — engagement-controlled motion where offset rings fail (umbrella)


zabooma:
Keep the existing pocket — offset rings, rounded corners, slot-feed reduction — as the strategy for open material, and add engagement-controlled motion only where offset clearing measurably degrades. This issue holds the analysis and the tier order; children carry the individual capabilities.

**Status: backlog. Nothing here is approved or planned.** Each child writes its own plan and waits for approval before any code.

**Revised 2026-08-14 after an external review** (see the review record at the bottom). Four claims in the first draft were wrong and are corrected in place.

## Where the offset pocket degrades today

| # | Symptom | What happens now | Where |
| --- | --- | --- | --- |
| 1 | Inside corner on a ring: radial engagement spikes well past the stepover | Deliberately ignored — ring corners are "over-engaged at most transiently … and keep the normal feed" | `src/engine/toolpaths/pocket.ts:265` |
| 2 | Innermost loop of each section is a true slot | Feed cut to `pocketSlotFeedPercent` for the whole run | `src/engine/toolpaths/pocket.ts:436`, `src/types/project.ts:542` |
| 3 | Neck narrower than `2r + stepover`: rings collapse and the tool slots for a long distance | Same one reduced feed | `src/engine/toolpaths/pocket.ts:436` |
| 4 | Pinch corridor between an island and a wall | As #3, and the boundary geometry does not look narrow | — |

`roundOutsideCorners` is not a mitigation for any of these. It manages *stock left behind*, and the wall-adjacent root ring is deliberately kept sharp (`src/engine/toolpaths/pocket.ts:1261`).

The classifier is binary — engaged or not — so #2 and #3 pay one conservative feed across their whole length, and #1 pays nothing at all.

## The enabling primitive

Every tier needs one number: the angular measure of the cutter's **leading semicircle** sitting in uncut material.

The method is exact angular intervals — for each prior swept capsule in the existing spatial index (`src/engine/toolpaths/pocket.ts:314`), compute in closed form the arc of the cutter circle it covers, then union. Point-sampling the tool circle aliases exactly where the answer matters, and a whole-pocket raster is a lot of memory for a number that only needs to be right near the tool. Details and the conservative-bias rule live in #498.

[...]

PR #549, opened 2026-08-17
feat(pocket): tangential ring-to-ring link junctions (#545)

### LINKED ISSUE #545: Tangential ring-to-ring links in offset pocket clearing
Between one offset ring and the next, the tool steps across radially. That link makes two abrupt direction changes — one leaving the ring it was on, one picking up the next — and the machine has to decelerate at both.

Proposal: bring the link in **tangent** to the next ring, in the direction of travel, so the tool eases onto it instead of turning onto it.

## Why this is worth looking at

While measuring the corner work under #497/#499 I counted the junctions that force a real slowdown across the eight-fixture pack. **244 of them are not at corners.** They are ring-to-ring links and slot reversals. They did not move when the corner radius was tripled, and they do not move for any corner-geometry change, because they are not corner geometry.

That is a cost nobody has looked at, and it may be larger than the corner cost that #497 was opened for.

## Unlike the corner work, nothing here looks blocked

This is independent of everything #499 ran into:

- it does not need cleared material to detour into, which is what killed the corner unwind;
- it does not depend on detecting corners, which is what the qualifier could not do on rounded geometry;
- it does not change engagement, so #498's feed reduction is unaffected.

It is purely a linking change.

## Before implementing — read this

**Size it on real `.camj` files, not the synthetic fixture pack.** The pack misled the corner investigation twice (see the fixture-representativeness issue filed alongside this one). The 244 figure comes from that pack and should be re-measured on real parts before it is trusted.

Also worth checking early: whether a tangential approach changes what the link sweeps, and therefore whether it can leave material or cut into an unexpected place. The link currently runs through cleared material; a tangent arc bulges off that line.

## Status

Not sized, not planned. Filed for the record so the finding is not lost.

zabooma:
Closes #545

## What

The straight link between offset clearing rings made two sharp turns — one leaving the ring just cut, one entering the next. The link is now a **tangent S-curve**: it departs ring N along its travel tangent, curves out, runs a straight diagonal, curves back in, and arrives on a vertex of ring N+1 along ring N+1's own travel tangent — tangent at both ends, departure and arrival.

Key design points:

- **The arrival point slides.** The S lands on a chosen vertex of the next ring (which re-seams there), and the solver searches the ring's vertices within the length budget.
- **Solver:** arc-line-arc — for each arrival vertex and a sweep of middle directions (interpolated plus an absolute sweep around the bisector, so parallel tangents are reachable), one arc radius is pinned to the floor (0.25x tool diameter), the closure solves the other radius and the straight length; the shortest candidate whose path stays inside the cleared domain wins. Length budget 2.5x tool diameter.
- **Fallback:** when no S fits, the straight link stays — today's behaviour.
- The S replaces the straight link entirely; no corner fillets. The next ring re-seams at the arrival vertex, and the S's first segment continues the previous ring's closing cut tangent — the whole transition is one smooth flow.

## Control

New Operation field roundLinkCorners (additive schema, no migration): default true for new operations, backfilled on load like roundOutsideCorners (per the owner's decision on #545), with a checkbox beside roundOutsideCorners in CAMPanel (offset pockets only — parallel is a documented no-op). The engine reads absent as off, so raw/synthetic inputs stay byte-identical; parallel pattern, edge routes, v-carve and the entry handoff are untouched.

## Measured acceptance (production streams, scripts/issue-545-junction-census.ts)

| | tracked corpus | full corpus (incl. gitignored work/) |
| --- | --- | --- |
| sharp exit junctions | 146 -> 55 (-62%) | 1,409 -> 400 (-72%) |
| sharp entry junctions | 134 -> 59 (-56%) | 1,447 -> 483 (-67%) |
| ring corner junctions | 57 -> 62 (+5) | 2,451 -> 2,447 |

The tracked-corpus corner delta is the seam re-classification: the S re-seams each arrival ring, so a corner that sat on the seam counts as an exit junction in the legacy stream and as a ring corner in the S stream. It is pinned by a delta assertion (<= 5) on a tracked fixture, so it stays small and visible rather than drifting.

Remaining sharp junctions are the gate's legitimate rejections — wall-adjacent entries whose bulge would enter wall-side stock, sub-floor curvature on tightly-curved tessellation, and budget misses. The integration tests prove the rejection by re-running the solver with the domain removed, rather than asserting the count alone.

## Review fixes (after the PR review)

- **Domain gate now samples along every segment** at the floor-radius chord budget (with an absolute floor) — the straight middle, a single segment up to the whole length budget, was previously unchecked and a straddling concavity passed. Reproduced by the reviewer; now rejected, with a dense-resample test.
- **Boundary parity fixed:** the domain predicate no longer ray-casts its own vertices — boundary points of outer and island polygons count as inside (the ring paths ride the boundary), recovering roughly half the wall-adjacent arrivals.
- **Parallel lateral step fixed:** the middle-direction sweep now includes an absolute sweep around the bisector; the interpolated sweep collapsed for parallel tangents. A genuine parallel-side unit test is added (the previous test picked a corner arrival).
- **Already-tangent skip corrected:** it now skips only when both ends are smooth; previously it skipped when either end was shallow, leaving links with one 71-degree end straight.
- The missing **contiguity regression test** is added (every consecutive move must connect, across three synthetic fixtures and a tracked one).
- Stale fillet-design comments replaced; a Z-guard added on the spliced link; the checkbox is hidden for parallel pockets; the census lint cleaned.

## The disconnected-tangent bug (fixed earlier on this branch)

The intermediate fillet approach emitted arcs whose tangent points missed the truncated chords on curved rings (28 gaps per pocket, worst 1.5 mm, on work/rounded-corners-poc-test-1.camj). The S-link design removed that machinery entirely; the contiguity regression test now pins it.

## Tests

- tangentLink.test.ts — the solver: tangency at both ends, the lateral step, the parallel lateral step, along-segment domain sampling, boundary acceptance, domain rejection, budget, degenerates, determinism, option gating.
- tangentLinkIntegration.test.ts — the wired feature: sharp-junction reduction with gate-rejection proof (production solver, real ring polylines), stream contiguity, swept-coverage parity, wall/island domain respect, byte-identity (absent vs false, parallel, edge, v-carve), entry-handoff preservation, composition with roundOutsideCorners, slot-feed stamping on a genuinely slotting link, arc-run budget, corner-delta pin, load-time backfill path.
- engagementPocket.test.ts — never-raise matrix gains a roundLinkCorners row; the engagement-cache, index-work and arc-run guards keep measuring the legacy stream (their target is the engagement mechanism, not the link shape).

npm run build green; npm test — 192/192 test files pass.

## Notes

- No e2e added deliberately: the only UI change is one checkbox that mirrors the existing roundOutsideCorners control; the behavior lives entirely in engine logic covered by structural tests.
- The operation booklet does not yet show the new setting (kept out of scope; flagged as a follow-up).


zabooma:
## Review

Read against #545 and its full comment history (both default decisions, the six review points), plus #497/#498/#499/#185 for the tier context. Branch checked out at `6db47e1`.

### Verified independently

- `npm run build` green — docs, lint, i18n, tsc, **192/192 test files**, vite build.
- The headline reproduces. `npx tsx scripts/issue-545-junction-census.ts tracked` prints sharp exit `146 → 98`, sharp entry `134 → 91`, exactly as claimed.
- **Review point 5 is materially answered** even though only arc-*runs* were bounded. Posting the three tracked fixtures through `runPostProcessor` (generic definition), off vs on:

  | fixture | G-code bytes | recovered G2/G3 |
  | --- | --- | --- |
  | `pocket-feed-reduction` | 5,888 → 5,897 (+0%) | 41 → 44 |
  | `pocket-feed-reduction-2` | 12,064 → 13,330 (+10%) | 78 → 84 |
  | `pocket-feed-reduction-3` | 24,342 → 24,451 (+0%) | 46 → 50 |

  Arc fitting absorbs the tessellation, and recovered arcs go *up* in every case. Worth putting in the PR body — it closes the point rather than leaving it half-addressed.
- Contiguity: 0 gaps, worst 0.0000, across all five tracked `.camj` projects, on and off.
- `testSweptCoverageParity` is not vacuous — the S genuinely fires on all three synthetic fixtures (square 124 → 272 moves, circle 1318 → 1576, neck 68 → 216), so the "no new material touched" half is doing work.

The design is sound and the measurement discipline is the good kind — the census is committed, the tracked-corpus restriction discharges review point 6, and the numbers hold up. Five findings below; the first is the one I would treat as blocking.

---

### 1. Blocking — the domain gate samples only path vertices, so the straight middle is unchecked

`src/engine/toolpaths/tangentLink.ts:184`

The containment loop tests `candidate[step]` and `candidate[step + 1]` and nothing between them. The arcs are tessellated at 5°, so their chords are harmless. The **straight middle is emitted as a single segment** (`tangentLink.ts:105`) up to 2.5 × tool diameter long. Any domain concavity strictly between its two endpoints is invisible to the gate.

Reproduced — a small island expansion straddling the middle run of the S the solver picks on an otherwise open domain:

```ts
const island = [{x:0.85,y:1.95},{x:1.20,y:1.95},{x:1.20,y:2.60},{x:0.85,y:2.60}]
const domain = buildOffsetDomainCheck([{ outer: [{x:-30,y:-30},{x:30,y:-30},{x:30,y:30},{x:-30,y:30}], islands: [island] }])
const ring = [{x:0.5,y:1.2},{x:4,y:1.2},{x:4,y:3},{x:0.5,y:3}]
const a = 20 * Math.PI / 180
tangentSLink({x:0,y:0}, {x:Math.cos(a),y:Math.sin(a)}, ring,
  { minRadius: 0.25, maxLength: 10, isInsideDomain: domain })
```

Result: **the gate accepts it.** Every emitted vertex passes `isInsideDomain`. Densely resampling the emitted polyline at 0.002 puts **325 points inside the island** — first at `(1.028, 1.951)`, on the segment `(1.039, 1.526) → (1.000, 3.007)`, length 1.482. That is the cutter driving through expanded island stock. The identical mechanism applies to a concave outer feature — a stock finger protruding into the pocket, a keyhole — where the consequence is a gouge into the wall.

The domain check is the entire safety mechanism for this feature, and the hole sits exactly where the path is longest and least sampled.

`testWallAndIslandDomainRespected` cannot catch it: it iterates `[move.from, move.to]`, which is the same vertex-only sampling. So this is invisible on both axes — production and suite.

Fix is small: sample the straight run at the same chord budget the arcs already use, or clip it against the domain before accepting the candidate. Whichever, the test needs a matching along-the-segment assertion, otherwise the next variant reintroduces it.

### 2. `ring corners 57 → 57` contradicts the committed instrument — it is 57 → 62

The census printed by this PR's own script on the tracked corpus:

```
PRODUCTION roundLinkCorners=on: sharp exit 98 (was 146), sharp entry 91 (was 134), corners 62 (was 57)
```

The PR table and the plan comment both state `57 → 57`, and the acceptance criteria list **"ring corners unchanged"** as a gate. On the tracked corpus that criterion is not met.

The net is still strongly positive — exits + entries `280 → 189` — and the mechanism looks benign: the re-seam moves a ring corner off the seam, where it was previously counted as an exit junction, into the interior where it counts as a ring corner. So it is plausibly a re-classification rather than five new decelerations. But that is an argument to make explicitly with the number in front of it, not a reason to publish the number as unchanged.

`testSharpLinkJunctionsDisappear` asserts `cornerSharp` equality only on the synthetic circle, where it does hold, and `testRealFixtureBackfillPath` asserts nothing about corners — so the suite does not see this either.

Either correct the table and restate the criterion (e.g. "ring corners +5, accounted for by seam re-classification"), or assert the corner count on a tracked fixture so the claim is pinned wherever it lands.

### 3. The contiguity regression test does not exist

PR body:

> The S-link design removed that machinery entirely; the contiguity regression test pins it (every consecutive move must connect).

`src/engine/toolpaths/INDEX.md`:

> `tangentLinkIntegration.test.ts` — … stream contiguity (the reported disconnected-tangent regression) …

There is no such assertion in either test file. The only `contiguous` in `tangentLinkIntegration.test.ts` is a local variable inside `arcRunCount`, which is the arc-run budget helper.

The output *is* contiguous — I checked all five tracked projects, 0 gaps — so this is not a live bug. But the disconnected-tangent regression is the one defect on this branch that forced a redesign (28 gaps per pocket, worst 1.5 mm), and the thing claimed to pin it is not there. Both the PR body and the INDEX entry currently assert coverage that does not exist.

It is a four-line test.

### 4. The domain predicate is boundary-ambiguous, and that is what rejects wall-adjacent arrivals

`src/engine/toolpaths/tangentLink.ts:208`

Root rings are never smoothed (`smoothRadius && depth > 0`), so an arrival vertex on the wall-adjacent ring is **bitwise identical** to a vertex of the domain polygon itself. A plain ray-cast evaluated on its own vertex is a parity coin flip. Measured on a 64-gon domain built from `buildOffsetDomainCheck`:

```
vertices of the domain's own outer ring accepted by the gate: 32/64
edge midpoints accepted: 31/64
```

Split by which side of the polygon the vertex sits on, not by geometry. Deterministic, but arbitrary — and it mirrors under an x-flip of the part.

Two consequences:

- The PR's characterisation of the remainder — *"wall-adjacent entries whose bulge would enter wall-side stock"* — is not what the code decides for that class. Roughly half of them are rejected for being on the left-hand half of the ring.
- `assertSharpLinkJunctionsAreGateRejections` cannot separate the two. It re-runs the same predicate and requires it to agree with itself, so a parity rejection and a genuine bulge-into-stock rejection are indistinguishable to it. That was the falsifiability the plan's review point 3 asked for, and this class slips through it.

The integration test already concedes the point in its own comment — *"The wall ring is cut exactly on the domain boundary, so points ON the boundary are legitimate"* — and compensates with `pointOnPolys`. Production has no equivalent. An on-boundary tolerance in `buildOffsetDomainCheck` would make the gate mean what the PR says it means, and would recover roughly half the wall-adjacent arrivals for free.

### 5. The middle-direction sweep cannot reach across a parallel step

`src/engine/toolpaths/tangentLink.ts:150`

The candidate middle directions are `rot(exitTangent, phi * k / 20)` for `k ∈ [-20, 20]` — strictly interpolations between the exit tangent and the arrival tangent. When those two are parallel (the literal lateral step this feature is named for) the sweep collapses onto a single direction, `phi1 = phi2 = 0`, and every candidate is discarded before a radius is ever solved.

Measured against a densely tessellated parallel arrival ring one unit away, budget 10, `minRadius` 0.25:

| exit tangent vs ring tangent | −45° | −20° | −5° | −1° | 0° | +1° | +5° | +10° | +20° | +45° |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| result | NULL | NULL | NULL | NULL | NULL | NULL | **S** | **S** | **S** | **S** |

So the S fires only where the exit tangent already turns toward the next ring, by at least ~5°, in the correct direction. That is family A's infeasibility re-inherited through the search parameterisation rather than dissolved by the sliding arrival point — the free arrival helps, but it cannot help if no candidate middle direction points across the step.

This is efficacy, not correctness, and the measured 33%/59% reductions stand. But it bears on the **Known limitation** section: the three straight links on `work/reduced-feed-demo.camj` are attributed to pinch gaps below the 0.25×D arc floor, and this sweep collapses independently of any radius. Worth confirming which one is actually firing before that attribution is carried into the corner-rounding follow-up as a floor-tuning problem — if it is the sweep, retuning the floor will not move it.

The unit test named `testLateralStepSTangentAtBothEnds` does not cover this: on its fixture the solver picks arrival index 1, the corner at `(3, 1)`, where the ring tangent turns 90°. Arrival index 0 — the genuinely parallel side — yields no candidate. The parallel lateral step is currently untested.

---

### Minor

- **Stale comments from the withdrawn fillet design.** `pocket.ts:2395` — "fillet the two ring↔link corners with the largest radius the cleared domain admits" (the solver pins one radius to the floor and solves the other). `project.ts:561` — "replaced with tangent-arc fillets". `operationDefaults.ts:335` — "the fillet only reshapes how the tool travels through junctions it already sweeps", which is the justification for defaulting it on; the S can cut material the straight link did not, which is precisely why the domain gate exists.
- **`pocket.ts:1837` uses `linkMove.from.z`, not `.to.z`.** Unreachable today — pocket levels retract to safe Z, so `transitionToCutEntry` cannot hand this code a single ramping cut link — but if one ever reached it, the whole S would be emitted at the previous level's Z and the ring would resume at `z` with a silent Z gap. A `from.z === to.z` guard costs nothing and makes the assumption explicit.
- **The checkbox renders for parallel pockets**, where the option is a documented no-op (`testParallelAndOtherOperationsByteIdentical` pins that). Gating it on `pocketPattern === 'offset'` would stop it reading as a control that does nothing.
- **Operation booklet.** `report.ts:313` emits `roundOutsideCorners`; `roundLinkCorners` is absent. Acknowledged as a follow-up in the PR — noting it only because this is default-on and changes emitted motion, and the booklet is the operator's setup sheet.
- `scripts/issue-545-junction-census.ts:594` — `prefer-const` under `npm run lint:scripts`. Not a gate, so the build stays green.
- **The S budget (2.5×D) exceeds `maxLinkDistance` (1×D)** for the straight link it replaces. Safe given the domain gate, but it is an implicit widening of that bound and the ring re-seams up to 2.5×D from where `rotateContourToBestEntry` put it.

zabooma:
All five findings addressed on the branch (`b14c709`); re-measured, not just patched.

1. **Blocking — middle sampling:** the gate now samples every candidate segment at the floor-radius chord budget (absolute floor included so a degenerate `minRadius` cannot explode the count). Your island-straddling repro is a unit test now: the returned path is densely resampled and must never enter the island, and the predicate rejects island-interior points.
2. **Corners 57 → 62:** accepted as the seam re-classification you identified, and the PR/plan tables now state it. Pinned by a delta assertion (≤ 5) in `testRealFixtureBackfillPath` so it cannot drift silently.
3. **Contiguity test added** — `testStreamContiguity`, every consecutive move must connect across the three synthetic fixtures plus `pocket-feed-reduction`. The INDEX/PR claims now correspond to an actual assertion.
4. **Boundary parity fixed** — `buildOffsetDomainCheck` treats boundary points of outer and island polygons as inside (the ring paths ride the boundary); only strictly-inside island points reject. Unit test: all 64 vertices of the domain's own ring accepted (was 32).
5. **Parallel-step sweep fixed** — an absolute sweep around the bisector is added alongside the interpolated one; the parallel lateral step now fires. A genuine parallel-side unit test is added (the old one picked a corner arrival). The sweep was indeed what blocked two of the three straight links on `work/reduced-feed-demo.camj`; the third was the inverted already-tangent skip (it skipped whenever *either* end was shallow — a link with a 3° exit and a 71° entry stayed straight). All three are S-curves now; all nine links on that file are.

Minors: stale fillet comments replaced in `project.ts` / `operationDefaults.ts` / `pocket.ts`; `from.z === to.z` guard on the spliced link; the checkbox is hidden for parallel pockets; the census `prefer-const` fixed and the dead family-D plumbing removed; the 2.5×D budget relative to `maxLinkDistance` is documented at the call site.

Re-measured: tracked exits 146 → **55** (−62%), entries 134 → **59** (−56%), corners 57 → 62 (+5); full corpus exits 1,409 → **400** (−72%), entries 1,447 → **483** (−67%). `npm run build` green, 192/192 test files pass.

PR #563, opened 2026-08-19
Replace workspace layout presets with two panel pull-out handles (#557)

### LINKED ISSUE #557: Replace the four workspace layout presets with two panel pull-out handles
The four workspace layout buttons in the upper-right of the centre tab bar are compact, but they ask the user to decode four abstract three-bar glyphs into left/right panel combinations. They also sit nowhere near the panels they control — the button for hiding the far-left Project Tree column is pinned to the far right of the screen.

From the same UI/UX review as #556. Full notes live in `work/ui-ux-review-notes.md` in the primary checkout — `work/` is gitignored, so this issue stands alone.

## Today

- Four preset buttons `lcr` / `lc` / `c` / `cr`, each an abstract icon of three `<span>` bars, rendered at the far right of the centre tab strip ([AppShell.tsx:563](../blob/main/src/components/layout/AppShell.tsx#L563), styles at [layout.css:1314](../blob/main/src/styles/layout.css#L1314)).
- Their only distinguishing affordance is a hover tooltip: "Show left, center, and right panels", etc. ([appShell.ts:33-37](../blob/main/src/i18n/locales/en/appShell.ts#L33)).
- The preset drives one CSS class on `.app-body`, which swaps `grid-template-columns` and `display: none`s the hidden panel ([layout.css:694-812](../blob/main/src/styles/layout.css#L694)).
- The control is desktop-only — `{!tabletShell && …}`. Tablet already uses a different, spatial model: overlay drawers with edge toggles and a scrim ([AppShell.tsx:408](../blob/main/src/components/layout/AppShell.tsx#L408)).
- State is plain `useState` in [App.tsx:62](../blob/main/src/App.tsx#L62) — not persisted, so the choice resets on every reload.

## Problems

1. **Four symbols for what is really two booleans.** The four presets are exactly the cross-product of "left panel visible?" × "right panel visible?", but the user has to recognise each combination as its own glyph rather than toggling the thing they want.
2. **No spatial relationship to the target.** The control for the left column lives at the far right. Nothing about its position says which panel it affects.
3. **Tooltip-dependent.** Same failure mode as #556 — the meaning is only in hover text.
4. **Inconsistent with tablet.** Tablet already has understandable pull-out drawers; desktop uses an unrelated symbolic model for the same idea.
5. **Reads as tab-scroll arrows.** Sitting flush at the end of a tab strip, the group looks like tab overflow controls rather than panel visibility.

## Preferred direction

Replace the four presets with **two independent pull-out handles at the panel boundaries**.

Both handles live in the centre panel's tab strip — the one surface that is always visible — pinned to its **left and right corners**, each against the boundary of the panel it controls. They are **icon-only and fixed-width in every state and every language**, so nothing moves relative to the centre panel: the Sketch tab stays in its upper-left corner whatever the panel state is.

| Panel state | Left handle | Right handle |
| --- | --- | --- |
| Visible | `<<` — collapse it | `>>` — collapse it |
| Collapsed | `>>` — bring it back | `<<` — bring it back |

- The chevron always describes the motion of the panel edge: pushed away to collapse, pulled back toward centre to restore. It always describes the *next* action and flips with the state.
- The panel name lives in the tooltip and the accessible name — **Show project panel** / **Hide project panel** / **Show CAM panel** / **Hide CAM panel** — never as visible text, so the button's width is identical in `en`, `de`, `fr`, `es` and `zh-CN`.
- The collapsed state gets a distinct *resting* style at identical width (e.g. a subtle filled background) so it reads as "something is stowed here" rather than inert chrome. State is carried by fill and chevron direction, never by size.
- Remove the four-button group once the handles cover the same four combinations.

**Why tooltip-only is acceptable here, when #556 rejects it.** This control is desktop-only — it renders under `{!tabletShell && …}` ([AppShell.tsx:561](../blob/main/src/components/layout/AppShell.tsx#L561)) and tablet uses its own overlay drawers with their own toggles. Hover exists on every surface where this button exists, so the touch objection from the review does not bind.

### Alternatives considered and rejected

- **Vertical edge rail carrying the panel name** (JetBrains tool-window stripe). Killed by i18n: a fixed-width stripe cannot size for `CAM panel` / `CAM-Bereich` / `Panneau FAO` / `Panel CAM` at once, and rotating `CAM 面板` -90° is typographically wrong for a CJK reader — vertical CJK sets glyphs upright, not rotated. It also spends permanent width on a mostly empty bar.
- **A second row of chips above the tabs.** Costs canvas height permanently, and a chip directly above the Sketch tab in the same visual band reads as *more* tab-like, not less.
- **Label appears when collapsed, width reserved when open.** Either the tabs shift, or a dead gutter sits before the Sketch tab forever. Both break the fixed-position guarantee above.

**Placement caveat:** the notes suggest putting the right handle "immediately before the Operations tab", but the Operations/Tools tab strip lives inside `.panel-right` ([AppShell.tsx:625](../blob/main/src/components/layout/AppShell.tsx#L625)), which is `display: none` when collapsed — a handle in there would vanish with the panel it is meant to restore. The always-present centre tab strip is the workable home for both handles.

**Implementation note:** the CSS side is already a two-boolean model in disguise. `lcr` / `lc` / `c` / `cr` map one-to-one onto (leftVisible, rightVisible), so the existing `.app-body--*` classes and grid rules can stay exactly as they are — derive the class name from two booleans instead of storing a preset id. The change is the control surface and the state shape, not the layout engine.

## Interaction details

- The four `appShell.layout.*` preset strings are replaced by four state-specific ones. The button has no visible text, so its accessible name is the only thing a screen reader has to go on — name the panel, don't say "expand".
- **Don't give it a tab silhouette.** Sitting flush before the Sketch tab, a tab-shaped background would read as tab #1. Use a quiet icon-button treatment and a small gap so the two are visually separate control families.
- **Get it out of `role="tablist"`.** `.panel-tabs-header` carries `role="tablist"` ([AppShell.tsx:478](../blob/main/src/components/layout/AppShell.tsx#L478)) and today already wraps the layout preset group — a non-tab child of a tablist is an existing a11y bug that this issue should fix rather than inherit. Either move the buttons outside the tablist element or move the role onto an inner wrapper holding only the tabs.
- Keep a comfortably large pointer/touch target even if the visible handle is narrow.
- Preserve panel visibility across the Sketch, 3D and Simulation tabs unless a view has a strong reason to override it.
- On desktop, collapsing reclaims the space for the canvas (already true of the current grid rules). Tablet keeps its overlay-drawer behaviour but should share the same visual language.
- Don't style the handles like tab-scroll arrows; align each to its panel boundary so it reads as a drawer handle.

## Open questions (decide during planning)

- Should panel visibility persist across app restarts, or stay session-only as it is today? If persisted, where — `localStorage` or project meta?
- Does the left handle always hide Project Tree and Properties together? (They are one `.panel-left` aside with a `PanelSplit`, so today: yes.)
- Does the right handle always hide Operations and Tools together? (Same: one `.panel-right` aside.)
- Double chevrons `<<`, or a panel-with-chevron icon with a single arrow? Either way the glyph must be the same width in both states.
- How strongly should the collapsed state be marked? Enough to notice a panel is stowed, not so much that it reads as an error or a badge.
- Should the existing `.panel-resize-divider` pill in each gutter pick up any visual relationship to these buttons, or stay purely a resize grip?
- Should tablet's existing drawer toggles be re-skinned to match, or left alone in this issue?

## Acceptance criteria (subject to the plan)

- [ ] Two independent handles control left and right panel visibility; all four previous combinations remain reachable.
- [ ] Both handles sit in the always-visible centre tab strip, in its left and right corners, and remain visible and operable when their panel is collapsed.
- [ ] Each handle occupies identical width in both states and in all five locales; the Sketch tab's position relative to the centre panel never changes.
- [ ] The chevron direction and the accessible name both reflect the *next* action, and update when the state changes.
- [ ] The collapsed state is visually distinguishable from the open state without hover.
- [ ] The buttons are not children of a `role="tablist"` element.
- [ ] The four preset buttons and their `appShell.layout.*` strings are removed, not left alongside.
- [ ] Panel visibility survives switching between Sketch / 3D / Simulation.
- [ ] i18n keys updated across `en`, `de`, `fr`, `es`, `zh-CN`.
- [ ] e2e covers: collapse left → canvas widens and Project Tree is gone → restore from the handle; same for right.

## Notes

Related: #556 (sketch edit session control hierarchy) — same review, independent change. #170 covers the global toolbar redesign, not the panel chrome.

zabooma:
Closes #557

Replaces the four `lcr`/`lc`/`c`/`cr` preset buttons with two independent pull-out handles pinned to the left and right corners of the centre tab strip, per the approved plan in the issue.

## What changed

- **Control surface:** two icon-only chevron handles (`<<`/`>>`, new `chevrons-left`/`chevrons-right` icons) at the panel boundaries. Fixed width in both states and all locales — the Sketch tab never moves. Chevron + accessible name always describe the next action; the collapsed state gets a quiet filled resting background.
- **State:** panel visibility becomes two persisted booleans (`panel-visibility:left/right`, default visible) owned by `AppShell`; `App.tsx` drops the preset state and props entirely. The existing `.app-body--*` grid engine is untouched — the class is derived from the booleans; tablet pins to `lcr` and keeps its drawer model.
- **a11y:** `role="tablist"` moved from the strip header to an inner `.panel-tablist` wrapper holding only the three tabs, fixing the pre-existing non-tab-children-in-tablist bug.
- **i18n:** five `appShell.layout.*` keys removed; `appShell.panel.showProject/hideProject/showCam/hideCam` added in en, de, fr, es, zh-CN.
- **e2e:** new `e2e/panelHandles.smoke.spec.ts` in the `workflow-ui` lane — collapse/restore both panels, canvas reclamation measured against stage width, chevron/name/resting-style state flips, handles outside the tablist, visibility surviving Sketch → 3D → Simulation.

## Verification

- `npm run build` green (docs + lint + icons + tsc + 199 structural test files + vite).
- `test:e2e:workflow-ui` lane: 36/36 passed at the CI profile (2 workers, 1 retry).
- e2e coverage is deliberate: the change is rendered DOM + state persistence, covered by the new spec; no engine/gcode/format paths touched.

## Out of scope (flagged for follow-up)

- Tablet drawer toggles keep their current skin (noted in the plan).
- `.panel-resize-divider` pills unchanged, still pure resize grips.


zabooma:
CI note: the e2e lane failures are runner infrastructure, not this branch. Every failed attempt died in `npx playwright install --with-deps chromium` — apt downloads from `azure.archive.ubuntu.com` crawling at ~200 KB/s — before a single test ran; the lane's 10-minute timeout then cancelled the job. Evidence: `build` ✓, `fast-lane-guard` ✓, both `gcode-conformance` ✓; `Project input` passed on attempt 1 and `Settings` on attempt 3 with this exact code; the workflow-ui lane passes 36/36 locally at the CI profile (2 workers, 1 retry). Once the mirror recovers, rerunning the e2e jobs should go green.

PR #567, opened 2026-08-20
docs: add Scope Discipline rules and fence the perf-measurement section

zabooma:
## The problem

Three failure modes kept recurring, across models — DeepSeek does the same
thing on the same docs, which is what pointed at the docs rather than at any
one agent:

1. A question ("look at X", "why does Y") gets answered with an investigation:
   tables of estimated numbers, then a self-correction, then a re-derivation of
   the same numbers.
2. An approved plan gets re-analysed at implementation time, declared wrong,
   and quietly replaced with a different design.
3. Started work gets reverted and its issue closed on the agent's own
   judgment, with "decided it wasn't worth it" as the report.

## Where the docs caused it

**`Performance assertions` (`AGENTS.md`)** was the strongest driver. It is the
densest, most vivid prose in the file — measured tables, geometric mid-points,
verify-by-mutation, and at one point a literal self-retraction (*"The earlier
claim in this section was wrong"*). It is a recipe for one narrow act (writing a
timing assertion into a test) but said so nowhere, so it read as the house style
for what good work looks like here: produce numbers, tabulate, distrust the
previous figure, redo it.

**Nothing said an approved plan was binding.** The loop has
Plan → Approve → Implement but no rule for what to do when implementation
disagrees with the plan. With no rule, the analysis gets reopened — and having
just re-derived it, the agent "discovers" the plan was wrong.

**Nothing forbade abandoning.** R2 says *"closing a speculative item costs
nothing"*, which is correct for unstarted filings and was unscoped, so it also
licensed bailing on approved in-flight work.

**Questions had no lane.** "Every task gets an issue" — but a question is not a
task, and the doc never said so, so questions entered the intake/plan machinery
instead of just getting an answer.

## The change

New **`## Scope Discipline`** section in `AGENTS.md`, five rules:

- **A question is not a task** — answered in the conversation, in prose. No
  issue, no plan, no board card, no numbers.
- **One investigation pass** — read, then answer. No second pass auditing the
  first, no re-deriving a conclusion already reached this session.
- **Do not produce numbers you did not measure** — no estimated tables or
  invented ratios. Say what is true in words, or say you do not know.
- **The approved issue is the contract** — after approval, implement it. If the
  plan looks wrong, say so in two sentences and wait.
- **Abandonment is the user's call, never yours** — no unilateral reverts,
  closures, or "not worth building" verdicts on approved work.

Plus two targeted scope fences:

- **`Performance assertions`** now opens by stating it applies *only* when
  writing a timing assertion into a test, and licenses nothing else.
- **R2's** "closing costs nothing" is now explicitly limited to *unstarted*
  filings.

`CLAUDE.md` gets a short pointer, since it is the always-loaded entry point and
these rules only work if they are seen before the behaviour starts.

## Notes

Docs-only, no code paths touched. `npm run build` green locally.

Process deviation, called out deliberately: `AGENTS.md` is a protected path, so
this is normally full-lane (issue → plan → approval). The plan was reviewed and
approved in conversation, and the PR was asked to carry the explanation rather
than opening an issue for a fix to the issue ceremony itself.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -63,6 +63,43 @@ gh pr list --state merged --limit 50 --json number,title,labels,mergedAt
 
 Re-run the eligibility check against each labeled PR's merge base. If fast-lane PRs exceed a third of the window, the criteria are being stretched — tighten them.
 
+## Scope Discipline
+
+The loop above is for *work*. These rules keep it from being applied to things
+that are not work, and keep an approved plan from being relitigated. Each exists
+because its absence produced the same failure: a question answered with an
+investigation, an approved feature reopened as an argument, and effort spent
+proving a point nobody asked about.
+
+**A question is not a task.** "Look at X", "why does Y do this", "what do you
+think about Z" gets an answer in the conversation, in prose. No issue, no plan,
+no board card, no numbers. The loop starts when the user asks for the work, not
+when a topic comes up. Answering a question by opening an issue is the most
+expensive way to say a sentence.
+
+**One investigation pass.** Read what you need, then answer. Do not open a
+second pass to audit the first, and do not re-derive a conclusion already
+reached in this session. A push-back from the user or newly landed code is a new
+pass with a reason; nothing else is.
+
+**Do not produce numbers you did not measure.** No estimated tables, no invented
+percentages, no ratios assembled from guesses. With no measurement in hand, say
+what is true in words or say you do not know — an unfollowable table of figures
+is worse than the sentence it replaced. **Performance assertions** below is the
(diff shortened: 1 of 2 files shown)


zabooma:
Retroactive approval record: [#605](https://github.com/PureCutCNC/purecutcnc/issues/605).

This PR edited `AGENTS.md`/`CLAUDE.md` (protected paths) without an issue — acknowledged deliberately in this PR's description at the time, with the plan approved in conversation. Issue #605 now exists as the traceable record the workflow requires; the content of this PR stands unchanged. Going forward, changes to the ceremony itself get an issue like any other task.

PR #589, opened 2026-08-23
drilling: Retract Height is a distance above the material (format 3.1)

### LINKED ISSUE #481: drilling: Retract Height reads as an absolute Z but its label and icon promise a distance above the material
Follow-up to #479, promised in that issue's plan and deferred out of #480.

#480 closed the safety hole: a retract plane below the material surface is now clamped and warned about, so no drilling operation can rapid into the part. It deliberately did **not** touch what the field *means*, which is the other half of what the external reporter identified.

## The mismatch

`retractHeight` is stored and consumed as an **absolute project-space Z**. Everything the user sees says otherwise:

- The label is "Retract Height" — a height, i.e. a distance.
- `OperationParameterReference`'s `retractHeight` diagram draws the stock as a box and dimensions the arrow from the stock top up to a dashed retract plane. It depicts a relative distance.
- The default is `stock.thickness + 1`, which reads as "1 mm above the material" and only works because the stock's top happens to sit at project Z `= thickness`.

A user entering `2` reasonably means "2 mm above the part". They get project Z 2, which on any stock thicker than 2 mm is inside it. After #480 that is caught rather than cut, but the operator still has to learn that a field called "height" is an absolute coordinate.

## What makes this non-trivial

The obvious fix — store it relative to `project.origin.z` — is **not sufficient on its own**, which is why #480 clamped instead of reinterpreting.

`origin.z` is a free numeric field in the origin properties panel, and the presets include **"bottom left", which sets it to `0`**. Z-zero on the spoilboard is a normal setup. "2 mm above the origin" would then resolve to 2 mm above the table — inside a 12 mm part. So any relative reading still needs #480's clamp underneath it; the clamp is the invariant, the semantics are the ergonomics.

The reference point therefore wants to be the **material surface** (`max(stock.thickness, highest feature top)`, the same quantity #480 clamps to), not the machine origin.

## Scope

- `src/types/project.ts` — the field's documented meaning.
- `src/store/helpers/projectFormat.ts` + `src/import/camj.ts` — a migration, since the stored number changes meaning. Existing values are absolute Z; reinterpreting them relative would silently move every saved retract plane. Note the migration's failure mode is safe-side only when `origin.z >= stock top`, which is exactly what cannot be assumed (see above) — so this needs a real migration keyed on format version, not a permissive re-read.
- `src/store/helpers/operationDefaults.ts` — the default becomes a plain offset rather than `stock.thickness + 1`.
- `src/components/cam/CAMPanel.tsx` — the field's `min` becomes `0` again (a distance cannot be negative) once the value is relative.
- `src/engine/toolpaths/drilling.ts` — resolve to `materialTopZ + retractHeight`; #480's clamp stays as the backstop.
- `src/engine/operationBooklet/report.ts` — the booklet prints this value and would need the same reading.

`.camj` format and migrations are a protected path (`AGENTS.md` fast lane), so this takes the full lane: plan in the issue, explicit approval, then implementation.

## Alternative worth weighing first

Keep absolute semantics and fix the presentation instead — rename the field to something that reads as a coordinate, and redraw the parameter-reference icon to dimension from project Z0 rather than from the stock top. Cheaper, no migration, no compatibility risk, and it makes the existing behaviour honest rather than changing it.

I'd want that trade-off decided before any code moves: a migration on saved projects is a real cost, and the safety consequence is already handled.

## Acceptance criteria

To be written once the direction above is chosen.

zabooma:
Closes #481

## What

Drilling `retractHeight` is stored and consumed as a **distance above the material surface** (`max(stock.thickness, highest feature top)`) instead of an absolute project-space Z. A user entering `2` now gets "2 mm above the part" on any stock thickness and any origin preset — including Z-zero-on-the-spoilboard setups, which is what made a naive relative reading unsafe before #480 landed the clamp.

- **Format 3.1** (`LATEST_PROJECT_VERSION` bump + version union). Older files are migrated in memory on load (`migrateRetractHeightToRelative` in `projectFormat.ts`): stored Z minus the material top, clamped at 0. Values below the old surface land on the surface offset — exactly what the #479 runtime clamp already produced at generation time — so no saved project changes behaviour. The version gate makes the migration idempotent; ≥ 3.1 files are never rewritten.
- **Engine** (`drilling.ts`): resolves the stored offset against the material surface; the #479 clamp + warning stay verbatim as the backstop (they now fire only on negative distances, which the UI prevents). The warning reports resolved absolute Zs as before.
- **UI** (`CAMPanel`): field `min` becomes `0`; undefined-display fallback becomes the plain default. The parameter-reference icon needed **no** change — it already dimensioned stock-surface → retract plane, which is now what the number means.
- **Default** (`operationDefaults.ts`): plain `convertLength(1, 'mm', units)` instead of `stock.thickness + 1`.
- **Booklet**: prints the stored value unchanged; it is now the operator-measurable distance.
- Warning copy updated in all five locales (the "this is an absolute Z, not a distance" sentence flipped).

The project-wide feature-top maximum in the migration is a deliberate approximation of the engine's per-operation quantity: worst case a migrated offset comes out slightly smaller than it was, never below the surface (runtime clamp backstops).

## Verification

- New migration tests in `operationMigration.test.ts`: 3.0 absolute→offset (22 over 20 mm stock → 2), below-surface → surface offset 0, and a 3.1 round-trip guard proving no re-migration.
- `drillingRetract.test.ts` rewritten for relative semantics; the #479 invariant (no rapid endpoint where no fed move has cut) asserted unchanged over toolpath *and* posted program, plus a new test pinning the same retract plane from two different `origin.z` presets.
- `camOperationSmoke.test.ts` #479 case driven with a negative distance through `postToolpath`.
- Full `npm run build` green (docs/lint/e2e-lanes/colors/i18n/icons/tsc/204 test files/vite), re-run after rebasing onto current main.
- No new e2e spec: the only rendered change is one numeric input's `min`/fallback value; structural tests cover the behavior, deliberately.


zabooma:
## Review notes — actionable items

Reviewed at `3efd8d2`. The two blockers below were confirmed by running `normalizeProject` on constructed inputs, not by reading alone.

### Blocking

**1. The migration runs after `normalizeOperation` has already injected the *new* relative default** — `projectFormat.ts:351`

`normalizeProject` builds `operations` through `normalizeOperation` (line 533), which spreads `defaultOperationForTarget` — now `convertLength(1, 'mm', units)` after this PR — and only then calls `migrateRetractHeightToRelative` (line 537). So a pre-3.1 drilling op with **no** stored `retractHeight` gets the fresh 1 mm relative default re-read as an absolute Z and collapsed to 0.

```
3.0 file, 20 mm stock, drilling op with no retractHeight key  ->  retractHeight = 0
byte-identical file stamped 3.1                               ->  retractHeight = 1
```

The retract plane lands exactly on the material surface instead of 1 mm above it, and the 0 is written back on the next save. Fix: run the migration before `normalizeOperation` fills defaults, or only migrate values that were actually read from the file.

**2. The migration resolves against the project-wide feature max; the engine resolves per-operation** — `projectFormat.ts:342`

`migrateRetractHeightToRelative` uses `max(stock.thickness, highest z_top in the whole project)`, while `drilling.ts:539` uses `max(stock.thickness, highest top among *this operation's* targets)`.

```
18 mm stock + one non-target feature at z_top 30
3.0 op storing retractHeight: 20 (2 mm above the stock)  ->  migrated to 0
engine then resolves materialTopZ = 18 from the drill targets  ->  plane drops Z 20 -> Z 18
```

Rapids between holes now skim the stock surface with zero clearance, and the 0 persists. The PR describes this as "worst case slightly smaller, never below the surface", but the shortfall is unbounded — it is the full gap between the tallest feature and the drilled surface. Resolve against the same per-operation quantity the engine uses, or leave the value alone when the project-wide max exceeds `stock.thickness`.

**3. Neither failure mode is covered** — `operationMigration.test.ts:213`

`projectWithDrillingOp` always passes an explicit number, and `legacyProjectWithRecursiveOp` uses `defaultStock(200, 200, 20)` with a single feature at or below Z 20 — so `materialTopZ` is always just `stock.thickness` and the `project.features.reduce` at `projectFormat.ts:342` is never meaningfully evaluated. Both bugs above pass the suite green. Please add:

- `projectWithDrillingOp('3.0', undefined)` asserting the 1 mm default survives
- a fixture with a feature taller than the stock, asserting the stored clearance is preserved

### Should fix

**4. The warning reports absolute Zs while its own new sentence says the value is a distance** — `drilling.ts:553`

`retractHeight: -2` on 4 mm stock (asserted at `drillingRetract.test.ts:287`) renders as:

> Retract height (**2**) is below the top of the material; raised to 4. Retract height is stored as a distance above the material surface.

The operator sees `-2` in the CAM panel and `2` in the warning, and the trailing sentence contradicts the units of both numbers. Either interpolate `requestedRetractOffset` / the clamped offset so the numbers match the field, or drop the surface-relative sentence.

**5. `(issue #481)` shipped inside the translated warning text** — `src/i18n/locales/*/warnings.ts`

`grep 'issue #' src/i18n/locales/en/*.ts` shows this is the only tracker reference in any translated string in the repo — every other hit is a code comment. A CNC operator sees an issue number they cannot open. The provenance already lives in the code comment at `drilling.ts:531`; please drop it from all five locales.

**6. `min={0}` makes a zero-clearance retract plane silently legal** — `CAMPanel.tsx:1677`

Combined with the migration's `Math.max(0, …)` floor, "retract exactly at the material surface" is now reachable, persisted, and unwarned — the warning only fires on `requestedRetractOffset < 0`. `public/examples/t-style-body.camj` is a live case: it stores `1.5394` on a 1.75" stock, previously clamped-and-warned on every generation, and now migrates to `0` with no warning at all. Consider flooring the input at a small positive clearance, or warning when the resolved plane lands exactly on the surface.

**7. Booklet prints `?? 0` where the engine falls back to `1`** — `operationBooklet/report.ts:361`

Under the new semantics "0 mm" reads as *on the surface*, while the program actually retracts 1 project unit above it. Since this number is now the operator-measurable distance, an incorrect value is worse than the old meaningless absolute-Z 0. Use the engine's fallback.

**8. Three defaults for the same value disagree on units** — `drilling.ts:537`, `CAMPanel.tsx:1673`, `operationDefaults.ts:366`

`defaultRetractOffset = 1` and `operation.retractHeight ?? 1` are raw project units; `operationDefaults` uses `convertLength(1, 'mm', units)`. In an inch project that is 1″ vs 0.0394″. `normalizeOperation` currently masks it by filling the field on load, so it is latent rather than live — but the PR rewrote all three lines and left them disagreeing.

### Minor

**9. NaN path** — `projectFormat.ts:346`. The reduce seeds on `project.stock.thickness` with no finite check, and `loadProject`/`openProjectFromText` apply `defaultStock` *after* `normalizeProject`. A file whose stock omits `thickness` gives `Math.max(undefined, …) = NaN`, `Math.max(0, r - NaN) = NaN`, and `next === operation.retractHeight` is false for `NaN`, so it commits — `drilling.ts:541` then posts `Z NaN`. The `z_top` branch already guards with `Number.isFinite`; same guard on the seed.

**10. Duplicated version parsing** — `projectFormat.ts:336`. The `split('.')` + `Number.parseInt(…) || 0` pair is verbatim `isProjectVersionNewerThanSupported`'s `parse` closure (`types/project.ts:1452-1455`). Two copies of the version-comparison rule to keep in sync at every bump; worth exporting a shared `parseProjectVersion`.

**11. Stale doc gloss** — `ARCHITECTURE.md:52`. The opening still reads "`Project.version` `3.0` is the first strict lightweight-instance format (`LATEST_PROJECT_VERSION` in `src/types/project.ts`)", pointing the reader at a constant that is now `3.1` and contradicting the new sentence three lines later.

**12. Hardcoded `'3.1'` in four test files** — `featureReferencesMigration.test.ts:189`, `constructionWorkflows.test.ts`, `machineSnapshot.test.ts`, `e2e/featureReferences.smoke.spec.ts`. `projectFixtures.ts` was changed in this same PR to import `LATEST_PROJECT_VERSION`. Several of the assertion messages already say "the current format" rather than naming a version, so the literal is incidental — those should use the constant; keep a literal only where pinning a specific version is the point.

### Verified as correct

- The parameter-reference icon genuinely already dimensions stock-surface → retract plane (`OperationParameterReference.tsx:141`), so leaving it unchanged is right.
- `min` on `DraftLengthInput` *is* enforced on commit (`CAMPanel.tsx:169`), so "the UI prevents a negative" holds.
- Idempotency across `normalizeProject`'s non-file callers (undo/redo, `saveProject`, templates, `newProject`) is sound — every one of them carries a `3.1` stamp or an empty `operations` array.
- `.camj` merge import is *improved* by this change: a relative distance survives being merged into a project with different stock, where an absolute Z did not.

zabooma:
All twelve actioned in 608c1b8. Dispositions where you offered a choice:

1. **Gated on file-read values** (your second option). `normalizeProject` snapshots which operation positions carry a numeric `retractHeight` in the raw input and passes that set down; untagged positions keep normalizeOperation's injected relative default. Your repro case now yields 1.
2. **Per-operation resolution**, but without duplicating the engine's read model: the surface is `max(stock.thickness, serialized z_top of this operation's target features)` — instance Z bounds are authoritative, so no geometry resolution needed. Your 18 mm + tall-feature case migrates to exactly 2. Targets whose top can't be determined (constraint-linked DimensionRef tops; outright missing targets can't reach the migration because normalizeOperation swaps invalid drilling targets for their fallback first) are left untouched — resolves high under the new reading, capped at clearance, never skims.
3. Added all three requested cases plus two more: no-key → default survives; tall bystander AND tall-as-target preserve the offset; DimensionRef-top target left alone; malformed stock skips migration entirely (`operationMigration.test.ts`, 11 cases).
4. Warning interpolates `requestedRetractOffset` / the clamped offset, so panel and warning show the same units.
5. Tracker reference dropped from all five locales.
6. Took the **warning** branch: the engine now warns at ≤ 0 ("at or below"), which also restores the warn-per-generation behaviour `t-style-body.camj` used to have after its value migrates to 0. Kept `min={0}` — zero is enterable but never silent.
7. Booklet falls back to `defaultRetractOffset(units)`.
8. One definition, `defaultRetractOffset(units)` in `types/project.ts` (1 mm / 1/25.4 inch), shared by seed, panel fallback, and engine fallback.
9. Migration early-returns when `stock.thickness` isn't finite; test drives it.
10. Exported `parseProjectVersion()`; both call sites share it.
11. Gloss reworded to name 3.0 as historical and point at the constant for current.
12. Structural save-stamp assertions use `LATEST_PROJECT_VERSION`; the e2e spec imports it from `src/types/project`. The one remaining literal is `featureReferencesMigration.test.ts:269`, which pins what's on disk at serialization time — that pin is the point of the assertion.

Full build green locally (incl. i18n canonicalization via `check:i18n -- --fix`), touched e2e spec verified against an isolated dev server before push.

PR #592, opened 2026-08-23
perf: fine top rungs on the engagement feed ladder

### LINKED ISSUE #591: Engagement feed buckets are too coarse near full feed: 3.6% over-engagement charged 8%
## Problem

`engagementFeedScale` (`src/engine/toolpaths/engagement.ts`) quantizes the feed
scale into `ENGAGEMENT_FEED_BUCKET_COUNT = 6` evenly spaced rungs and rounds
**down**. It leaves the top rung the moment engagement exceeds nominal by more
than `ENGAGEMENT_ESTIMATE_EPSILON` — 8.4e-8 rad, a floating-point-dust deadband
— so there is nothing between 1.00 and the next rung.

At a 60 % slot feed the rungs are 1.00 / 0.92 / 0.84 / 0.76 / 0.68 / 0.60. A cut
whose true continuous scale is **0.991** is therefore charged **0.92**: a 0.9 %
modelled reduction billed as an 8 % feed cut.

## Where it bites

Constant-engagement curved laps — which is what `seeded_offset` (#554) is made
of. A concentric lap grown outward by one stepover removes an annulus exactly
one stepover wide, but its contact arc is wider than a straight cut at the same
stepover, because the previously cut boundary is concave toward the tool. That
excess is real, and it decays like 1/radius:

| lap radius | engagement | vs nominal (1.2025 rad) | continuous scale | charged |
| --- | --- | --- | --- | --- |
| 0.615 in | 1.4760 | +22.7 % | 0.9436 | 0.92 |
| 1.335 in | 1.3307 | +10.7 % | 0.9736 | 0.92 |
| 2.055 in | 1.2864 | +7.0 % | 0.9827 | 0.92 |
| 3.975 in | 1.2462 | **+3.6 %** | **0.9910** | **0.92** |
| → ∞ (straight) | 1.2025 | 0 % | 1.0000 | 1.00 |

Measured on `work/big-pocket-test.camj`, op0047 (Ø0.75 in endmill, 0.240 in
stepover, feed 55 ipm, `pocketFeedReduction: engagement`,
`pocketSlotFeedPercent: 60`), on `perf/issue-576-seed-handoff`:

| stream | at 1.00 | at 0.92 | below |
| --- | --- | --- | --- |
| seed circles (352.3 in) | 8 % | **82 %** | 10 % |
| offset rings + excursions (211.1 in) | 53 % | 4 % | 43 % |
| plain `offset` pattern, whole stream | 78 % | 1.2 % | 21 % |

The plain `offset` pattern is almost untouched — its reductions sit in the deep
rungs (8.9 % at 0.70–0.79, 9.5 % at 0.60–0.69), which are corners and slot-like
stretches that genuinely should slow down. This is specifically a top-of-range
problem, and it lands on curved constant-engagement work.

Size of it on that file: if the 289.1 in currently pinned to 0.92 ran at full
feed, 0.46 min comes back — enough to move `seeded_offset` from 4.5 % behind
`offset` to level. That is arithmetic on measured lengths, not a prediction
about any particular fix.

## Direction (not yet a plan)

**A continuous/linear scale is not the answer.** The quantization is
load-bearing: arc fitting refuses to join moves whose `feedScale` differs at
all, so a continuous scale shatters every arc run into linear moves. On the
file above that is the 352 in of seed circles — emitted as exact circles
precisely so `arcReconstruction` recovers them as true G2/G3 — turning into
thousands of G1 segments, plus a feed change per segment for the controller.

**Uniformly more rungs is the wrong shape too.** It halves the spacing at the
bottom, where the pricing is already fine, and every extra rung is another
place a path can fragment.

**Non-uniform rungs, fine near 1.0 and coarse near slot** — e.g.
1.00 / 0.98 / 0.96 / 0.92 / 0.84 / 0.72 / 0.60. The error is concentrated at the
top (a 3.6 % over-engagement charged 8 % is a 2x overcharge; a 40 % one charged
44 % is noise), and fine rungs near 1.0 fragment nothing, because that is
exactly where engagement is constant along the path — a seed lap is one
engagement value end to end and gets one scale either way, just a better one.
Corners stay coarse.

Keep as they are:

- **The deadband.** 8.4e-8 rad is the right guard for estimator noise; the
  problem is what you fall to on crossing it, not where the crossing is.
- **Rounding down.** With a 2 % rung at the top the conservatism costs nothing.

## Risk to design around

`EngagementFeedQuantizer`'s hysteresis is `DEFAULT_HYSTERESIS_FRACTION = 0.25`
**of the bucket width**. Narrower top rungs narrow the anti-chatter margin there
too, so the feed could flutter between 0.98 and 1.00 along a ring whose
engagement wanders. That margin likely has to become absolute rather than
proportional. The minimum-fragment rule (ring perimeter / 8) wants the same
look.

## Not in scope

The engagement model itself is not wrong here. A 3.6 % wider contact arc really
is 3.6 % more load — the cut genuinely is slightly heavier than the stepover the
user dialled in. This issue is about pricing that correctly, not about exempting
curved paths.

## Acceptance criteria (draft)

1. An outer seed lap at ~3.6 % over nominal is charged within a few percent of
   its continuous scale, not a full rung below it.
2. Arc runs do not fragment more than they do today — `seeded_offset` on the
   fixture above still emits its laps as single arc runs.
3. No feed *increase* anywhere: every path that is charged a reduction today is
   charged at least as much of one.
4. Deep-rung behaviour (corners, slotting) is unchanged.
5. `npm run build` green, with the bucket ladder and hysteresis covered by
   `engagement.test.ts`.

zabooma:
Closes #591

## What

Replaces the engagement feed ladder's 6 uniform rungs with an **8-rung non-uniform ladder**: the issue's fine top rungs *plus every existing rung verbatim*. Defined as cumulative drop fractions `F = [0, .05, .10, .20, .40, .60, .80, 1]` of `(1 − slotScale)` — scale-invariant across any `pocketSlotFeedPercent`.

At a 60 % slot feed: `1.00 / 0.98 / 0.96 / 0.92 / 0.84 / 0.76 / 0.68 / 0.60`. The issue's measured case — a curved lap +3.6 % over nominal, continuous entitlement 0.991 — now charges **0.98** instead of falling a whole rung to **0.92**.

## How

- `engagement.ts`: ladder table + `engagementFeedRungs(slotScale)` as the single source of truth; round-down scan keeps the old `1e-12` boundary snap (straight offset rings hit boundaries analytically exactly).
- Hysteresis rise margin = max(¼ × target-rung width, absolute floor at the uniform ladder's share of the drop), **capped at half the climbable gap** so every rung stays reachable — without the cap, a pocket that touched one corner ran its following straight stretch at slot feed forever.
- One-rung merge admissibility (`fragments()` + pocket Pass D) judged by shared rung-index predicate instead of uniform bucket width.
- `palette.ts` consumes `engagementFeedRungs` so the colour ramp cannot desync from emitted scales (`feedColourStep` step == engine rung index).

## Acceptance criteria

1. ✅ top-range pricing within ~1 % of continuous (new §5d test pins 0.991 → 0.98)
2. ✅ no extra arc-run fragmentation — `seeded_offset`-style laps are one engagement value end-to-end; the S9 arc-run bound (+25 % / longest ≥ 50) passes unchanged
3. ⚠️ restated per the plan comment: round-down preserved everywhere (≤ continuous + dust); no hysteresis/merge decision resolves upward. The literal "no feed increase" wording contradicted criterion 1 and was flagged on the issue
4. ✅ deep-ladder freeze test: all rungs ≤ 0.92 byte-equal to the old uniform six
5. ✅ `npm run build` green; new assertions verified by mutation (floor removed → flutter test fails with 13 alternating fragments; uniform ladder reverted → pricing test fails)

## Tests

- New: §5d top-range pricing + mid-boundary round-down walk + deep-ladder freeze; §5e no-flutter guard.
- Rewritten against the engine ladder: §5 membership, §5c exact-set, ramp tests (now engine-derived rather than a formula oracle), legend step values, pocket carve-out predicate (rung-index distance).
- All pre-existing quantizer fixtures land on identical rungs — deep behaviour untouched.

- Review pass: restored `BUCKET_CASES_40` in the ramp test as an independent oracle (the engine-derived assertions alone compare the module to itself); widened §5d's round-down walk to the 1.0/0.98 boundary #591 adds.

**e2e updated, not added:** `e2e/feedColours.smoke.spec.ts` asserts the legend's rendered text, and the finer ladder changes it — the 75 % fixture now lists `100/99/98/90/85/75` and the mixed-pocket union gains the two fine top rungs. Two of its five tests were failing before that update. Legend labels also escalate decimal places until distinct (`toolpathVisibility.ts`), because at high slot feeds several rungs round to the same whole percent (slot 99 % needs two decimals). No menu, dialog, or boot-path change, so no new spec.


src/engine/toolpaths/engagement.ts
```diff
@@ -765,7 +879,20 @@ export class EngagementFeedQuantizer {
...
+    // stretch after a corner recovers onto the fine top rungs.
+    const margin = rawScale >= 1
+      ? 0
+      : Math.max(
+          this.hysteresisFraction * this.rungWidth(rawScale),
+          Math.min(
+            MIN_HYSTERESIS_DROP_FRACTION * (1 - this.slotScale),
+            this.gapAbove(rawScale) / 2,
+          ),
+        )
```
zabooma, line 915: **The climbability cap only guards the floor term.**

The class doc (line 752) says "The floor is capped at half the gap to the rung above so every rung stays climbable" — but `Math.min(..., gapAbove/2)` wraps only the `MIN_HYSTERESIS_DROP_FRACTION` term. The `hysteresisFraction * rungWidth` term is unbounded, so the invariant holds only for the default 0.25:

```
h=0.25: rungs unclimbable  []
h=0.5 : rungs unclimbable  [0.92]
h=1   : rungs unclimbable  [0.98, 0.96, 0.92, 0.76, 0.60]
```

Measured at `{ nominal: π/2, slotScale: 0.6, minFragmentLength: 0, hysteresis: 1 }`, a rising sweep up to continuous 0.9599 never leaves 0.84.

Capping the whole expression instead is byte-identical at `hysteresis: 0.25` (verified at every rung for slot 1–99%) and makes the invariant hold for any fraction:

```suggestion
    const margin = rawScale >= 1
      ? 0
      : Math.min(
          this.gapAbove(rawScale) / 2,
          Math.max(
            this.hysteresisFraction * this.rungWidth(rawScale),
            MIN_HYSTERESIS_DROP_FRACTION * (1 - this.slotScale),
          ),
        )
```

Separate symptom of the same expression, no clean mechanical fix: the floor also overrides a *low* configured fraction, so `hysteresis: 0` can no longer produce a zero margin. Same setup with `hysteresis: 0` refuses a rise from 0.96 to 0.98 at continuous 0.985 (margin floors at 0.01); the pre-#591 code rose immediately. Nothing in the tree passes a custom `hysteresis` today, so this is latent — but the option now means "lower bound on the margin", not "the margin". Worth either scaling the floor by `hysteresisFraction / DEFAULT_HYSTERESIS_FRACTION` or saying so in the option's doc.

src/engine/toolpaths/engagement.ts
```diff
@@ -78,13 +78,62 @@ const ZERO_LENGTH_EPS = 1e-9
...
+/**
+ * The emitted rung set for a slot scale, descending from full feed (index 0)
+ * to `slotScale` (last index). Single source of truth: the theme colour ramp
+ * and the structural tests consume this instead of re-deriving the arithmetic,
+ * so what they render and assert can never desync from what the engine emits.
+ */
+export function engagementFeedRungs(slotScale: number): readonly number[] {
+  const clamped = Math.min(1, Math.max(0, slotScale))
+  const drop = 1 - clamped
+  return FEED_RUNG_DROP_FRACTIONS.map((fraction) => clamped + drop * (1 - fraction))
```
zabooma: **`engagementFeedRungs` allocates a fresh array on every call, and it now sits on two hot paths.**

The pre-#591 code precomputed `bucketWidth` once and did pure arithmetic. Now every call `.map`s a new 8-element array, and it is called:

- once per engagement sample from `engagementFeedScale`,
- up to **five** more times per rise check — `rungWidth` and `gapAbove` each call `engagementFeedRungs` directly *and* `rungIndexOf`, which calls it again,
- once per cut move from `feedColourStep` on the render path.

Measured on this branch, 2M calls each:

| | this branch | pre-#591 arithmetic form |
|---|---|---|
| `engagementFeedScale` | 210 ms | 17 ms (~12×) |
| `feedColourStep` | 258 ms | 17 ms (~15×) |

`feedColourStep` runs per cut move in `previewPrimitives.ts:646` on every canvas redraw and `Viewport3D.tsx:351` on every rebuild — the comment block in `toolpathVisibility.ts` already notes large projects carry tens of thousands of moves, which is ~6 ms of pure allocation per redraw at 50k.

Two cheap fixes that keep the single-source-of-truth property:
1. memoize the table for the last `slotScale` (it changes at most once per operation), or precompute it once in the `EngagementFeedQuantizer` constructor and pass it into `rungWidth`/`gapAbove`;
2. have the scan in `engagementFeedScale` keep the rung **index** it already found, so `rungWidth`/`gapAbove` don't re-scan for it three more times.

src/theme/palette.ts
```diff
@@ -478,24 +479,17 @@ export function pocketSlotFeedPercent(operation: Operation | null | undefined):
...
  *  scaled moves there. */
 export function feedColourScales(slotScale: number): readonly number[] {
-  const clamped = Math.min(1, Math.max(0, slotScale))
-  const bucketWidth = (1 - clamped) / (FEED_COLOUR_RUNG_COUNT - 1)
-  const rungs: number[] = []
-  for (let k = 0; k < FEED_COLOUR_RUNG_COUNT; k += 1) {
-    rungs[FEED_COLOUR_RUNG_COUNT - 1 - k] = clamped + k * bucketWidth
-  }
-  return rungs
+  return engagementFeedRungs(slotScale)
```
zabooma, line 487: **The three new top rungs collide once the legend rounds them to whole percent.**

`ToolpathVisibilityPanel.tsx:117,120` formats both the swatch tooltip and the label with `Math.round(scale * 100)`. Because rungs 1–3 now sit within 5/10/20 % of the drop instead of 20/40/60 %, they stop being label-distinct well before `pocketSlotFeedPercent` reaches its 99 maximum:

```
slot 85%:  100  99  99  97  94  91  88  85      <- two "99%"
slot 90%:  100 100  99  98  96  94  92  90      <- two "100%"
slot 95%:  100 100 100  99  98  97  96  95      <- three "100%"
slot 99%:  100 100 100 100 100  99  99  99
```

The pre-#591 ladder was label-distinct at every slot percent up to 98. And the two identically-labelled swatches are now harder to separate visually too: with 8 ramp steps instead of 6, adjacent-step ΔRGB dropped from 27→19 (dark) and 22→16 (light).

Not a blocker at the 60 % default, but a user picking a light reduction gets a legend with two entries that read the same and paint differently. Formatting to one decimal when the rounded labels collide would cover it.

src/engine/toolpaths/engagement.ts
```diff
@@ -654,6 +712,21 @@ export interface EngagementFeedFragment {
...
+ * one-fifth width); on the non-uniform ladder the narrow top rungs would
+ * otherwise leave an anti-chatter band several times finer than any engagement
+ * sample can resolve, letting the feed alternate across their boundaries along
+ * a ring whose engagement barely wanders. The floor keeps every rung's rise
+ * margin at least as wide — in absolute scale terms — as the uniform ladder's
+ * ever was (#591). A side effect is that the finest top rungs are drop-only:
+ * the feed falls onto them immediately but re-enters full feed through the
+ * deadband, like everywhere else.
+ */
+const MIN_HYSTERESIS_DROP_FRACTION = 0.05
```
zabooma, line 744: This block states two things the code doesn't do, and contradicts the class doc at line 752.

1. "keeps every rung's rise margin at least as wide — in absolute scale terms — as the uniform ladder's ever was". At slotScale 0.6 (drop 0.4, so the documented floor is 0.02) the actual margins are:

```
rung 0.98  width 0.020  gapAbove 0.020  margin 0.010   <- half the documented floor
rung 0.96  width 0.040  gapAbove 0.020  margin 0.010   <- half the documented floor
rung 0.92  width 0.080  gapAbove 0.040  margin 0.020
rung 0.84 … 0.60                        margin 0.020
```

Rung 0.98 is the *only* rung where the floor changes anything at all (elsewhere `0.25 × rungWidth` already equals `0.05 × drop`), and there the cap takes it back down to `gapAbove/2`.

2. "the finest top rungs are drop-only: the feed falls onto them immediately but re-enters full feed through the deadband". They are climbable — that is exactly what the cap at line 893 is for, and line 752 says so ("every rung stays climbable"). §5e's own test name ("the absolute rise-margin floor at a narrow top boundary") also describes the floor while the value it actually exercises is the cap, 0.01 rather than 0.02.

Worth rewriting as: the floor is the target, the cap is what actually binds at the top two rungs, and the realised minimum is `min(0.05 × drop, gapAbove/2)`.

src/engine/toolpaths/engagement.test.ts
```diff
@@ -440,6 +440,84 @@ console.log('Testing the bucket ladder includes 1.0 as its top rung...')
...
+  // rung to 0.92; now it lands on the second rung, ~1% under entitlement.
+  assert(
+    approx(scaleAtContinuous(0.991), 0.98, 1e-12),
+    `continuous 0.991 must charge the 0.98 rung, got ${scaleAtContinuous(0.991)}`,
+  )
+
+  // Mid-boundary samples charge the LOWER rung: quantization rounds down
+  // everywhere on the ladder. (The top boundary sits inside the estimator
+  // deadband by design, so the walk starts below the second rung.)
+  for (let i = 2; i < rungs.length; i += 1) {
```
zabooma: **This walk skips the one boundary #591 introduces, and the stated reason isn't true.**

`ENGAGEMENT_ESTIMATE_EPSILON` is `sqrt(2·16·2⁻⁵²)` ≈ **8.43e-8 rad**. The 1.0/0.98 boundary (continuous 0.99) sits **0.0485 rad** above nominal at this fixture — five orders of magnitude outside the deadband.

Verified against this branch: `engagementFeedScale(engagementAtContinuous(0.99), 1.2025, 0.6)` returns `0.98`, so the assertion passes with the loop starting at `i = 1`.

As written, the top boundary — the pair of rungs this PR exists to add, and the one place criterion 3's round-down direction is newly load-bearing — is the only boundary the walk doesn't check, and the comment tells the next reader not to add it.

```suggestion
  // Mid-boundary samples charge the LOWER rung: quantization rounds down
  // everywhere on the ladder, including the fine top boundary #591 adds.
  for (let i = 1; i < rungs.length; i += 1) {
```

src/theme/feedColourRamp.test.ts
```diff
@@ -56,85 +57,67 @@ const SLOT_PERCENTS = [40, 60, 75] as const
...
+// distinct colour step per rung (#591 made the set non-uniform).
 
 for (const slotPercent of SLOT_PERCENTS) {
   const slotScale = slotPercent / 100
   const scales = feedColourScales(slotScale)
+  const engineRungs = engagementFeedRungs(slotScale)
 
-  assert(scales.length === 6, `feedColourScales(${slotScale}) has ${scales.length} rungs, expected 6`)
+  assert(scales.length === engineRungs.length, `ramp must expose all ${engineRungs.length} engine rungs at ${slotPercent}%, got ${scales.length}`)
+  assert(scales.every((value, i) => value === engineRungs[i]), `ramp rungs must equal the engine ladder at ${slotPercent}%`)
```
zabooma, line 136: `feedColourScales` is now a one-line `return engagementFeedRungs(slotScale)` (palette.ts:487), so this asserts the function equals its own body. Every other assertion in the file derives its expectation from `feedColourScales`/`engagementFeedRungs` too — the per-rung distinct-step check, the between-rung rounding check, the top/bottom anchors.

With `BUCKET_CASES_40` and `BUCKET_CASES_75` deleted, nothing in the theme suite pins an actual rung value any more: swap `FEED_RUNG_DROP_FRACTIONS` for any other 8-entry table and this file still passes green.

The file's own header calls this out as the anti-pattern it's fixing — "Both were the code's own assumption written back as an oracle". Keeping one small concrete table (e.g. the 40 % cases, retargeted to the new ladder) restores an independent oracle without giving up the desync guarantee, which is really carried by the `scales === engineRungs` identity plus §5d's freeze test in the engine suite.

src/theme/palette.ts
```diff
@@ -504,17 +498,23 @@ export function feedColourStep(feedScale: number | undefined, slotScale: number)
...
     }
   }
-  return FEED_COLOUR_RUNG_COUNT - 1
+  return rungs.length - 1
 }
 
 interface Rgb { r: number; g: number; b: number }
 
+/** Number of ramp steps — one per engine rung (the count is slot-independent). */
+const FEED_COLOUR_STEP_COUNT = engagementFeedRungs(0).length
```
zabooma: `engagement.ts:106` already exports this exact value — `export const ENGAGEMENT_FEED_BUCKET_COUNT = FEED_RUNG_DROP_FRACTIONS.length`. Re-deriving it here (by allocating an 8-element array at module-init time) gives the module two names for one fact, in the file whose stated point is that the ramp can't desync from the engine.

```suggestion
const FEED_COLOUR_STEP_COUNT = ENGAGEMENT_FEED_BUCKET_COUNT
```

…with `ENGAGEMENT_FEED_BUCKET_COUNT` added to the existing import on line 19.

src/engine/toolpaths/engagement.ts
```diff
@@ -720,7 +798,43 @@ export class EngagementFeedQuantizer {
...
+  /**
+   * Width of the continuous-scale bucket a rung owns: the gap down to the
+   * next lower rung; the bottom rung reuses the final gap so its width is
+   * never zero.
+   */
+  private rungWidth(scale: number): number {
+    const rungs = engagementFeedRungs(this.slotScale)
+    const index = rungIndexOf(scale, this.slotScale)
+    if (index < 0 || index >= rungs.length) {
+      return (1 - this.slotScale) / (ENGAGEMENT_FEED_BUCKET_COUNT - 1)
```
zabooma: These fallbacks (here and the matching one in `gapAbove` at line 835) are unreachable: `rawScale` always comes out of `engagementFeedScale`, so it is always an emitted rung and `rungIndexOf` always finds it; `gapAbove`'s `index <= 0` case is additionally short-circuited by `rawScale >= 1 ? 0` at the call site.

They also no longer read as what they mean. `(1 - this.slotScale) / (ENGAGEMENT_FEED_BUCKET_COUNT - 1)` was the uniform ladder's `drop/5`; now that the count is 8 it evaluates to `drop/7`, a width belonging to no ladder that ever shipped.

If a future caller does pass a non-rung scale, silently returning that width makes the hysteresis margin arbitrary. Either drop the branches or throw — an unrecognised scale is a programming error, not a degenerate input.

src/engine/toolpaths/INDEX.md
```diff
@@ -26,7 +26,7 @@ Toolpath generators. Each file owns one strategy. `index.ts` re-exports everythi
...
+- `engagement.ts` — pure cutter-engagement estimator (issue #498): `SweptMaterialIndex` models each prior sweep as an exact tool-radius capsule in a centreline-supercover spatial index, and returns the leading-semicircle measure in uncut material in `[0, π]` through closed-form capsule-arc unions; `nominalEngagement` gives the analytic stepover wrap angle; `engagementFeedScale` maps engagement onto a fixed non-uniform rung ladder — fine near full feed, coarse near slot (issue #591), rounding down, conservative and `EngagementFeedQuantizer` applies hysteresis plus a minimum fragment length so `feedScale` changes stay controller- and arc-fitting-friendly; `EngagementTelemetryAccumulator` folds distance-weighted samples into per-operation max/p95/distance-above-nominal telemetry. Degenerate input resolves toward more engagement. Tests: `engagement.test.ts`, `engagementOracle.test.ts`, `engagementPocket.test.ts`
```
zabooma: The rewrite dropped the parentheses that used to isolate the round-down aside, so the two independent clauses now run together:

> …coarse near slot (issue #591), **rounding down, conservative and** `EngagementFeedQuantizer` applies hysteresis…

It parses as a list rather than as two statements. `(rounding down, conservative), and \`EngagementFeedQuantizer\` applies…` restores it. Worth fixing since CLAUDE.md points contributors and agents at INDEX.md first for this module.

src/engine/toolpaths/engagement.ts
zabooma: Fixed in 03b7ca0 — the cap now wraps the whole , and the floor scales by , so  is a true zero margin again and the climbability invariant holds for any fraction. Byte-identical at the default (your sweep case now rises off 0.84).

src/engine/toolpaths/engagement.ts
zabooma: Fixed in 03b7ca0 —  memoizes the table for the most recent slot scale, the quantizer captures its ladder once in the constructor, and / look up through that cached array. No allocation on either hot path.

src/theme/palette.ts
zabooma: Fixed in 03b7ca0 — new : whole percents unless rounding collides, then one decimal for every entry. Also turned out to be load-bearing: the  lane caught the same phenomenon (75 % fixture now emits the two fine top rungs), spec expectations updated.

src/engine/toolpaths/engagement.ts
zabooma: Fixed in 03b7ca0 — docblock rewritten: floor is the target, the cap binds on the top two rungs, realised minimum stated as such. §5e'\''s title/comment renamed from floor to cap.

src/engine/toolpaths/engagement.test.ts
zabooma: Fixed in 03b7ca0 — walk starts at , deadband rationale removed. Confirmed green against this branch before pushing.

src/theme/feedColourRamp.test.ts
zabooma: Fixed in 03b7ca0 — concrete 40 % oracle table restored (retargeted to the new ladder) alongside the derived assertions.

src/theme/palette.ts
zabooma: Fixed in 03b7ca0 — .

src/engine/toolpaths/engagement.ts
zabooma: Fixed in 03b7ca0 — both fallbacks replaced with  throws; an unrecognised scale is now loud.

src/engine/toolpaths/INDEX.md
zabooma: Fixed in 03b7ca0 — parentheses restored.

src/theme/palette.ts
zabooma: Verified on `03b7ca0` — `feedLegendStepLabels` fixes 85 %, 90 %, 95 %, 97 % and 98 %. One case still collides, at the top of the range:

```
slot 97%: 100% 99.8% 99.7% 99.4% 98.8% 98.2% 97.6% 97%   ok
slot 98%: 100% 99.9% 99.8% 99.6% 99.2% 98.8% 98.4% 98%   ok
slot 99%: 100% 100%  99.9% 99.8% 99.6% 99.4% 99.2% 99%   <- still duplicate
```

`(0.9995 * 100).toFixed(1)` is `"100.0"`, which the `.replace(/\.0$/, '')` turns back into `"100"` — same label as the full-feed rung. `feedLegendStepLabels([1, 0.9995])` returns `['100%', '100%']`. 99 is reachable: `CAMPanel.tsx:1740` clamps to 1–100 and `resolveSlotFeedScale` only disables at ≥ 100.

The single-shot escalation is the issue — one decimal isn't guaranteed to separate rungs at every slot feed. Escalating until distinct covers the whole range:

```suggestion
export function feedLegendStepLabels(scales: ReadonlyArray<number>): string[] {
  const distinct = new Set(scales).size
  for (let decimals = 0; decimals <= 3; decimals += 1) {
    const labels = scales.map((scale) => `${Number((scale * 100).toFixed(decimals))}%`)
    if (new Set(labels).size === distinct) return labels
  }
  return scales.map((scale) => `${Number((scale * 100).toFixed(3))}%`)
}
```

Keying on `new Set(scales).size` rather than `scales.length` also handles the case `unionFeedColourLegendSteps` documents — the same scale carried at two ramp steps by two operations. Today `feedLegendStepLabels([1, 0.9, 0.9])` drops everything to decimals for a pair that is *supposed* to read identically.

zabooma: Fixed in 430ba2a — decimals now escalate (0 → 3) until every distinct scale reads distinctly, since no single-shot fallback covers the whole 1–99 % range. Distinctness keys on distinct scales rather than entries, so the same scale carried at two ramp steps by two operations keeps its whole-percent label. Verified against the full slot-99 ladder head: all eight labels distinct (100 / 99.95 / 99.9 / 99.8 / 99.6 / 99.4 / 99.2 / 99).

Issue #620, opened 2026-08-24
Machining order for `surface_clean` and `rough_surface`


zabooma:
Parent: #614 — the `surface_clean` and `rough_surface` × machining order cells. Depends on the declaration sub-issue.

## The cells

`machiningOrder` is offered on `pocket` and the two edge routes (`operationFields.ts:370`, `report.ts:78`). Feature-first ordering is generic: `isFeatureFirst`/`perFeatureOperations` (`multiFeature.ts:34`, `:61`) split an operation per machining feature and merge the parts nearest-block-first. `pocket.ts:3889`, `vcarve.ts:101` and `vcarveMedial/index.ts:83` opt in; the surface generators do not.

Both kinds can hold more than one machining feature, so the split is reachable:
- `surface_clean` requires `machiningFeatures.length > 0`, each `add` or `model` with a closed profile (`operationDefaults.ts:145`).
- `rough_surface` requires at least one STL model among its machining features and closed regions (`operationDefaults.ts:189`) — `.some(...)`, not exactly one.

## The cell that does not apply

`finish_surface_cleanup` × machining order — **does not apply**, and the reason belongs in the declaration: its target validity admits exactly one STL model plus closed regions (`operationDefaults.ts:165`, `modelCount !== 1` rejects), and `perFeatureOperations` drops region features before splitting, so the split can never produce more than one part. No implementation, reason recorded.

## The hazard

`machiningOrder` is **stored in saved files**, not merely defaulted: both committed 3D fixtures carry `"machiningOrder": "feature_first"` on their `rough_surface` and `finish_surface_cleanup` operations, from `operationDefaults.ts:350`. A `surface_clean` or `rough_surface` that starts honouring it re-orders its emitted blocks — same output, different sequence, and a different program at the machine.

Same fork as the feed-reduction sub-issue: (b) neutralise on load for kinds that never offered the control, or (c) take the re-ordering with fixture evidence. Decide once for both.

## Shape

- Cells flip to `applies`; `operationFields.ts` and `report.ts` read the declaration.
- `generateSurfaceCleanToolpath` and `generateRoughSurfaceToolpath` gain the `isFeatureFirst` → `perFeatureOperations` → `mergePocketToolpathResults` wrapper `generatePocketToolpath` uses at `pocket.ts:3889`, including the shared engagement telemetry hand-off if the feed-reduction sub-issue has landed.

## Verification

- Byte-identity per kind, per the fork chosen.
[...]

Issue #619, opened 2026-08-24
Slot feed and engagement reduction for `rough_surface` and `finish_surface_cleanup`


zabooma:
Parent: #614 — the `rough_surface` and `finish_surface_cleanup` × slot feed + engagement cells. Depends on the declaration sub-issue.

One issue for both kinds because it is one gate, one hazard and one verification story: `resolveSlotFeedScale` (`pocket.ts:386`) returns `null` for anything that is not `pocket` or `surface_clean`, and `cutsSlots` (`operationFields.ts:116`) hides both rows for the same reason.

## The cells

- `finish_surface_cleanup` clears its floor with the same offset rings a pocket does (`finishSurfaceCleanup.ts:766` onward) and never reduces feed. #614: *"This is the gap #583 closed for `surface_clean` and left open here."*
- `rough_surface` clears each level with `cutOffsetRegionRecursive` (`roughSurface.ts:89`) and likewise never reduces feed — a full-width pass through standing stock runs at the full cut feed.

## The hazard, and the fork this issue has to settle

Both parameters are already persisted on operations that never offered the control. `operationDefaults.ts:330` sets `pocketSlotFeedPercent: 60` and `:332` sets `pocketFeedReduction: 'slots_only'` for **every** kind, and `normalizeOperation` (`normalize.ts:246`) spreads those defaults under the stored operation. Neither committed 3D fixture stores either field — checked in `model-in-pocket.camj` and `3d-imported-block-test3.camj` — so both load at 60% today, unused.

Flipping the gate therefore changes emitted feed rates on every saved project the moment it loads. Two ways to take it, and this is the owner's call:

**(b) Preserve saved output.** A load-time neutralisation writing `pocketSlotFeedPercent: 100` for kinds that never offered the control, on the `#481`/`#599` format-migration-plus-warning pattern, so existing files keep their program and newly created operations get the reduction.

**(c) Take the change.** Accept that saved 3D roughing and cleanup operations start cutting slots slower, called out with fixture evidence per #614's acceptance criteria.

Nothing else in the issue is blocked on this answer; the implementation is the same either way.

## Shape

- Cells flip to `applies` in `CLEARING_CONTROL_SUPPORT`; `cutsSlots` and `resolveSlotFeedScale` both read it, so panel and engine cannot diverge.
- `pass`/`finishFloor` half of `cutsSlots` stays as it is — that is a per-operation question, not a per-kind one.
[...]

Issue #614, opened 2026-08-24
Clearing controls are inconsistent across the level-based operations


zabooma:
## Problem

Features added to the 2.5D clearing operations have landed one operation at a time, and the set each kind exposes now disagrees for no stated reason. A user who learns a control on Pocket does not find it on Surface clean or Rough surface, and cannot tell whether it is missing because it does not apply or because nobody wired it.

This is the same shape as #609 — capability decided per call site rather than declared once — but about which controls a kind offers at all, rather than which patterns its generator implements.

## The stated principle

Every **level-based** area-clearing toolpath should offer the same clearing controls unless there is a recorded reason it cannot: slot feed and engagement feed reduction, tangential S-links, rounded corners, and the seeded clearing pattern wherever an area is cleared.

**`finish_surface` is excluded and is not in scope here.** It is the only generator that moves in XYZ — the cutter descends along the surface within a pass — so the 2.5D clearing controls do not transfer and must not be applied to it by analogy. Measured on the committed fixtures, share of cut moves whose start and end Z differ:

| kind | fixture | sloped (XYZ) cuts | distinct Z |
| --- | --- | ---: | ---: |
| `finish_surface` | `issue-401-cone-finish` | **100%** | 6419 |
| `finish_surface` | `3d-imported-block-test3` | 2% | 4 |
| `rough_surface` | `3d-imported-block-test3` | 0% | 19 |
| `rough_surface` | `model-in-pocket` | 0% | 5 |
| `finish_surface_cleanup` | `model-in-pocket` | 0% | 2 |
| `pocket` | `another-pocket-test` | 0% | 1 |

`rough_surface` and `finish_surface_cleanup` are model-aware, which is why they run at many Z levels — but every cut is flat. They clear area a level at a time exactly as `pocket` does, so they are in scope. Anything `finish_surface` needs is its own issue with its own reasoning.

**Reading the table above:** the distinction is whether the *clearing pass* descends, not whether the operation ever emits a sloped move. A level-based kind emits sloped cuts too once helix or ramp entries are enabled (#413) — the fixtures above happen to use plunge entries, which is why they read 0%. `ToolpathMove.source` is only populated under `debugToolpath`, so there is no dependable flag to subtract entries by; re-running this probe on a helix-entry pocket will read as sloped and does not mean it became a 3D operation.

## Measured today

[...]

PR #611, opened 2026-08-24
test: backfill surface.test.ts — the five cases required by #583

zabooma:
Closes #600.

Backfills `src/engine/toolpaths/surface.test.ts` with the five unit-test cases #583's acceptance criteria required, none of which shipped with #584 (only an engagement-telemetry assertion in `camOperationSmoke.test.ts` landed). Baseline decision recorded per the issue plan: current output is pinned as the regression golden — reconstructing the pre-`3f05497` baseline was judged not worth a throwaway checkout given the behavioral assertions below.

## The five cases

1. **No-seed byte identity** — 4×4 boss → ~8×8 tool-centre region cannot fit three seed circles; `seeded_offset` is byte-identical to plain `offset`, zero circular laps.
2. **Seeded rough + finish-floor emission** — 40×40 boss: circular-lap detector finds concentric seed laps sharing one anchor, inner-first within every Z pass, ring tree continuing after; seeded output differs from plain.
3. **Feed reduction across branches** — slots_only (60%) stamps exactly 0.6 on slotting cuts in rough-offset, rough-parallel and finish-floor; absent field ≡ 100% byte-identical.
4. **Engagement telemetry** — engagement mode attaches non-empty telemetry, quantized scales ≤ `ENGAGEMENT_FEED_BUCKET_COUNT`, at least one <1×.
5. **Corner settings** — see ⚠ below.

## ⚠ Case 5 diverges from #583's wording — needs your ruling

#583 wrote: *"roundLinkCorners/cleanWallCorners have no effect on surface-clean (it does not expose them)"*. **That premise does not match what #584 shipped**: `operationFields.ts` exposes both knobs for surface_clean (`appliesTo` explicitly includes it), and `surface.ts` honors them — tangent ring links under `roundLinkCorners` (surface.ts ~554), wall cleanup under `roundOutsideCorners && cleanWallCorners` (~563/~918).

The test therefore locks the **shipped contract** instead, with the divergence documented in a comment block: knobs surface_clean genuinely doesn't read (`cornerRelief`) stay inert, and `roundLinkCorners` demonstrably switches links. If you'd rather enforce #583's original intent, that's an engine gating change (drop `surface_clean` from those two gates) + UI exposure change — say the word and I'll file it separately.

## Verification

- All five cases green: `npx tsx src/engine/toolpaths/surface.test.ts`.
- Full `npm run build` green in the worktree.
- No e2e needed: pure test addition, no product code touched.

## Notes

[...]

PR #623, opened 2026-08-24
refactor(engine): declare the clearing-control set once, per kind (#616)

### LINKED ISSUE #616: Declare the clearing-control set once, per kind
Parent: #614 — steps 1, 2 and 4 of its Proposed shape. **Blocks every other sub-issue**, because each of them edits this table.

## What

One declaration of which clearing controls each level-based kind offers, and the recorded reason for every cell that does not. It encodes **today's** matrix exactly, so no toolpath and no emitted program changes.

## The answer lives in four places today

| where | decides |
| --- | --- |
| `src/components/cam/operationFields.ts:116` (`cutsSlots`) plus the inline kind lists at `:370`, `:456`, `:465`, `:470`, `:478` | which rows the panel renders |
| `src/engine/toolpaths/pocket.ts:386` (`resolveSlotFeedScale`) | whether the engine scales cut feed |
| `src/engine/operationBooklet/report.ts:78`, `:97`, `:105`, `:346` | what the operation booklet prints |
| `src/engine/toolpaths/pocketPatterns.ts` (`usesTangentLinks`) | S-links — the precedent this follows |

**They already disagree.** `report.ts:346` gates the feed-reduction row to `kind === 'pocket'`, while the panel offers the control on `surface_clean` (`cutsSlots`, `operationFields.ts:116`) and the engine honours it there (`surface.ts:401`, `:849`). A `surface_clean` operation set to engagement mode gets the control, gets the behaviour, and prints no row at the machine.

## Shape

`src/engine/toolpaths/clearingControls.ts`, next to `pocketPatterns.ts` and built the same way:

```ts
export type ClearingControl =
  | 'slotFeed' | 'engagementMode' | 'roundOutsideCorners'
  | 'cleanWallCorners' | 'cornerRelief' | 'machiningOrder' | 'tangentLinks'

export type ControlSupport =
  | { readonly applies: true }
  | { readonly applies: false; readonly reason: string }

export const CLEARING_CONTROL_SUPPORT:
  Readonly<Record<OperationKind, Readonly<Record<ClearingControl, ControlSupport>>>>
```

- `Record<OperationKind, …>` — a new operation kind does not compile until it classifies every control.
- `Record<ClearingControl, …>` on each row — a new control does not compile until every kind classifies it.
- Pattern-dependent cells (`tangentLinks`, `cleanWallCorners`) declare only whether the kind participates **at all**. Which patterns stays with `usesTangentLinks`, called after this gate: the table says "this kind links", the pattern function says "on these patterns". Two questions, one owner each.
- The `reason` string is the deliverable for a `no` cell — that is AC 1 of #614, recorded in the repo rather than in the issue.
- `finish_surface` is not a clearing kind. It declares `applies: false` on every row with the out-of-scope reason from #614, so the record accounts for every kind without implying the controls transfer to it.

## Consumers rewired

`operationFields.ts`, `pocket.ts:resolveSlotFeedScale` and `report.ts` all read the table. None of them keeps an independent kind list for these controls afterwards.

## The one deliberate change

The booklet starts printing the feed-reduction row for `surface_clean`, matching the panel and the engine. Booklet text only — no toolpath, no G-code.

## Verification

- Byte-identity sweep across the committed fixtures: no emitted move changes, since every cell is encoded as it behaves today.
- Structural test (AC 3 of #614): for a representative operation per clearing kind, the set of `OPERATION_FIELDS` whose `appliesTo` returns true, restricted to `ClearingControl`, equals the declared applies-set. A control wired to one kind and forgotten on another fails it.
- Booklet expectations updated for the `surface_clean` row.
- `npm run build`.

## Acceptance criteria

- One declaration; `operationFields.ts`, `pocket.ts` and `report.ts` carry no independent kind list for these controls.
- Every cell of #614's matrix resolves to *applies*, or to *does not apply, because …*, in the file.
- The structural test fails when a kind's exposed set diverges from the declaration.
- No change to emitted G-code for any fixture.

zabooma:
Closes #616

Parent: #614 (steps 1, 2, 4). Encodes today's clearing-control matrix exactly, so no toolpath changes.

## What

- New `src/engine/toolpaths/clearingControls.ts`, shaped like `OPERATION_PATTERN_SUPPORT` (#609): `CLEARING_CONTROL_SUPPORT` has one entry per `OperationKind` — either the six control cells (`slotFeed`, `engagementMode`, `roundOutsideCorners`, `cleanWallCorners`, `cornerRelief`, `machiningOrder`) or the one-line reason the kind clears nothing (`finish_surface` carries #614's XYZ reasoning). Both Record exhaustiveness obligations stay compile-time: a new kind or a new control fails to build until classified.
- Settled does-not-apply reasons are recorded verbatim from the owner's comment; cells owned by sibling issues (#619 slot/engagement, #620 machining order) say so instead of pre-deciding them.
- Consumers rewired onto the table, per the agreed approach: `operationFields.ts` (the kind halves of five field predicates), `pocket.ts` `resolveSlotFeedScale`, the `surface.ts` tangent-link gate pointed at `usesTangentLinks` and its wall-cleanup gate at the table, and `operationBooklet/report.ts`.
- Boundaries kept deliberate: S-links stay with `usesTangentLinks` in `pocketPatterns.ts`; the edge routes keep an inline half at each consumer (`clearingControlApplies(...) || edge route`), since they are not clearing kinds and sit outside #614's scope.
- Structural test in `operationFields.test.ts` rides `everyOperationShape()`; `seeded_offset` added to its `ALL_PATTERNS` (seeded shapes were never enumerated before).

## The one deliberate change

The booklet starts printing `surface_clean`'s feed-reduction rows — the panel already offered the controls and the engine already honoured them. Booklet text only; no toolpath changes.

Scope note: beyond the four cited report sites (:78, :97, :105, :346), two more independent kind lists for these same controls lived in that file — the pocket-only slot-feed row at :342 and the pocket-only clean-wall row at :326. The acceptance criterion ("no independent kind list for these controls") pulled both in, so those rows also start printing for `surface_clean`. No committed fixture contains a `surface_clean` operation, so no committed artifact changes.

## Verification

- **G-code byte identity:** sha256 of the posted program for all 12 committed fixtures x 18 operations — identical between `main` and this branch.
- **Structural test:** exposed clearing set equals the declaration on every clearing kind's canonical shape (all secondary gates satisfied), and across every enumerated shape no kind exposes an undeclared control except the deliberate edge-route half.
- **Proof split per the approach comment:** the fixture sweep cannot see `surface_clean` (no committed fixture has one), so that kind's generator claims rest on `surface.test.ts` (#611 backfill); booklet claims rest on `operationBooklet.test.ts`.
- `npm run build` green (docs check, lint, tsc, 208 test files, vite).
- No e2e added deliberately: rendered DOM, menus and dialogs are untouched — panel predicates resolve identically for every kind; unit-level structural tests own this change.

## Review fixes (post-review commits)

- `surface.ts` finish-pass gates rewired: `floorTangentLink` now calls `usesTangentLinks`, `floorWallCleanup` reads the table — the two independent kind lists that had survived 350 lines below their rough-pass siblings.
- The panel's feed rows each read their own cell now (`feedReductionApplies(operation, control)`), so `engagementMode` is declared **and** read; #619 flipping one cell cannot drag the other.
- `rough_surface` / `finish_surface_cleanup` x `cleanWallCorners` reasons reworded from analogy-settled to undecided: #614 deliberately excluded that control from its matrix, so no owner call exists yet.
- Re-verified after the fixes: all 12 fixtures x 18 operations still post byte-identical G-code; affected test files and full `npm run build` green.


zabooma:
Both pre-merge items are fixed (73cf908, f4a4aba):

1. **surface.ts :918/:925** — `floorTangentLink` now calls `usesTangentLinks` and `floorWallCleanup` reads the table. For every reachable input the answers are unchanged (`isParallelPocket` is effective-raster, which for these kinds is exactly stored `parallel`). Engine-wide grep confirms no inline kind list for the six controls survives outside `pocketPatterns.ts` and the deliberate edge-route halves.
2. **engagementMode** — `cutsSlots` split into `feedReductionApplies(operation, control)`; both panel rows read their own cell, matching the booklet. Declining `pocket.engagementMode` now fails the panel-side structural assertion too.

On the minor notes: took the second — the clean-wall cells are recorded as **undecided** (#614 deliberately left that control out of its matrix), consistent with how the #619/#620 cells are marked. Left `report.ts:340` alone: it is a #609 leftover about pattern rows, outside the six controls this issue declares — happy to file it as its own one-liner if you want it closed.

Re-verified: byte-identity sweep 18/18 identical vs main, affected suites green, full build green.

Issue #622, opened 2026-08-24
Feed reduction and corner controls: audit where every switch actually lands


zabooma:
> **Scope widened 2026-08-25 (owner call).** This started as one defect — the pocket finish pass never feed-classifies its wall cuts. Investigating it turned up four more in the same feature, so the issue now covers the whole audit: where every corner and feed switch actually lands, for every operation kind, including the non-clearing ones.

## Original report

Pocket finish does not recognize slot / high engagement cuts for wall cleaning.

Use purecutcnc example file for testing.

<img width="787" height="549" alt="Image" src="https://github.com/user-attachments/assets/fb986f41-b83a-4c66-8c4a-c129dd9a6416" />

Label mapping, because it matters throughout: **"Round corners" is `roundOutsideCorners`; "Round wall corners" is `cleanWallCorners`.**

Fixtures: `public/examples/purecutcnc.camj` (D1) and `work/purecutcnc-test-feed-reduction.camj` (D3–D6).

---

## D1 — The pocket finish pass never feed-classifies its wall block

`generateFinishBandMoves` (`pocket.ts`) emits a finish band as two blocks at the same Z: the **floor** block (`for (const z of floorStepLevels)`) and then the **wall** block (`for (const z of wallStepLevels)`) — `wallOuterContours`, the acute island-corner cleanup segments, and `wallFinalContours`/`wallContours`.

`applyLevelFeed` — the one place `pocketSlotFeedPercent` / `pocketFeedReduction` become a `feedScale` — is called **only at the end of the floor block**. The wall block is emitted after it and never classified, so every wall-finish and wall-cleanup cut leaves the engine at 100 % feed however engaged it is. Both reduction modes; in `engagement` mode the wall block is also missing from the operation's telemetry.

Measured, `public/examples/purecutcnc.camj` op0017 (1/8″ tool, 70 % slot feed, `slots_only`):

| variant | cut moves | cut length | scaled moves | scaled length |
| --- | --- | --- | --- | --- |
| shipped (floor + walls) | 1472 | 64.203 in | 94 | 5.449 in |
| floor only (`finishWalls: false`) | 1316 | 48.775 in | 94 | 5.449 in |
| **walls only (`finishFloor: false`)** | **156** | **15.428 in** | **0** | **0 in** |

[...]

PR #639, opened 2026-08-26
fix(worktrees): contributor-agnostic base & never-inside-checkout guard (#637)

### LINKED ISSUE #637: Worktrees: contributor-agnostic base, never created inside a checkout
## Summary

`dispatch-task.sh`, `finish-task.sh` and `worker-status.sh` derive the worktree
base from **the script's own location** rather than from the repository. Run from
the primary checkout that happens to be correct. Run from any worktree — which is
normal for an integration-manager session — and the base resolves *inside* that
worktree, so the next delegated slice gets a worktree nested in a worktree.

It recurses: dispatch from the nested one and you get another level.

## The derivation

All three scripts share these two lines:

```bash
readonly REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
WORKTREE_BASE="${PURECUT_WORKTREE_BASE:-$(dirname "$REPO_ROOT")/worktrees/$(basename "$REPO_ROOT")}"
```

`SCRIPT_DIR` is wherever the invoked copy of the script lives. Every worktree has
its own `scripts/`, so `REPO_ROOT` becomes the worktree, not the repo.

| run from | `REPO_ROOT` resolves to | `WORKTREE_BASE` becomes |
| --- | --- | --- |
| `~/Projects/purecutcnc` (primary) | `~/Projects/purecutcnc` | `~/Projects/worktrees/purecutcnc` ✅ |
| `~/Projects/worktrees/purecutcnc/issue-554-dd50d5` | that worktree | `~/Projects/worktrees/purecutcnc/worktrees/issue-554-dd50d5` ❌ |

`PURECUT_WORKTREE_BASE` is unset in the normal environment, so the fallback is
what runs.

## Evidence — this already happened

Issue #622's slice 5 was dispatched from the worktree holding the integration
branch. Its worker ran here:

```
~/Projects/worktrees/purecutcnc/worktrees/issue-554-dd50d5/control-effect-contract
```

`git worktree list` showed it at that path for the whole slice. `finish-task.sh`
removed it correctly afterwards, so nothing was lost — the breakage is location,
not data.

## Reproduce

```bash
cd "$(git rev-parse --show-toplevel)"          # from any non-primary worktree
bash -c 'SCRIPT_DIR="$PWD/scripts"; REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)";
         echo "$(dirname "$REPO_ROOT")/worktrees/$(basename "$REPO_ROOT")"'
```

Prints a path under the current worktree instead of the shared worktree base.

## Proposed fix

Derive the repository root from git rather than the script path. `--git-common-dir`
points at the primary checkout's `.git` from every worktree:

```bash
readonly REPO_ROOT="$(dirname "$(git rev-parse --path-format=absolute --git-common-dir)")"
```

Verified both directions:

```
from a worktree:      script-derived -> .../worktrees/purecutcnc/issue-554-dd50d5
                      git-derived    -> ~/Projects/purecutcnc
from primary:         both           -> ~/Projects/purecutcnc
```

`SCRIPT_DIR` is still wanted for locating the agent leaf scripts
(`run-dsh-agent.sh`, `run-claude-deepseek-agent.sh`) — that use is correct and
should stay as it is. Only the `REPO_ROOT`/`WORKTREE_BASE` derivation changes.

## Affected files

- `scripts/dispatch-task.sh:12` and `:17`
- `scripts/finish-task.sh:12` and `:14`
- `scripts/worker-status.sh:13` and `:14`

`scripts/test-dispatch-task.sh` already sets `PURECUT_WORKTREE_BASE` explicitly,
so it is unaffected and can gain a regression case that invokes a script through
a worktree copy.

## Two related nits, same area

**`worktrees/` is not gitignored.** `git check-ignore worktrees` reports nothing,
so while a nested worktree exists it shows as untracked in the parent's
`git status`. Cheap to add regardless of whether the nesting is fixed.

**Worktree directories are named after the session, not the branch.** Names are
minted once from whatever the session was first about and then the branch inside
gets swapped, so the directory name goes stale. Currently 3 of 6 under
`Projects/worktrees/purecutcnc/` describe something other than their contents:

| directory | branch actually checked out |
| --- | --- |
| `issue-554-dd50d5` | `fix/issue-622-feed-and-corner-controls` |
| `issue-614-subtasks-f91880` | `feat/issue-633-wall-cleanup` |
| `0-4-0-release-draft-3e0f7d` | `brave-bartik-c315df` (plain `main`) |

There are also two naming schemes in use — twelve worktrees at
`Projects/purecutcnc-wt-NNN` (issue-numbered, self-describing) and six at
`Projects/worktrees/purecutcnc/<session>-<hash>`. Worth settling on one and
naming by branch, but that is a convention decision, not a defect.

---

## Hard rule to enforce (owner, 2026-08-26)

**A worktree is never created inside a checkout of this repository** — not inside
the primary project folder, not inside another worktree, and **not inside a
dot-directory** of either (`.claude/worktrees/`, `.worktrees/`, `.tmp/`, …). A
leading dot does not make it acceptable; hidden is not outside.

All worktrees live under an external base, outside every checkout:

```
<worktrees-base>/purecutcnc/<branch-name>
```

There is no exception for a scratch tree, a delegation slice, or one that gets
removed afterwards.

**Why it has to be a guard and not just a convention:** a worktree inside a
checkout pollutes the parent's `git status` (nothing ignores `worktrees/`),
breaks every script that derives paths by walking up from `cwd`, makes cleanup
ambiguous about which tree it is acting on, and recurses — a worktree nested in a
worktree can nest again.

### The resulting convention must be contributor-agnostic

The fix is not "point it at the right folder on one machine" — the outcome should
be a worktree scheme any contributor gets for free on a fresh clone, wherever
they keep the repo, with **no per-machine configuration**.

Usefully, the existing fallback shape already is that:

```
<parent-of-repo>/worktrees/<repo-name>/<branch-name>
```

A contributor with the repo at `/src/purecutcnc` gets `/src/worktrees/purecutcnc/`;
one at `/work/code/purecutcnc` gets `/work/code/worktrees/purecutcnc/`. It is
always a sibling of the checkout, so it is outside it by construction, and it
needs nothing set. **Only the anchor is wrong** — `REPO_ROOT` is taken from the
invoked script's location instead of the repository, so from a worktree the
"parent" is the worktree itself. Fix the anchor and the default becomes correct
everywhere, for everyone, with no new config surface.

`PURECUT_WORKTREE_BASE` stays as an escape hatch for anyone who wants a different
layout, but it must be an absolute path and must be validated by the same guard —
an explicitly-set base pointing inside a checkout is still wrong.

No documentation or default may name a specific user's home directory.

### Acceptance criteria

- [ ] `REPO_ROOT` derived from `git rev-parse --path-format=absolute --git-common-dir`
      in all three scripts, so the base is identical from any checkout.
- [ ] `dispatch-task.sh` **refuses to run** if the resolved worktree path would
      land inside any path reported by `git worktree list --porcelain`, with an
      error naming the offending parent. Compare after symlink resolution
      (`pwd -P`), not textually.
- [ ] The same guard rejects a `PURECUT_WORKTREE_BASE` that points inside a
      checkout, so an explicitly-set-but-wrong base cannot bypass it.
- [ ] `worktrees/` added to `.gitignore` as belt-and-braces.
- [ ] Regression case in `scripts/test-dispatch-task.sh` that invokes a script
      through a worktree copy and asserts the base resolves outside it, plus one
      asserting the guard rejects an inside-checkout base.
- [ ] Default base needs **no configuration** and is correct on a fresh clone at
      an arbitrary path — verified by cloning to a throwaway directory and
      dispatching from both the clone and a worktree of it.
- [ ] Rule and convention written into `AGENTS.md` so it binds every agent
      session, not just the scripts — stated in the generic
      `<parent-of-repo>/worktrees/<repo-name>/` form, with no absolute example
      paths and no contributor's home directory.

### Secondary cleanup, same area

The 8 one-off probes under `scripts/` hardcode an absolute path into a
contributor's home directory (`issue-545-junction-census.ts`, `list-ops.ts`,
`profile-stl.ts`, `test-clipper-fallback.ts`, `test-rough-pipeline.ts`,
`test-stl-import.ts`, `test-stl-silhouette.ts`, `trace-o-emission.ts`). None is
wired into `package.json`, CI, or the test runner (`run-tests.ts` globs
`src/**/*.test.ts` only), and most target files under the gitignored `work/`, so
they cannot run for anyone else regardless. They should take the target as
`process.argv[2]` with a repo-relative default — cosmetic, but it reads badly in
a public repo.

Audited at the same time and confirmed clean of absolute paths: `package.json`,
`vite.config.ts`, `playwright.config.ts`, `tsconfig.json`,
`src-tauri/tauri.conf.json`, tracked `.claude/`, and `.github/`. The one absolute
literal in a shell script is a deliberate fake in `scripts/test-dsh-agent.sh`
asserting the launcher unsets a credential variable.

zabooma:
## Summary
Resolves #637. The worktree base is derived from the repository (via
`git rev-parse --path-format=absolute --git-common-dir`) rather than the
invoked copy's location, so dispatching from any checkout resolves to
`<parent-of-repo>/worktrees/<repo>/<slug>` — correct on a fresh clone with
zero configuration. `dispatch-task.sh` now refuses to create (or accept via
`PURECUT_WORKTREE_BASE`) a worktree inside any registered checkout, comparing
symlink-resolved paths and naming the offending parent.

## Notes for review
- **finish-task.sh semantics change:** with `REPO_ROOT` now anchored to the
  repository, `finish-task.sh` merges into the primary/anchor checkout instead
  of whichever worktree the manager invoked it from. This is intentional per the
  issue ("base identical from any checkout"); the integration branch is shared
  by all worktrees via refs anyway. Verified via a real dispatch->finish
  round-trip on a throwaway clone.
- **Full lane:** this touches `AGENTS.md` (protected, see `check-fast-lane.sh`)
  and exceeds the 3-file/25-line budget, so it cannot claim fast lane. No fast
  lane label is applied.
## Verification
- `bash scripts/test-dispatch-task.sh` -> passed (incl. two new #637 cases:
  default base resolves outside the invoking worktree, and guard rejects an
  inside-checkout `PURECUT_WORKTREE_BASE`).
- Fresh clone to a throwaway directory: dispatch from both the clone and a
  worktree-of-the-clone created the task worktree at `<clone-parent>/worktrees/purecutcnc/<slug>`,
  created no nesting, and the guard rejected a dot-dir inside-checkout base.
- End-to-end dispatch->finish round-trip merged cleanly and removed the worktree.

Closes #637

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,10 @@ coverage
 work/
 temp/
 
+# Worktrees must live outside every checkout (issue #637); ignored as
+# belt-and-braces so a stray nested one still cannot pollute git status.
+worktrees/
+
 # Env files (.env.agent contains the local DeepSeek worker token)
 .env
 .env.*
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -292,6 +292,16 @@ Two independent choices: **where** the work happens, and **who** does it.
 
 **Where — default to an isolated worktree.** New work starts in its own `git worktree` on its own branch unless the user says otherwise. This includes fast-lane changes: the reason is parallelism, not diff size. A worktree lets several branches be in flight at once without one agent's edits, build artifacts, or checkout state disturbing another's, and a one-line fix benefits from that as much as a large change does. Create it outside the repository directory, and do not commit the path.
 
+**Hard rule — a worktree is never created inside a checkout of this repository** (issue #637). Not inside the primary project folder, not inside another worktree, and not inside a dot-directory of either (`.claude/worktrees/`, `.tmp/`, …) — a leading dot does not put it outside. There is no exception for a scratch tree, a delegation slice, or one removed afterwards: a nested worktree pollutes the parent's `git status`, breaks scripts that derive paths from their location, recurses, and makes cleanup ambiguous.
+
+All task worktrees live under an external base, outside every checkout:
+
+```
+<parent-of-repo>/worktrees/<repo-name>/<branch-or-slug>
+```
+
+This is contributor-agnostic by construction: the base is always a sibling of whatever directory holds the checkout, on any machine, with nothing configured. `PURECUT_WORKTREE_BASE` is an escape hatch for a different layout; if set, it must be an absolute path, and `scripts/dispatch-task.sh` validates both the default and the override against every registered checkout (symlink-resolved), refusing to create a worktree inside one.
+
 A fresh worktree has no `node_modules`. Symlinking the primary checkout's copy avoids a full install:
 
 ```bash
--- a/scripts/dispatch-task.sh
+++ b/scripts/dispatch-task.sh
@@ -9,7 +9,16 @@
 set -euo pipefail
 
 readonly SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-readonly REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+# REPO_ROOT must be the repository, never the checkout the invoked copy lives
+# in (issue #637): every worktree has its own scripts/, so deriving it from
+# SCRIPT_DIR nests new worktrees inside the current one. The common dir points
+# at the primary checkout's .git from every worktree.
+if repo_common_dir="$(git -C "$SCRIPT_DIR" rev-parse --path-format=absolute --git-common-dir 2>/dev/null)"; then
+  readonly REPO_ROOT="$(dirname "$repo_common_dir")"
+else
+  printf 'dispatch-task: not inside a git repository: %s\n' "$SCRIPT_DIR" >&2
+  exit 1
+fi
 readonly CLAUDE_DEEPSEEK_LEAF="$SCRIPT_DIR/run-claude-deepseek-agent.sh"
 readonly DSH_LEAF="$SCRIPT_DIR/run-dsh-agent.sh"
 readonly DEFAULT_BASE="feat/core-arch-simplification"
@@ -67,6 +76,37 @@ EOF
 }
 
 fail() { printf 'dispatch-task: %s\n' "$*" >&2; exit 1; }
+
+# Absolute, symlink-resolved path (like realpath). The leaf may not exist yet,
+# so resolve the deepest existing ancestor and append the remainder textually.
+abs_path() {
+  local target="$1" resolved=""
+  if resolved="$(cd "$target" 2>/dev/null && pwd -P)"; then
+    printf '%s\n' "$resolved"
+    return 0
+  fi
+  local parent base
+  parent="$(dirname "$target")"
+  base="$(basename "$target")"
+  if ! resolved="$(cd "$parent" 2>/dev/null && pwd -P)"; then
+    resolved="$parent"
+  fi
+  printf '%s/%s\n' "$resolved" "$base"
+}
+
+# Hard rule (issue #637): a worktree is never created inside a checkout of this
+# repository — not the primary, not another worktree, not a dot-directory of
+# either. Compares symlink-resolved paths against every checkout git knows.
+assert_outside_checkouts() {
+  local candidate="$(abs_path "$1")" registered="" other=""
+  while IFS= read -r registered; do
+    [[ -n "$registered" ]] || continue
+    other="$(abs_path "$registered")"
+    if [[ "$candidate" == "$other" || "$candidate" == "$other"/* ]]; then
+      fail "refusing to create a worktree inside a checkout: $candidate would land inside checkout $registered; worktrees must live under an external base like <parent-of-repo>/worktrees/<repo-name>/"
+    fi
+  done < <(git -C "$REPO_ROOT" worktree list --porcelain | sed -n 's/^worktree //p')
+}
 
 mode="implement"
 provider="claude-deepseek"
@@ -200,6 +240,10 @@ fi
 
 branch="feat/issue-${issue}-${slug}"
 worktree_dir="$WORKTREE_BASE/$slug"
+
+# An explicitly-set base pointing inside a checkout must not bypass the guard.
+assert_outside_checkouts "$WORKTREE_BASE"
+assert_outside_checkouts "$worktree_dir"
 
 git -C "$REPO_ROOT" rev-parse --is-inside-work-tree >/dev/null 2>&1 \
   || fail "primary checkout is not a git repo: $REPO_ROOT"
--- a/scripts/dispatch-task.sh
+++ b/scripts/dispatch-task.sh
@@ -24,6 +24,10 @@ readonly DSH_LEAF="$SCRIPT_DIR/run-dsh-agent.sh"
 readonly DEFAULT_BASE="feat/core-arch-simplification"
 readonly MAX_DIRECT_PROMPT_BYTES=4096
 WORKTREE_BASE="${PURECUT_WORKTREE_BASE:-$(dirname "$REPO_ROOT")/worktrees/$(basename "$REPO_ROOT")}"
+if [[ -n "${PURECUT_WORKTREE_BASE:-}" && "$PURECUT_WORKTREE_BASE" != /* ]]; then
+  printf 'dispatch-task: PURECUT_WORKTREE_BASE must be an absolute path: %s\n' "$PURECUT_WORKTREE_BASE" >&2
+  exit 1
+fi
 
 usage() {
   cat <<'EOF'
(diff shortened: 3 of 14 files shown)


zabooma:
## Review

Reviewed the branch at `92716d0`, ran `scripts/test-dispatch-task.sh` (passes, including the two new #637 cases), and reproduced each finding below on throwaway git fixtures.

The `REPO_ROOT` re-anchoring is the right fix and the guard is a good addition. Five problems, three of them blocking.

---

### 1. `finish-task.sh` now hard-fails in the exact scenario #637 describes

`scripts/finish-task.sh:86` runs `git -C "$REPO_ROOT" checkout "$base"`. With `REPO_ROOT` re-anchored to the primary checkout, if the integration branch is checked out in the manager's own worktree — which the issue itself documents as normal ("#622's slice 5 was dispatched from the worktree holding the integration branch") — git refuses:

```
== checking out integ and merging feat/issue-1-myslug (--no-ff) ==
fatal: 'integ' is already used by worktree at '.../worktrees/repo/manager'
finish-task: could not checkout integ
EXIT: 1
```

The same fixture, with the `main` version of the script and an explicit `PURECUT_WORKTREE_BASE`, merges cleanly. So the last step of the delegation loop regresses from working to blocked. Nothing is lost (it fails before the merge), but the manager cannot land the slice.

The "Notes for review" section frames this as an intentional semantics change, but the round-trip that verified it evidently ran from the clone itself, and `test-dispatch-task.sh` does not exercise `finish-task.sh` at all.

**Suggested fix:** find the checkout that already has `$base` from `git worktree list --porcelain` and merge there, falling back to `REPO_ROOT` when no worktree holds it. `REPO_ROOT` is still the right anchor for *locating* the worktree base — it is the merge target that should not blindly be the primary.

### 2. A relative `PURECUT_WORKTREE_BASE` walks straight through the new guard

`abs_path` (`scripts/dispatch-task.sh:82`) gives up after one level: when neither the target nor its parent exists, it returns a **relative** string, which then never prefix-matches the absolute checkout paths in `assert_outside_checkouts`. Reproduced end-to-end from inside the primary checkout:

```
PURECUT_WORKTREE_BASE="tmpwt/slices" ./scripts/dispatch-task.sh \
  --provider dsh --issue 200 --task-slug relbase --base integ --skip-build
```

```
.../guard/proj/repo                       [main]
.../guard/proj/repo/tmpwt/slices/relbase  [feat/issue-200-relbase]   <- inside the checkout

$ git status --short
?? tmpwt/
```

That is precisely the hard rule this PR adds. `AGENTS.md` states the override "must be an absolute path", but nothing enforces it, and the new `.gitignore worktrees/` entry does not help because the directory is not named `worktrees`.

**Suggested fix:** `fail` on a `PURECUT_WORKTREE_BASE` that does not start with `/`, and make `abs_path` walk up to the deepest existing ancestor rather than stopping at the immediate parent.

### 3. `scripts/issue-545-junction-census.ts` is broken in its own documented invocation

`scripts/issue-545-junction-census.ts:885` takes `ROOT` from `process.argv[2]`, which collides with the positional flags the same file documents at line 35 (`Usage: npx tsx scripts/issue-545-junction-census.ts [tracked] [nomesh]`):

```
$ node repro.mjs tracked
{ argv: [ 'tracked' ], trackedOnly: true, noMesh: false, ROOT: 'tracked' }
find FAILED: spawnSync /bin/sh ENOENT
```

The `execSync(find…, { cwd: ROOT })` on the very next line throws at module load, so the script will not start. The usage comment was not updated either. Note that `tsc -b` covers only `src` and `vite.config.ts`, so `npm run build` gives this file no coverage.

**Suggested fix:** take the root from a named `--root DIR` flag, or use the first argv entry that is not `tracked`/`nomesh`.

The other six `.ts` scripts have no competing positional arguments — those changes are fine.

### 4. `finish-task.sh` silently repoints the user's primary checkout

`original_branch` (`scripts/finish-task.sh:83`) is only restored on merge failure. Reproduced: the primary checkout went `main` -> `integ` and stayed there after a successful merge. That was defensible when `REPO_ROOT` was the checkout you invoked from; now it mutates a checkout the manager is not in.

The dirty-check at `scripts/finish-task.sh:80` flipped the same way — it now gates on the primary checkout's unrelated WIP instead of the manager's.

**Suggested fix:** restore `original_branch` after a successful merge as well.

### 5. Re-dispatching an existing slug gives a self-contradictory error

The guard at `scripts/dispatch-task.sh:246` runs before the `worktree path already exists` check at line 254, so a still-registered worktree at the target path reports:

```
dispatch-task: refusing to create a worktree inside a checkout: <path> would land inside checkout <path>; ...
```

The path lands "inside" itself. On `main` this hit the clear `worktree path already exists` message.

**Suggested fix:** move the `[[ -e "$worktree_dir" ]]` check ahead of the guard, or give the `candidate == other` case its own message.

---

### Smaller notes, not blocking

- `.gitignore worktrees/` sits slightly at odds with the hard rule's intent: hiding a stray nested worktree from `git status` removes the signal that the rule was violated.
- `assert_outside_checkouts` reads from a process substitution, so a failing `git worktree list` makes the guard a silent no-op. Harmless in practice, since the `--is-inside-work-tree` check at `scripts/dispatch-task.sh:248` catches a bad `REPO_ROOT` immediately after — but the ordering means the guard runs against an unvalidated `REPO_ROOT`.

zabooma:
All 5 findings addressed in 0b50e3d:

**Blocking:**
1. `finish-task.sh` now uses `find_checkout_for_branch` to locate the worktree that already has `$base` checked out (from `git worktree list --porcelain`), falls back to REPO_ROOT when nobody does. Merge runs there, not in REPO_ROOT blindly. Verified: manager worktree holds the integration branch, dispatch+finish round-trip merges cleanly without the "already used by worktree" error.

2. `abs_path` walks up to the deepest existing ancestor (not just one level). Added explicit validation that `PURECUT_WORKTREE_BASE` must start with `/` — `fail` on relative paths. Verified: `PURECUT_WORKTREE_BASE="tmpwt/slices"` now gets "must be an absolute path" error instead of silently nesting inside the checkout.

3. `issue-545-junction-census.ts` now uses `--root DIR` named flag (argv parsing) instead of `argv[2]`, so `tracked`/`nomesh` positional flags don't collide. Usage comment updated.

**Non-blocking:**
4. `finish-task.sh` restores `original_branch` after a successful merge too (not just on failure). The checkout used for merging is now the one holding `$base` (the manager's worktree in the normal case), so the primary is never repointed.

5. `[[ -e "$worktree_dir" ]]` check moved ahead of the guard — re-dispatching an existing slug now gets "worktree path already exists" instead of the confusing "would land inside checkout itself".

All fresh-clone checks pass: dispatch from clone, dispatch from clone's worktree (shared sibling base, no nesting), guard rejects inside-checkout base, guard rejects relative base, re-dispatch existing slug gives clear error. Dispatch+finish round-trip with integration branch in manager worktree works end-to-end.

Issue #675, opened 2026-08-29
Move toolpath generation off the main thread (worker-backed CAM pipeline)


zabooma:
# Detailed design: worker-backed CAM generation

**Status: proposed; awaiting maintainer approval.** Design requested on 2026-09-05. This section supersedes the exploratory sketch below as the proposed implementation contract. No implementation is authorized by publication of this design.

**Implementation prerequisite (maintainer direction, 2026-09-05):** Start #675 only after **both #708 and #458 have landed on main**. Base the implementation and its pre-extraction parity baseline on main containing both changes.

Verified against `main` at `b061c230c007e6501291ed8682fbbaeaf22b9f57`. References below use symbol names; rebase and re-check these seams before implementation. The maintainer is already working on #708 and #458; this plan neither takes over nor duplicates those changes. Integrate their landed behavior into the parity baseline before the worker rollout.

## Outcome and scope

An expensive CAM operation runs in a dedicated worker while the application can process input and Stop. A completed operation produces the same full toolpath and G-code as the existing implementation for the same project, operation, and export settings. Interrupted, failed, or superseded computations never become current results.

The first release uses **one worker, one active operation, one shared generation service**. It moves the existing generation pipeline across a thread boundary without redesigning machining algorithms or the cache dependency model. It adapts every application caller of generation to an asynchronous contract.

Included: generator dispatch, existing tab processing, move optimization, clamp post-processing, demand scheduling, cache integration, cancellation, operation/stage status, export/booklet/debug/simulation input acquisition, failure handling, and an application-local opt-in rollout.

Excluded: a worker pool; parallelizing one operation; SharedArrayBuffer/Atomics; a new packed move format; mesh transfer ownership changes; simulation heightfield computation/rendering (#113); CSG, import, model export, and postprocessor threading; new machining strategies; #116 visibility semantics; 3D face picking or #242 redesign. Simulation's *acquisition of toolpaths* must change here, but which operations it simulates and how it replays them must remain unchanged.

The claim is narrower than “the app can no longer pause”: copying large inputs/results, postprocessing, simulation replay, and rendering can still occupy the main thread. Transport and result-install costs are measured separately before enabling the worker by default.

## Current implementation and the migration boundary

| Current symbol | Verified responsibility | Proposed handling |
| --- | --- | --- |
[...]

PR #677, opened 2026-08-29
perf(toolpaths): decimate mesh slice contours before Clipper, budget the rest (#674)

### LINKED ISSUE #673: Firefox hang on clicking 3d roughing operation
Image

Console shows Loading failed for the <script> with source “https://gc.zgo.at/count.js”. [app:1:1](https://purecutcnc.github.io/app/)
WebGL warning: drawElementsInstanced: Drawing to a destination rect smaller than the viewport rect. (This warning will only be given once)
Script terminated by timeout at:
t</e.Clipper.prototype.BuildIntersectList@https://purecutcnc.github.io/app/assets/clipper-Bry7t3-W.js:1:63726
t</e.Clipper.prototype.ProcessIntersections@https://purecutcnc.github.io/app/assets/clipper-Bry7t3-W.js:1:63116
t</e.Clipper.prototype.ExecuteInternal@https://purecutcnc.github.io/app/assets/clipper-Bry7t3-W.js:1:42638
t</e.Clipper.prototype.Execute@https://purecutcnc.github.io/app/assets/clipper-Bry7t3-W.js:1:41455
t</e.ClipperOffset.prototype.Execute@https://purecutcnc.github.io/app/assets/clipper-Bry7t3-W.js:1:91487
Vw@https://purecutcnc.github.io/app/assets/index-2cexR75w.js:32:304936
TT@https://purecutcnc.github.io/app/assets/index-2cexR75w.js:32:314453
_J/r<@https://purecutcnc.github.io/app/assets/index-2cexR75w.js:261:69538
_J@https://purecutcnc.github.io/app/assets/index-2cexR75w.js:261:69525
xJ@https://purecutcnc.github.io/app/assets/index-2cexR75w.js:261:70781
bJ@https://purecutcnc.github.io/app/assets/index-2cexR75w.js:261:70746
DZ/s</<@https://purecutcnc.github.io/app/assets/index-2cexR75w.js:261:152545
d@https://purecutcnc.github.io/app/assets/index-2cexR75w.js:261:151789
[clipper-Bry7t3-W.js:1:63726](https://purecutcnc.github.io/app/assets/clipper-Bry7t3-W.js)
Unknown property ‘-moz-osx-font-smoothing’.  Declaration dropped. [index-COvozoSn.css:1:75004](https://purecutcnc.github.io/app/assets/index-COvozoSn.css)
Security Error: Content at https://purecutcnc.github.io/app/ may not load or link to file:///.

I imported an STL.  Rotated, resized and moved the depthmap.  The clicked 3d roughing toolpath.  Firefox hung and the error in the pic appeared.

### LINKED ISSUE #674: 3D roughing hangs the browser on detailed relief meshes: undecimated slice contours make island offsetting quadratic
Fixes the browser hang reported in #673.

## The report

An imported STL relief (a Makera cameo — 4 x 3 in plaque, `Z top` ~ 0.112 in, fine lace/beading/hair detail) was rotated, resized and moved, then 3D surface roughing was clicked. Firefox showed *"This page is slowing down Firefox"* and killed the script.

## Where it hangs

Frames de-minified against the deployed bundle `index-2cexR75w.js`:

| minified frame | source |
| --- | --- |
| `Vw` | `offsetPaths` — `src/engine/toolpaths/pocket.ts:129` |
| `TT` | `buildInsetRegions` — `src/engine/toolpaths/pocket.ts:1252` |
| `_J` | `resolve3DSurfaceStepdown` — `src/engine/toolpaths/surfaceStepdown3d.ts:138` |
| `xJ` / `bJ` | `generateRoughSurfaceToolpathSingle` / `generateRoughSurfaceToolpath` |
| `DZ/s</<`, `d` | `startToolpathGenerationPipeline` -> `computeNext` — `src/app/useToolpathGeneration.ts:278` |

The reported column inside `TT` lands on the **second** `offsetPaths` call — `offsetPaths(islandPaths, delta * scale, islandJoinType)` (`pocket.ts:1267`), which expands the level's *islands* by the tool inset. Not the outer contour.

This is not a Firefox bug. Firefox is simply the browser with a slow-script watchdog; Chrome spins silently on the same work.

## Root cause

`sliceMeshAtZDetailed` emits one contour vertex per triangle-edge crossing, and **nothing decimates it anywhere downstream**. The only cleanup on this path is `cleanClipperPath(path, 1.0)` (`pocket.ts:114`), which at `DEFAULT_CLIPPER_SCALE = 10_000` drops vertices closer than 0.1 um — i.e. nothing for a real mesh.

Every island of every rough level then enters a single `ClipperOffset.Execute`, whose internal union is quadratic in total vertex count. `resolve3DSurfaceStepdown` repeats that for every Z level, while `protectedAbovePaths` unions grow alongside it.

`finish_surface_cleanup` shares the same resolver (`finishSurfaceCleanup.ts:643`) and has the same exposure.

## Measurement

Isolated harness against `clipper-lib` 6.4.2 (the shipped version), reproducing that exact call: `jtMiter`, `EndType.etClosedPolygon`, delta `= +1.5875 mm x 10_000`, `ArcTolerance` set as `pocket.ts:139` sets it, islands spread over a 101.6 x 76.2 mm plaque. Node v26.0.0, Intel i7-8850H. Single runs; two rows re-run showed ~15% spread.

**Island count, at a fixed 0.3 mm vertex spacing:**

| islands | vertices | offset ms |
| ---: | ---: | ---: |
| 400 | 10 000 | 504 |
| 800 | 20 000 | 2 081 |
| 1600 | 40 000 | 8 970 |
| 3200 | 80 000 | 42 430 |

**1600 islands, vertex spacing varied:**

| spacing (mm) | vertices | offset ms |
| ---: | ---: | ---: |
| 1.2 | 12 800 | 408 |
| 0.6 | 20 800 | 1 952 |
| 0.3 | 40 000 | 9 305 |
| 0.15 | 80 000 | 44 759 |

**Total vertices held at ~40 000, traded between island count and spacing:**

| islands | spacing (mm) | vertices | offset ms |
| ---: | ---: | ---: | ---: |
| 400 | 0.075 | 40 400 | 14 730 |
| 800 | 0.15 | 40 000 | 11 933 |
| 1600 | 0.3 | 40 000 | 10 891 |
| 3200 | 0.6 | 41 600 | 7 707 |

Reading: roughly **4.7x time per 2x vertices**, and at constant total vertices the island count barely matters. **The driver is total vertex count in a single `Execute`.** That is one level; the operation repeats it per Z.

Caveat: this is a synthetic island set, not the reporter's model. It establishes the hot call and the scaling law, not an absolute time for that file.

## Proposed plan

**S1 — Decimate mesh-slice contours before they reach Clipper.**

Add a simplification pass in `slicePolygonsToClipperPaths` (`surfaceStepdown3d.ts:100`), before `toClipperPath`. `rdpSimplifyOpenRun` already exists in `arcReconstruction.ts:1069` and can be reused or adapted for closed contours. Because the curve is quadratic, halving vertex count is worth ~4.7x.

**Safety requirement, and the reason this is not a free change:** these contours are the *keep-out* for the rough pass. Douglas-Peucker can cut a convex corner inward, which would shrink the protected region and let the cutter gouge the model. The proposed mechanism is to add the decimation tolerance `eps` to the island expansion delta, so the expanded keep-out provably still contains the true one — a scalar change with no extra Clipper work. Tolerance must stay well under `stockToLeaveRadial` and under the finish pass's own tolerance.

This changes emitted G-code for 3D roughing and 3D cleanup on mesh models. Protected path: unit tests required, and `e2e/*.spec.ts` needs a grep before claiming no e2e impact.

**S2 — Complexity budget with an actionable warning.**

S1 buys a large constant factor but does not change the complexity class, so a dense enough mesh still hangs. Add a per-operation budget in the same shape as the trochoidal one: count slice vertices, and past a threshold emit a new warning code (`surface3dMeshTooDense`, with an i18n key) and return no moves rather than eating the tab. Threshold derived from the curve above, sized so a level stays inside a bounded time.

## Acceptance criteria

- A synthetic dense-relief fixture that today takes tens of seconds in `resolve3DSurfaceStepdown` completes in a small fraction of that after S1, with the roughed area unchanged within the decimation tolerance.
- The decimated keep-out is proven to *contain* the undecimated one on that fixture — no point of the true model silhouette falls outside the expanded protected region.
- A mesh past the S2 threshold returns the new warning and zero moves, verified on a fixture that actually trips the budget (not one that merely passes it).
- The new warning code has an i18n key in every catalog.
- Any timing assertion follows AGENTS.md "Performance assertions" — a ratio against invariant work or against input size, never a millisecond constant.
- Tests verified by mutation: break the decimation and the containment check, confirm the tests go red.
- `npm run build` green.

## Out of scope

Moving generation off the main thread — that is #675. It makes the hang cancellable and the UI responsive, but does not make this work faster.

zabooma:
Closes #674.

Makes 3D roughing and 3D cleanup substantially cheaper on meshes whose slice
contours are dense, and bounds the tail so a mesh that cannot be handled refuses
instead of spinning.

**This does not close #673.** We have the reporter's STL now and measured it:
318 s CPU on `main`, 185 s with this branch — a real 1.7x, and still a frozen
tab. Its slice contours are ordinary and never approach the budget; what costs
the time is running the per-level pipeline 436 times, because the mesh is a
quantized depth map and every plateau becomes a critical floor level. That is
filed separately as #682. See the comment thread below for the numbers.

## S1 — decimate the slice contours

`sliceMeshAtZDetailed` emits one contour vertex per triangle-edge crossing and
nothing downstream thinned it. `slicePolygonsToClipperPaths`
(`surfaceStepdown3d.ts`) now runs a closed-ring Douglas-Peucker over each slice
contour before it reaches Clipper, then re-expands the result so the thinned
keep-out provably contains the true cross-section.

**The margin is the error decimation actually introduced, not the tolerance.**
`simplifyClosedRing` reports the greatest distance any dropped vertex sits from
the polyline that replaced it, and 0 when it dropped nothing. That distinction
is what keeps this change off every project it would not help: a rectangular or
otherwise coarse cross-section has no vertex RDP can drop, so it is charged
zero, skips the expansion (`offsetClipperPaths` no-ops under a 1e-9 delta), and
emits **byte-identical G-code**. Only a mesh dense enough to actually thin pays
anything, and it pays exactly what thinning cost it.

**Two deviations from the plan as filed, both agreed before implementing:**

1. The `+eps` sits at the slice, not at the `buildInsetRegions(baseRegion,
   initialInset)` call site the plan named. That site produces
   `level.insetRegions` only, and `level.clearablePaths` is consumed
   independently at `roughSurface.ts:196` as the safe-link domain — a keep-out
   that shrank there would let the link planner rule a travel move safe across
   standing stock. Expanding at the source means `protectedAtLevel`,
   `protectedAbovePaths`, `clearablePaths`, `baseRegions`, `insetRegions` and
   that safe-link domain all inherit the containment. `pocket.ts` is untouched,
   so 2.5D output cannot move.
2. The margin is measured rather than nominal. A fixed margin collided with
   three separate one-micron slack constants in this subsystem — it collapsed
   the outer-wall band entirely on flat-bottomed models
   (`surface3dFloorCollapsed`), silently dropped a cleanup wall pass through
   `WALL_PROBE_EPSILON`, and cost two tangent S-links on the pocket-block
   fixture. Measuring the error rather than assuming it removes that whole class
   of knock-on effect at the source, and all three existing test files pass with
   their original constants.

Two consumers still have to know the shift, and `Resolved3DSurfaceStepdown`
now carries it: the machining envelope (the shift would otherwise eat the
one-micron `OUTER_WALL_MARGIN` band whole) and the cleanup wall probe. Both are
no-ops at shift 0.

## S2 — slice vertex budget

Decimation buys a large constant factor but not a complexity class, so a dense
enough mesh would still hang. A level past `DEFAULT_SURFACE_3D_SLICE_VERTEX_BUDGET`
= 20,000 contour vertices now refuses the whole operation with a new
`surface3dMeshTooDense` warning and zero moves. Checked after decimation and
before the level's protection union — the earliest point at which the driver is
known and the last before any Clipper work at that level. Fail closed, like the
trochoidal budget: a partial 3D rough path is not safe to run. The threshold
comes from the issue's own table (20,000 vertices measured 2,081 ms in one
`Execute`; 40,000 measured 8,970 ms).

`finish_surface_cleanup` shares `resolve3DSurfaceStepdown` and inherits both.

## Tests

New `surfaceStepdown3dDecimation.test.ts`, on a tapered relief tube whose
cross-section is a radial modulation of a circle — each level's ring strictly
contains the ones above it, so the level's own slice is what has to contain the
model rather than the protection inherited from above.

- **Containment.** No part of the true, undecimated cross-section may fall
  inside `clearablePaths` — both the cut domain and the safe-link domain.
  Measured 0, 0 and 2.6e-7 mm2 across the three levels.
- **A coarse box is charged nothing**, which is what keeps existing programs
  byte-identical.
- **The budget refuses**, verified on a fixture that actually trips it, paired
  against the *same vertex count* smooth enough to thin, which resolves
  normally.
- **Cost ratio** against an undecimable reference, per AGENTS.md § Performance
  assertions.

**Verified by mutation, both halves:**

| mutation | result |
| --- | --- |
| decimation removed | 4 tests red; ratio 0.221–0.258 against a 0.05 limit |
| containment margin removed | 0.849 mm2 of the model leaks into the clearable region, against a 1e-3 limit |

The perf reference is the same mesh with a sawtooth ring, so RDP keeps every
ring vertex. It is not perfectly invariant and the test records why: any
triangulated quad wall leaves half its slice points collinear to within 1e-4 mm,
so the reference thins 2x whatever the ring shape. That makes the ratio
understate the regression — the conservative direction — rather than drift.
Measured separation is 18x with under 10% spread either side; the limit is the
geometric mid-point of the worst pair.

## Verification

- `npm run build` — exit 0, 217 test files passed, 0 skipped.
- e2e: no spec reads 3D roughing or cleanup output. `rough_surface` appears once
  (`camOperations.smoke.spec.ts:257`), asserting operation parameters in project
  state, and every e2e fixture sets `modelAssets: {}`, so the resolver refuses
  before it can slice. Ran the mesh- and 3D-adjacent lanes anyway —
  `camOperations`, `modelOrientation`, `gcodeExport` — 18 passed.
- `surface3dMeshTooDense` has a key in all five catalogs, diffed by parsing
  key/value pairs: same three placeholders in each.
- Not fast-lane eligible (protected path), so no label.


zabooma:
## Follow-up: the margin needed to be free, and wasn't

Measuring this branch against real models turned up that the containment margin
was applied as a `ClipperOffset` of its own over the whole slice contour set —
the same expensive operation decimation exists to make cheaper. Per level, on a
101.6 x 76.2 mm relief plaque:

| worst contour, vertices | thinning (RDP) | slice union | **margin offset** |
| ---: | ---: | ---: | ---: |
| 600 | 1 ms | 6 ms | **14 ms** |
| 2 870 | 3 ms | 9 ms | **10 ms** |
| 11 134 | 12 ms | 31 ms | **107 ms** |
| 37 923 | 13 ms | 228 ms | **810 ms** |

The thinning is free; the margin was the whole cost. It only paid for itself
once contours were big enough for the downstream island offset to dominate, so
below roughly 5,000 contour vertices it was pure tax — and a mesh the budget was
about to refuse paid it in full first.

Now folded into the erode/dilate sliver cleanup already running on
`clearablePaths`. Eroding by `cleanupEpsilon + margin` and dilating by
`cleanupEpsilon` is an opening of `C ⊖ margin`, and opening is anti-extensive,
so the result stays inside `C ⊖ margin` — same guarantee, zero extra Clipper
calls. The envelope shift doubles to `2 × margin`, because an erosion takes the
outer-wall band in from both sides at once rather than only from the model side.

Resolve CPU, four generated relief plaques plus a 300k-triangle scanned coin
scaled to reporter size:

| fixture | main | own offset | folded |
| --- | ---: | ---: | ---: |
| coarse relief, 6k tri | 166 ms | 207 ms | **168 ms** |
| photo-style relief, 138k tri | 362 ms | 543 ms | **267 ms** |
| fine detail, 56k tri | 1 372 ms | 1 198 ms | **1 078 ms** |
| over-budget, 104k tri | 594 ms | 2 243 ms | **498 ms** (refuses either way) |
| denarius @ 100 mm, 300k tri | 3 220 ms | 1 863 ms | **1 754 ms** |

Every case is now at or better than `main`; three of four were worse before.

Both mutations re-verified against the new placement: decimation removed → 4
tests red, ratio 0.199–0.207 against a 0.042 limit; the margin dropped from the
erode → 0.849 mm² of the model leaks into the clearable region, against 0, 0 and
8.15e-7 mm² measured with it in place.

Also worth recording, since it shapes how this gets tested: **triangle count is
not the driver, contour character is.** The 300k-triangle coin has cross-sections
of only 1,500–4,900 points and never approaches the 20,000 budget, while a
56k-triangle relief with fine unsimplifiable detail resolves four times slower
than a 138k-triangle smooth one.

zabooma:
## The 20,000 budget was wrong, and would have refused working models

Testing against generated relief plaques turned up that the threshold I took
from the harness in the issue does not transfer to real meshes.

`d-toodense` — the plaque the guard refused — **resolves on `main` in 3.0 s CPU /
2.0 s wall and produces a usable toolpath.** Slow, not broken. Shipping the
20,000 budget would have turned that into a refusal: a capability regression, not
a fix.

The harness measured one `ClipperOffset.Execute` over 1,600 synthetic islands
offset by 1.5875 mm (20,000 vertices → 2,081 ms). That arrangement
self-intersects far more per vertex than an actual relief contour, so its curve
is much steeper than the one real meshes follow. I should have validated the
number against a mesh before picking it.

Re-measured through the whole resolver, with decimation and the budget both
disabled:

| worst level, counted vertices | raw contour vertices | resolve on `main` | verdict |
| ---: | ---: | ---: | --- |
| 26 092 | 35 308 | 3.0 s CPU / 2.0 s wall | usable |
| 108 922 | 151 173 | 45.5 s CPU / 40.8 s wall | not usable |

**50,000** is about the geometric mid-point of those and interpolates to roughly
ten seconds — also about where a browser's slow-script watchdog starts
complaining. It refuses what is genuinely unusable and allows what merely takes a
moment.

The reasoning for having a guard at all is unchanged: decimation buys a constant
factor, not a complexity class, and at half a million contour vertices the
operation runs for minutes. But the freeze itself belongs to #675 — this only
bounds it.

The over-budget test fixture grows to 30,000 ring points to stay past the new
line, and a **third mutation** now covers the guard itself: disabling the budget
must stop that fixture refusing, which proves the refusal comes from the guard
rather than from something else in the resolver. All three mutations verified red:

| mutation | result |
| --- | --- |
| budget disabled | over-budget fixture stops refusing |
| decimation removed | 4 tests red, ratio 0.205 against a 0.042 limit |
| containment margin removed | 0.849 mm² of the model leaks into the clearable region |

zabooma:
## Measured against the reporter's actual STL — this is not what limits it

We have the file from #673 now. It does not hit the pathology this PR fixes.

291,460 triangles, 223 x 244 x 12.4 mm, 3.175 mm tool at a 2.07 mm stepdown:

| | main | this PR |
| --- | ---: | ---: |
| resolve | 318 s CPU / 262 s wall | 185 s CPU / 160 s wall |
| rough levels | 436 | 436 |
| worst slice contour | 15,136 vertices, thins 1.6x | same |

A real 1.7x, and still a frozen tab. Its slice contours are ordinary and well
inside the 50,000 budget; what costs the time is running the per-level pipeline
**436 times**, because the mesh is a quantized depth map and every plateau
becomes a critical floor level. Filed separately as #682 with the numbers.

**This PR therefore does not close #673**, and the description should not be read
as claiming it does. `Closes #674` stands — the decimation and the budget are
real and measured on contour-heavy meshes — but #673 stays open against #682.

Issue #679, opened 2026-08-30
2D toolpath rendering is still O(moves) per frame — decimate by screen-space length and cull to the viewport


zabooma:
Split out of #664, after [#678](https://github.com/PureCutCNC/purecutcnc/pull/678) took the redundant per-frame work out.

## Where it stands

#678 removed everything the 2D canvas was re-deriving each frame — the layer split and the arrow-placement pass. What is left is the drawing itself, and it is still proportional to the move count on every frame.

Measured on `src/engine/test-fixtures/trochoidal-249k.camj` (249,663 moves), operation selected:

| per pan frame | ms |
| --- | ---: |
| line strokes (~500k `moveTo`/`lineTo`) | ~18 |
| arrow fills (9,899 individual paths) | ~27 |
| **total** | **~46** |

That is 22 fps. Zoom is worse — ~135 ms — because it invalidates the placement cache by design.

The maintainer's assessment of #678 on the real project was that it makes no perceptible difference, which is fair: 7 fps and 22 fps are both choppy.

## Why shaving further will not work

Two things are already ruled out by measurement, so nobody should re-try them:

- **Batching the arrow fills into one path is worse, not better.** Measured 86.6 ms → **317 ms**. A single canvas path with ~9,899 filled subpaths costs far more to rasterise than 9,899 small ones.
- **Removing allocation from the draw loop does nothing.** Measured 86.63 ms → 86.65 ms.

The cost is irreducible while the renderer touches every move every frame.

## What to do instead

The 3D viewport shows the same 249,663 moves at 60 fps because it uploads geometry once and never looks at a move again. The 2D canvas needs the same property. Two steps, and they cover opposite ends of the zoom range:

1. **Decimate by screen-space length.** Merge consecutive moves while their combined on-screen length stays under a pixel. Invisible by construction — it only discards detail below what the display can resolve — and it self-disables as you zoom in, because moves get longer and the threshold stops merging. Helps most when zoomed to fit, which is where this fixture hurts.
2. **Cull to the viewport.** Zoomed in, decimation stops helping because N falls to 1, but most of the path is off-screen. Skipping moves outside the view covers that end.

**This is display-only.** The stored toolpath, G-code and simulation still see every move, so it is not the toolpath decimation #664 rules out as a non-goal.

## Alternative if that is not enough

[...]

Issue #676, opened 2026-08-29
Trochoidal pocket clearing: offset ring centrelines as trochoidal guides


zabooma:
Deferred from #659, which held it back deliberately: *"It is deliberately not a child of this umbrella. Filing it now is speculative scope under Backlog Contract R5 — it becomes fileable once #662 has said what is affordable."* #662 shipped in #672, so the number now exists and this is fileable.

## The idea

Pocket clearing uses offset ring centrelines as **trochoidal guides**: the cutter orbits along each ring instead of dragging a full-width contour, and the channel acts as a virtual tool wider than the cutter. Peak radial engagement drops the same way it does on an edge route, which is what lets a hobby router use more flute length without chatter.

The geometry already exists on both sides — `pocket.ts` produces the ring tree, `trochoidalEdge.ts` samples an orbit around any closed guide. What was missing was whether the result fits in one operation.

## It now fits — measured

`buildTrochoidalContour` on `9cd0325`, `D = 6 mm`, channel `W = 1.5 x D`, `advance = 0.1 x D`, against #659's own worked example (a 200 x 150 pocket filled with 9 mm channels, ~3,600 mm of ring centreline):

| | at the time of #659 | on `9cd0325` |
| --- | --- | --- |
| points per mm of guide | 60 | **40.0** |
| points per level | ~216,000 | **144,025** |
| ceiling | 500,000 | **1,000,000** |
| levels affordable | **2** | **6** |

#659 called this "not buildable at the current ceiling". #660's sagitta bound took a third off the per-mm cost and #662 doubled the ceiling, so a six-level pocket is now inside one operation's budget where two levels were not.

## The catch, and it is a real one

Six levels of that example is ~864,000 emitted moves. #664 measures the app unusable at **249,663**. So this feature is affordable by the *budget* and squarely inside the range where the product currently stops responding — about 3.5x past it.

That is not an argument against the feature; it is an argument about ordering. **#664 should land first**, or the first thing a user does with trochoidal pocketing is generate a project they cannot then interact with, with no warning telling them why. #662 deliberately left that advisory to #664.

## Scope

- Ring centrelines from the existing pocket tree used as trochoidal guides, with the channel width as the effective clearing width.
- Guide-domain fragmentation, entry synthesis and the safety backstop reused from `trochoidalEdge`/`edge.ts` rather than re-derived — clipping a generated orbit is what `planning/TROCHOIDAL_EDGE_DESIGN.md` forbids.
[...]

PR #681, opened 2026-08-30
Optimize 2D toolpath display rendering

zabooma:
## Status

**Work in progress: #679's usable pan/zoom acceptance is not yet demonstrated. Do not merge as a completed performance fix.**

## Summary

- Cache and viewport-cull display-only 2D toolpath geometry, collisions, debug markers and arrows.
- Coalesce connected same-feed runs up to **3 canvas pixels of accumulated travel**, restoring finer geometry as zoom increases.
- Make near-full culls cheap: a layer that reaches 50% visibility (and at least 512 segments) draws its cached full list; sparse culls use reusable integer stamps and source-order compaction instead of a `Set` plus sort.
- Keep two recent scale/detail variants per toolpath, so an operation-booklet snapshot does not evict the live canvas cache. Snapshots retain every emitted move rather than using the interactive merge.
- Skip direction-arrow calculation and drawing during navigation. Restore after pointer release/cancel or 150 ms wheel inactivity without changing the Directions toggle. Debug and collision overlays are preserved.
- Add the requested parallel-raster trochoidal Engrave project as a larger workload for #676: **882,899 moves, no warnings**, using the existing generator. No pocket generator, machining logic, saved format or simulation changes.
- Add unit/real-fixture geometry checks and browser coverage for pan, wheel settling, pinch cancellation and Directions-off persistence. Temporary timing instrumentation was removed.

## Performance evidence and remaining limit

At raster fit-to-view scale, displayed segments fall from 882,720 to 439,488; maximum measured displacement is 0.1813 canvas pixels (rounded upward). Both generated ToolpathResult records retain identical before/after hashes. At closer zoom the original cutting detail is restored.

The review fix covers the problematic near-fit cull case. In a real generated raster query at 2× fit with a 1440×900 centred viewport, the layer takes the full-list fallback rather than materializing a mostly-visible subset; five Node/V8 query samples were 9–23 ms. This is pure-JS evidence, not browser frame timing.

Actual-app comparisons in isolated Chromium software rendering show reduced draw-callback work, but **callback durations are not frame times**. Raster pan frame delivery remained slow and wheel changes still rebuild display geometry at new scales. This is useful headroom, not proof of readiness for #676.

[Full measurements, methodology, fixture parameters, hashes and remaining limitation](https://github.com/PureCutCNC/purecutcnc/issues/679#issuecomment-5469031157).

## Verification

- Rebased onto main `4a00a58`.
- `npm run build`: **224 structural test files passed**, production bundle built; lint clean.
- Focused visibility/navigation browser spec: **4 passed**, including real two-finger browser input.
- Full isolated browser suite before this review fix: **175 passed, 1 failed**. The simulation camera-menu case did not find its button; its failure snapshot still showed Sketch selected. The complete viewport-menu spec then passed **3/3** in isolation on the same commit. No unrelated production/test change was made to hide this failure.
- Current head: `de948f1`. The PR stays draft for the unmet responsiveness acceptance.

Relates to #679


zabooma:
## Review of `9908aa7` — not ready to merge

CI is green on head `9908aa7` (all 8 checks) and `mergeable_state` is clean. Two blockers, one mechanical and one measured.

### 1. `Closes #679` contradicts the PR's own status

The body says "Work in progress: #679's usable pan/zoom acceptance is not yet demonstrated. Do not merge as a completed performance fix", and the branch is draft — but it also carries `Closes #679`. Merging as written closes an issue the same body says is unsatisfied. This needs to become a plain reference before merge, whatever happens to the rest.

### 2. The viewport cull regresses ordinary zoom on this PR's own fixture

`drawToolpath` calls `visibleDisplaySegments` on every frame, for every visible layer. A pan holds scale constant, so the geometry cache hits and the query is the whole per-frame cost.

Measured on this branch, 1440x900 viewport centred on the model, cuts layer, 5 timed runs after a warm-up:

| `trochoidal-raster.camj` | cut segments | query returns | query cost |
| --- | ---: | ---: | ---: |
| fit | 435,456 | all (fast path) | 0.0 ms |
| **2x fit** | 870,912 | **774,144** | **499 ms** |
| 4x fit | 870,912 | 195,840 | 81 ms |
| 8x fit | 870,912 | 49,032 | 32 ms |

At 2x fit — one or two wheel notches past fit — the cull rejects 11% of the layer and spends half a second per frame doing it.

The cost is the `Set` insert, `sort` and `map` over matched indices in `visibleSegmentIndices` (`toolpathDisplay.ts:177-201`): roughly 0.5 µs per segment returned. #679's own figure for plain path submission is ~18 ms per 249,663 moves, about 0.07 µs each. Deciding a segment is visible therefore costs several times more than submitting it, so culling only pays when it rejects nearly everything.

**Why the existing measurements did not show it.** They were taken at fit-to-view, where `contains(viewport, index.bounds)` short-circuits to "draw all" and the cull never executes. The reported 35.3 ms raster pan is consistent with drawing 439,488 merged segments and nothing else — that win came from the 3 px merge and arrow deferral, not from culling. The cull is inert exactly where it was measured.

`trochoidal-249k.camj` cannot expose it either: at 1-2x fit the model fits the canvas (fast path), and at 4-8x a centred viewport sits in the empty middle of the rectangle route, so it returns 0 segments in 0.0 ms.

**Fix direction.** The per-cell index lists are already ascending, so a k-way merge or a stamp array avoids both the `Set` and the `sort`; and an early bail to "draw all" once matched exceeds some fraction of the layer caps the worst case.

**Methodology.** The branch's own `toolpathDisplayGeometry` / `visibleDisplaySegments`, driven from the two tracked fixtures through their real generators, run under Node/V8 in a Linux container — not a browser. The query is pure JavaScript, so the shape and the per-segment cost hold, but these are not browser frame times and should not be quoted as such.

### Smaller findings

- **Snapshot rendering thrashes the display cache.** `displayGeometryCache` keeps one scale per toolpath, and `operationSnapshot.ts:263` draws the same `ToolpathResult` at its own scale, so generating booklet snapshots evicts the live sketch geometry and forces a full rebuild on return to the canvas. Two entries, or keying by scale, fixes it.
- **Booklet output changes, with nothing covering it.** Snapshots share `drawToolpath`, so exported images now go through the 3 px merge. `operationSnapshot.ts` has no test file and is not touched here. "No saved format changes" is accurate, but exported imagery does change.

### What is solid

- `toolpathDisplayFixture.test.ts` is the right kind of test: real generated toolpaths, every original cutting vertex measured against its replacement chord, a 0.25 px bound asserted, closer zoom asserted to restore full detail, and `toolpath.moves` identity checked so the machining record provably survives.
- The arrow-deferral lifecycle is properly scoped — pointer set, 150 ms idle, `blur`, `dispose` — and never writes the user's Directions setting.
- The new spec is registered in the `workflow-ui` lane, so `check:e2e-lanes` will keep it running.

The 3 px merge and the arrow deferral are worth keeping as they stand. It is the culling half that needs the fix before this is a merge candidate.

---

zabooma:
Addressed in `de948f1`.

- Replaced `Closes #679` with a plain `Relates to #679`; the PR remains a draft and #679 is not represented as complete.
- Reworked the per-pan cull query: reusable `Uint32Array` stamps deduplicate candidates, sparse results compact in source order, and a partial viewport that reaches 50% of a layer (minimum 512 segments) immediately uses the cached full list. This avoids the previous `Set`/sort/materialization path in the near-fit raster case.
- Kept two recent `(scale, detail)` geometries per toolpath, so a full-detail booklet snapshot no longer evicts the live canvas cache. Snapshots explicitly retain all emitted moves.
- Added structural coverage for the full-layer fallback, source ordering/stamp reset, snapshot cache residency, and full-detail drawing.

Verification after rebasing onto `4a00a58`:

- `npm run build` — passed (224 structural test files, lint, production build).
- The real generated raster 2×-fit, 1440×900 centred query takes the full-list fallback; five Node/V8 samples were 9–23 ms. This is query-only evidence, not browser-frame timing.

The PR remains draft because the broader responsiveness acceptance is still unmet.

Issue #697, opened 2026-08-31
feat(toolpaths): constant-scallop 3D finish strategy, with a slope filter


zabooma:
## Problem

The 3D surface finish operation has no strategy that holds a finish on shallow,
organic surface — which is the surface a carved instrument top, a relief, or any
sculpted model is mostly made of.

- **Waterline** places passes at constant Z. On a surface of slope θ the XY gap
  between contours grows as 1/sin(θ), so on near-horizontal ground the passes
  spread out without bound. The adaptive refinement exists to fill those gaps and
  does work, but it is handed a fixed ring spacing rather than deriving one, and
  it is charged against a fixed global budget.
- **Parallel** covers everything but at a spacing that is constant in XY, not
  along the surface, so the cusp varies with slope and the scanlines cut across
  the flow of the form. On a carved top the witness marks are visible in the
  scallop map before they are visible in the wood.
- **3D surface cleanup** finishes flat floors correctly (measured: 10 µm p90,
  0 % missed on a flat pocket floor) but is a terrace cleaner, not a surface
  finisher — 733 µm p90 on 5–20° ground.

Measured on a 332 × 470 mm carved guitar top (6 mm ball, 20 µm cusp budget,
0.692 mm surface spacing), at essentially identical cutting length:

| Strategy | Cut | Scallop p90 | p99 | Never cut | Over 20 µm |
| --- | --- | --- | --- | --- | --- |
| waterline | 257 m | — | — | **29.8 %** | 56.4 % |
| parallel | 232 m | 22.5 µm | 85 µm | 0.0 % | **25.7 %** |
| constant scallop (prototype) | 238 m | **15.0 µm** | **54 µm** | 0.0 % | **3.8 %** |

On the 62 % of that part which is the ~1° dome, out-of-budget area falls from
35.4 % to 0.7 %.

## Why this is not a patch to waterline

The 1/sin(θ) spread is what constant-Z *is*, not a defect in the implementation.
Forcing waterline to hold a cusp on shallow ground means spending the motion the
geometry demands: on `work/Makera_Model.camj`, with the ring budget lifted so it
actually holds its requested spacing, waterline cuts **3,107 in where single
coverage of the same area needs 443 in** — 7× redundant, 311 min against
parallel's 44 min, for a worse surface than the prototype produces in 51 min.

There are two genuine bugs in the current adaptive refinement — a global 1,000-ring
budget spent top-down with no user-visible warning (#698), and flat-topped features
[...]

PR #703, opened 2026-09-01
Add finish-surface slope filtering

### LINKED ISSUE #702: feat(toolpaths): optional slope filter for 3D surface finish (697 phase 1)
## Parent and scope

Phase 1 of #697: add an optional surface-slope range to `finish_surface`, usable with both existing `parallel` and `waterline` patterns. This is the first independently shippable slice; #697 stays open for the constant-scallop core, linking, cusp controls, rest pass and preset.

Implementation approval is pending. No production code has been changed.

## Implementation plan

1. Add optional minimum/maximum slope angles in degrees to the operation model, with disabled/unset as the default. Expose them only for 3D surface finish. Keep old files on the exact existing generation path; add normalization, save/load, undo/redo and toolpath-cache invalidation coverage. Validate finite angles in 0–90 degrees and ordered bounds. Invalid enabled ranges must produce a visible warning and no motion, never silently become unrestricted machining.
2. Build the slope field from the cutter-location surface sampled with the existing `safeToolTipZAt` and bounded height-map machinery. Compute gradients in project units and convert the slope to degrees. Keep missing samples distinct from flat ground; explicitly handle grid edges, discontinuities and regions with no eligible cells. Reuse existing height-map limits and visible warnings; do not introduce an unrestricted grid allocation.
3. Convert eligible cells into a tool-centre XY domain before generation. Resolve the user's ordered include/exclude regions with their existing cutter-clearance rules, then intersect with the slope domain through `resolveRegionDomainCentre`. For the slope mask itself use zero additional inset: it already describes permissible tool-centre positions. This preserves normal region-mask overreach without silently expanding the chosen slope range by a cutter radius. The tool footprint can cross a slope boundary; the centre must remain inside the accepted domain, within the documented grid tolerance. Intentional seam overlap will come from overlapping angle ranges, not hidden dilation.
4. Apply that same domain to parallel scanlines and all waterline contour/refinement emission paths. Do not delete finished moves after generation or turn mask edges into extra boundary cuts. Preserve gouge protection, stock-to-leave behavior, tabs, clamps and related-feature restrictions. At-depth links must stay inside the resolved domain as well as passing the existing surface-clearance check; disconnected pieces retract/travel at the existing operation safe Z. No new keep-down or spiral optimization in this phase.
5. Add localized controls/help text using the existing operation-settings UI and store actions. Document that this is tool-dependent CL-surface slope measured from horizontal, with the existing height-field/undercut limitation. Update the region design reference and nearest indexes only where their contracts/files change.

## Verification and acceptance

- Capture pre-change generated moves and posted G-code for the repository's tracked finish fixtures, both patterns and existing region/tab/clamp combinations. Unset slope fields and old saved operations must retain byte-identical output after the change. Loading and saving a filtered operation retains its bounds; changing or clearing bounds invalidates CAM output.
- Use tracked analytic flat, inclined and mixed-slope fixtures to verify membership independently of the new slope helper. Cover min-only, max-only, full-range, narrow bands, invalid bounds, missing samples, rotated geometry and mm/inch equivalence.
- Verify a 0–30 degree parallel filter on the guitar-top fixture specified by #697 using geometric segment/interior sampling, not move counts. Locate and retain its reproducible source in the tracked fixture/diagnostic set; do not substitute a synthetic result for the issue's named acceptance fixture. Any unavailable source is reported before claiming acceptance.
- Assert segment interiors and cutting links respect the composed domain, including holes/disjoint regions; assert safe-Z travel across excluded bands. Check both regular and adaptive waterline output, and independently verify protected geometry/gouge constraints.
- Extend an existing CAM-settings browser smoke test for enable/edit/clear, pattern switching, save/reload and visible validation. Verify translations. If a new spec is needed it must join a CI lane.
- Run focused engine/store tests, isolated affected Playwright coverage, and `npm run build`. Rebase onto current main and repeat the required gate before the PR. PR closes this child only, not #697.

## Current-code evidence

Checked against `31b36dbe256b68b11b702e44c1e79b1c607c0db2`:

- `generateFinishSurfaceToolpath` dispatches the existing two strategies and owns the final safe-Z retract.
- `generateFinishSurfaceParallel` builds/caches its height map internally; `safeToolTipZAt` is the existing cutter-location sampling seam.
- Both `generateFinishSurfaceParallel` and `generateFinishSurfaceWaterline` already use `resolveRegionDomainCentre`; the resolver dilates mask entries only when a nonzero inset is supplied.
- `operationComputationEquals` is the operation-cache comparison seam that must include the new settings.

## Non-goals

No constant-scallop pattern yet; no change to the existing parallel/waterline behavior when the filter is unset; no adaptive-budget or flat-top fix from #698/#699; no cutter-down linking optimization; no new scallop parameter, automatic steep/shallow pair or rest machining. This phase makes no new finish-quality or timing claim.

zabooma:
## Why

Finish Surface operations currently cover the whole allowed model footprint, so users cannot target shallow or steep cutter-location (CL) surface bands independently. This matters for workflows that split finishing between parallel and waterline strategies.

## What changed

- Add optional minimum and maximum CL-surface slope bounds, in degrees from horizontal, to Finish Surface operations.
- Apply the same heightmap-derived slope domain to Parallel and Waterline paths after region composition, while preserving tabs, clamps, and protected footprints.
- Require exact in-domain links and safe-Z retracts across disconnected slope bands; malformed saved values fail closed with a visible warning.
- Add CAM controls, persistence/cache coverage, localized labels and warnings, analytic fixtures, an original synthetic guitar-top fixture, and legacy-output parity hashes.
- Keep the disabled path byte-identical to existing motion and posted G-code.

On the tracked 332 × 470 mm synthetic guitar-top fixture with a 6 mm ball endmill and a 0–30° filter, cutting travel changed from 232,433.479 mm to 222,310.786 mm. Exact-domain retracts increased rapid travel from 24.632 mm to 42,918.214 mm; link optimization across the filtered bands remains outside this phase. These are path metrics for the procedural regression fixture, not a surface-quality claim.

## Verification

- `npm run build` — 232 structural test files passed and the production bundle completed
- `PURECUT_E2E_ISOLATED=1 PURECUT_E2E_PORT=1494 npx playwright test e2e/camOperations.smoke.spec.ts --grep 'surface slope filter' --workers=1`
- Baseline parity coverage for existing Finish Surface operations and posted G-code
- Mutation checks confirmed the focused tests reject disabled filtering, bypassed exact link checks, and omitted cache comparisons

Closes #702

Parent #697 remains open for the later phases.


src/engine/toolpaths/finishSurfaceSlope.ts
```diff
@@ -0,0 +1,143 @@
...
+  if (!pointInClipperPaths(paths, from) || !pointInClipperPaths(paths, to)) return false
+  const a = { X: Math.round(from.x * DEFAULT_CLIPPER_SCALE), Y: Math.round(from.y * DEFAULT_CLIPPER_SCALE) }
+  const b = { X: Math.round(to.x * DEFAULT_CLIPPER_SCALE), Y: Math.round(to.y * DEFAULT_CLIPPER_SCALE) }
+  if (a.X === b.X && a.Y === b.Y) return true
+  const clipper = new ClipperLib.Clipper()
+  addOpenSubject(clipper, [a, b])
+  clipper.AddPaths(paths, ClipperLib.PolyType.ptClip, true)
+  const tree = new ClipperLib.PolyTree()
+  clipper.Execute(ClipperLib.ClipType.ctDifference, tree,
+    ClipperLib.PolyFillType.pftNonZero, ClipperLib.PolyFillType.pftNonZero)
```
zabooma: **Exact link check treats interior holes as solid, so it passes links that cross an excluded band.**

The clip is filled `pftNonZero`, but the parallel strategy hands this function a domain whose hole orientation has been destroyed. `finishSurfaceParallel.ts:773` builds `linkDomain` through `coverageContoursToClipperPaths`, which runs every contour through `normalizeWinding(contour, false)` — outers *and* holes come out with the same winding, so under non-zero fill every hole is filled in and the difference finds nothing to reject.

The endpoint test above does not catch it: `pointInClipperPaths` is parity-based (XOR over paths), so it handles holes correctly. Only the segment interior is affected — exactly the case the comment on line 130 claims to cover.

Reproduced on a 100 × 100 mm plain with a steep cone in the middle, `finishSlopeMin: 0 / finishSlopeMax: 15`, driving the real `buildSurfaceSlopeDomain` → `intersectSurfaceSlopeDomain`, then replicating `coverageContoursToClipperPaths` verbatim:

```
domain contours: 2 (outer + hole)
endpoints in domain: true true
cone centre excluded: true
link check on well-oriented domain : false   (correct)
link check on linkDomain as built  : true    (should be false)
```

So a scanline-to-scanline link straight across the excluded cone is accepted and emitted as a cut-feed link. The kinematic Z check still runs, so this is not a gouge into the model — but it is a feed-rate traverse across ground the user excluded, and the same hole class covers protected/clamp footprints interior to the coverage, which the height map does not model at all.

Smallest fix is one line here — clip with even-odd, which also makes this function agree with its own endpoint test:

```ts
clipper.Execute(ClipperLib.ClipType.ctDifference, tree,
  ClipperLib.PolyFillType.pftNonZero, ClipperLib.PolyFillType.pftEvenOdd)
```

Both callers pass non-overlapping Clipper output, so even-odd is safe. Verified against the same fixtures:

```
crossing hole  (even-odd fill): false   (want false)
clear of hole  (even-odd fill): true    (want true)
well-oriented  (even-odd fill): false   (want false, unchanged for waterline)
two disjoint islands, gap crossed: false (want false)
```

The alternative is to stop round-tripping through tuple contours in the parallel path and keep the oriented `ClipperPath[]` — cleaner, but it touches `subtractProtectedContours`.

Worth a regression test with an interior hole in the domain: `finishSurfaceSlopeGuitar.test.ts`'s `inShallowBand` oracle scans a 3 × 3 cell neighbourhood, so it would not fail on a hole-crossing link unless the hole is more than one cell wide along the whole crossing.

---

src/engine/toolpaths/finishSurfaceParallel.ts
```diff
@@ -757,7 +767,10 @@ export function generateFinishSurfaceParallel(
   const coverageContours = regionMask
     ? clipperPathsToTupleContours(resolveRegionDomainCentre(baseCoveragePaths, regionMask, centreInset))
     : baseContours
-  const clippedContours = subtractProtectedContours(coverageContours, protectedPaths)
+  const slopeContours = slopeDomain === null ? coverageContours
+    : clipperPathsToTupleContours(intersectSurfaceSlopeDomain(coverageContoursToClipperPaths(coverageContours), slopeDomain))
+  const clippedContours = subtractProtectedContours(slopeContours, protectedPaths)
+  if (slopeDomain !== null) linkDomain = coverageContoursToClipperPaths(clippedContours)
```
zabooma: This is the source of the hole problem detailed on `finishSurfaceSlope.ts:141`. `coverageContoursToClipperPaths` (line 15 of this file) normalizes **every** contour to one winding:

```ts
.map((contour) => toClipperPath(
  normalizeWinding(contour.map(([x, y]) => ({ x, y })), false),
  DEFAULT_CLIPPER_SCALE,
))
```

`clippedContours` is Clipper output, so its holes arrive correctly oriented and then lose that here. It does not matter for the existing callers (`unionClipperPaths` is fed the same paths and rebuilds orientation, and `scanlineIntervalsForContours` pairs crossings even-odd, so cut segments do respect holes) — but `linkDomain` is consumed by a non-zero-filled Clipper difference, where it does matter.

If you take the even-odd fix in the other comment, this line can stay. Otherwise, build `linkDomain` from the oriented paths instead, e.g. keep the `ClipperPath[]` that `subtractProtectedContours` already computes internally rather than re-deriving it from tuples.

---

src/engine/toolpaths/finishSurfaceSlopeParity.test.ts
```diff
@@ -0,0 +1,157 @@
...
+
+// Captured BEFORE #702's generator changes at 31b36dbe256b68b11b702e44c1e79b1c607c0db2.
+// Both the raw toolpath result and posted program must remain byte-identical.
+const baseline: Record<string, {moves: number; motion: string; gcode: string}> = {
+  "3d-imported-block-test3.camj/op6792424/parallel": {
+    "moves": 3915,
+    "motion": "24336cbc8f719d4f7742fbfa1b97617a88337fa36067e97ce26ba8619704dbcc",
+    "gcode": "4bdc3fd0790fc22797aecfa1acd4db4ca09e0089fcaa12fa0313ef2dd8ff1500"
+  },
+  "3d-imported-block-test3.camj/op6792424/waterline": {
```
zabooma, line 37: **These baselines are already stale against `main`, and the PR is in conflict there.**

The header records they were captured at `31b36db`, which is this PR's merge base. `main` has since merged #704 (`c2354aa`, "machine flat tops in the waterline finish"), which deliberately changes waterline output. I copied this file into a `main` worktree and ran it unchanged:

```
not ok 1 - 3d-imported-block-test3.camj: ... expected: 2679  actual: 3321
not ok 3 - model-in-pocket.camj: ...      expected: 1468  actual: 3836
# pass 1   # fail 2
```

`issue-401-cone-finish.camj` still passes, so it is only the two fixtures with flat tops — consistent with #704 and not with anything in this PR. Every `waterline` row here will need re-capturing once `main` is merged in (`git merge-tree` shows conflicts in `finishSurface.ts`, `finishSurfaceWaterline.ts`, `src/engine/toolpaths/INDEX.md` and `src/test/INDEX.md`).

Two asks when you re-capture: update the "Captured BEFORE …" comment to name the new base commit, and confirm the delta is exactly #704's — a regenerated hash proves nothing about which change moved it.

---

src/engine/toolpaths/finishSurfaceWaterline.ts
```diff
@@ -1751,6 +1752,11 @@ export function generateFinishSurfaceWaterline(
     heightMapBbox,
     heightMapCellSize,
   )
+  const slopeMask = buildSurfaceSlopeDomain(operation, baseHeightMap,
+    (x, y) => safeToolTipZAt(x, y, baseHeightMap, tool), warnings)
+  const slopeDomain = slopeMask === null ? null
+    : intersectSurfaceSlopeDomain(compositeAllowedForRegion ?? modelSilhouettePaths, slopeMask)
+  if (slopeDomain !== null && slopeDomain.length === 0) return { moves: [], stepLevels: new Set() }
```
zabooma: **Silent empty result when the slope band survives the mask but not the composition.**

`finishSlopeEmpty` is pushed in one place only — `finishSurfaceSlope.ts:116`, when the *mask* union comes out empty. Here the mask was non-empty and it is the intersection with `compositeAllowedForRegion ?? modelSilhouettePaths` that emptied it, so this returns zero moves and zero warnings: the user sees an operation that produces nothing and no explanation.

The reachable case is a region-restricted operation — the height map is bounded by `clampExpandedBoundsToModel(regionBounds, …)`, so it covers cells outside the region, and those can carry the whole eligible band while the region itself carries none of it. (I confirmed the code path by reading; I did not manage to build a region fixture that lands on it, so treat the trigger as unproven and the branch itself as plainly warning-less.)

Suggest pushing `finishSlopeEmpty` on this branch too, and the same for the parallel path — there, an empty `slopeContours` leaves `clippedBounds` null and the only messages on that branch are behind `operation.debugToolpath`.

---

src/engine/toolpaths/finishSurfaceSlope.ts
```diff
@@ -0,0 +1,143 @@
...
+  }, 0)
+}
+
+/** Exact open-segment difference catches even a narrow hole between samples. */
+export function segmentInSurfaceDomain(paths: ClipperPath[], from: Point, to: Point): boolean {
+  if (!pointInClipperPaths(paths, from) || !pointInClipperPaths(paths, to)) return false
+  const a = { X: Math.round(from.x * DEFAULT_CLIPPER_SCALE), Y: Math.round(from.y * DEFAULT_CLIPPER_SCALE) }
+  const b = { X: Math.round(to.x * DEFAULT_CLIPPER_SCALE), Y: Math.round(to.y * DEFAULT_CLIPPER_SCALE) }
+  if (a.X === b.X && a.Y === b.Y) return true
+  const clipper = new ClipperLib.Clipper()
```
zabooma: **Each link test builds a fresh `Clipper` and re-adds the whole domain — ~1 ms per call, and it dominates the filtered-run cost.**

Measured on this PR's own guitar fixture (332 × 470 mm, 1 mm mesh, 6 mm ball, `0–30°`), instrumenting this function with a call counter and timer:

| | baseline | with filter | inside `segmentInSurfaceDomain` |
|---|---|---|---|
| parallel | 2.49 s | 6.24 s | 2,356 calls / 2.19 s |
| waterline | 31.2 s | 46.2 s | 4,001 calls / 5.45 s |

So the exact check is most of the parallel regression on its own, at ~0.93 ms (parallel) and ~1.36 ms (waterline) per segment. This is on the interactive path — `operationComputationEquals` now compares both slope fields, so every edit to any computation field re-runs it.

A bounding-box reject before the Clipper call would remove nearly all of it: precompute each domain path's XY bbox once, and only run the difference when the segment's bbox overlaps a hole or crosses the outer boundary. The endpoint parity test above already handles the common in-the-open case.

Not blocking, and no change to the disabled path — but worth doing before this grows to the remaining #697 phases.

---

src/store/INDEX.md
```diff
@@ -3,7 +3,8 @@
 Zustand store. The single source of truth for the current `.camj` project. **All project mutations must go through actions on `projectStore` — never mutate state directly.**
 
 ## Files
-- `projectStore.ts` — store composition root: initial state, shared dependencies, and slice assembly
+- `projectStore.ts` — store composition root
+- `surfaceSlope.test.ts` — slope-setting normalization, save/open, clearing, undo/redo, malformed-input fail-closed behavior, and cache invalidation lifecycle: initial state, shared dependencies, and slice assembly
```
zabooma: The tail of the old `projectStore.ts` entry ("initial state, shared dependencies, and slice assembly") got carried onto the new line, so it now describes the test file and `projectStore.ts` lost its own description:

```md
- `projectStore.ts` — store composition root
- `surfaceSlope.test.ts` — … and cache invalidation lifecycle: initial state, shared dependencies, and slice assembly
```

Should be:

```md
- `projectStore.ts` — store composition root: initial state, shared dependencies, and slice assembly
- `surfaceSlope.test.ts` — slope-setting normalization, save/open, clearing, undo/redo, malformed-input fail-closed behavior, and cache invalidation lifecycle
```

---

zabooma (commented):
Reviewed at `55e844d`, with the branch checked out and the new suites run locally (`finishSurfaceSlope`, `finishSurfaceSlopeGuitar`, `finishSurfaceSlopeParity` all pass; 0.9 s / 11.3 s / 5.5 s).

The shape of the feature is right: slope is derived from the CL height map rather than the mesh normals, composed as a second tool-centre domain at zero extra inset, malformed values fail closed before any geometry work, and the disabled path really is inert (`transitionToCutEntry`'s `safeLinkCheck` is optional, so passing `undefined` is identical to not passing it). i18n is complete across all five locales and `warningsCoverage` is updated. The conservative gradient — larger one-sided derivative per axis, `NaN` for a missing interior neighbour rather than a zero — is the correct choice at a ridge.

Six findings inline, ranked:

1. **`finishSurfaceSlope.ts:141` — the exact link check does not reject links crossing an interior hole.** The parallel path's `linkDomain` is rebuilt with all contours at one winding, and the clip is non-zero-filled, so holes fill in. Reproduced on a cone-in-a-plain fixture: a link straight across the excluded cone is rejected on the well-oriented domain and accepted on the domain as built. One-line even-odd fix verified against hole-crossing, disjoint-island and control cases.
2. **`finishSurfaceParallel.ts:773`** — where that winding is lost, with the alternative fix.
3. **`finishSurfaceSlopeParity.test.ts:36` — the parity baselines are already stale.** #704 landed on `main` after this branch's merge base and changes waterline output; running this file unchanged against `main` fails two of three fixtures (2679 → 3321, 1468 → 3836 moves). The branch is also in conflict with `main` on `finishSurface.ts`, `finishSurfaceWaterline.ts` and two `INDEX.md` files.
4. **`finishSurfaceWaterline.ts:1759` — silent empty result** when the mask survives but the composition empties it; `finishSlopeEmpty` only covers the empty-mask case.
5. **`finishSurfaceSlope.ts:136` — ~1 ms per link test**, a fresh `Clipper` with the whole domain re-added each call; measured as most of the parallel 2.49 s → 6.24 s regression on the guitar fixture.
6. **`src/store/INDEX.md:7`** — a sentence fragment landed on the wrong entry.

Two smaller notes, no comment needed:

- The waterline disables `simplifyContiguousCutMoves` for the whole operation whenever a filter is active, not just near the mask boundary. Measured cost on the guitar fixture is small — 265,935 moves against 263,987 with simplification restored, 0.7 % — so the blunt version looks fine to keep; worth revisiting only if a fixture shows worse.
- In `finishSurface.ts` the new `'invalid'` early return was inserted between the `isWaterline` comment block and the line it describes.

The rapid-travel figure in the description (24.632 mm → 42,918.214 mm) is a consequence of requiring exact in-domain links, and is honestly labelled as such — but note finding 1 means the current number was produced by a check that is *more* permissive than intended, so it will move again once holes are handled.

---

PR #713, opened 2026-09-02
fix(toolpaths): stop finishing air at the depth limit, and honour axial stock in parallel

### LINKED ISSUE #711: fix(toolpaths): parallel and constant-scallop finish machine a flat pass at the depth clamp instead of skipping unmachinable area
## Problem

When a `finish_surface` operation's depth is limited by a **related subtract**
feature — a 2.5D pocket carved around or through the model — the two height-map
strategies do not stop where they run out of machinable surface. They emit a
flat cutting pass at the depth limit instead.

Everywhere the model's cutter-location surface sits *below* that limit, every
contour point collapses onto the limit Z and is still emitted as a **cut** move.
The result is a large flat pass at the pocket floor, at cutting feed, over area
the pocket operation already cleared.

Reported from `work/Oldman-splash-final.camj`, where the 3D model is partly
buried in the body below a pocket floor: *"finish is aware of that … but there
is lots of air cutting at the floor level."*

**Affects `parallel` and `constant_scallop`. `waterline` is exempt** — for a
structural reason, not a better one; see below.

## Measured

`work/Oldman-splash-final.camj`, operation `op0024` (`3D Surface finish`), 1/8"
ball, stepover 0.175, no slope filter. The model's `Rect 2` subtract has
`z_bottom = 0.2`, and **38.9 % of the model's cutter-location surface sits below
that**, dipping as far as **0.179" under it**.

| pattern | cut length | at exactly Z = 0.2 | share of cut length |
| --- | --- | --- | --- |
| `constant_scallop` | 287.8 in | 78.5 in | **27.3 %** |
| `parallel` | 242.0 in | 65.0 in | **26.9 %** |
| `waterline` | 328.5 in | 23.9 in | 7.3 % (its genuine lowest ring) |

All **13,303** clamped moves on the constant-scallop run fall inside the
pocket's XY footprint (x 0.125–3.875, y 0.125–2.875) — precisely the area the
project's own `Pocket Finish` operation already took down to 0.2. So the passes
are redundant rather than dangerous: the tool is skimming an existing floor, not
plunging into solid stock.

At the operation's 30 ipm that is **157 minutes of its 576 minutes of cutting**.

## Root cause

1. The subtract's footprint intersects the model silhouette, so it is a *related
   subtract*, and
   [`finishSurface.ts:296`](../blob/main/src/engine/toolpaths/finishSurface.ts#L296)
   raises the whole operation's floor to it:

   ```ts
   effectiveBottom = Math.max(effectiveBottom, deepestRelatedBottom)   // → 0.2
   ```

   This part is correct and is what the reporter means by "finish is aware of
   that".

2. [`finishSurface.ts:354`](../blob/main/src/engine/toolpaths/finishSurface.ts#L354)
   hands the strategies `minCutZAtPoint`, which returns that floor (or a
   per-point subtract bottom inside a subtract's clearance paths).

3. Both height-map strategies lift every contour point with the same clamp —
   `finishSurfaceConstantScallop.ts`'s `liftFragment` and the equivalent in
   `finishSurfaceParallel.ts`:

   ```ts
   z: Math.max(surfaceZ + axialLeave, floor)
   ```

   The clamp exists to stop a pass going *deeper* than allowed. Nothing asks the
   prior question — whether there is any machinable surface at this XY at all —
   so "I am not allowed to reach the surface here" silently becomes "machine a
   flat pass at the limit".

**Why waterline is exempt:** it never receives `minCutZAtPoint`. It builds Z
levels from the top down to `effectiveBottom` and simply has no level below the
floor, so there is nothing to clamp up. Its 7.3 % above is a real bottom ring,
not clamped surface.

## Reproduction

`work/Oldman-splash-final.camj` is untracked and lives only in the primary
checkout, so it cannot be the acceptance fixture. The shape reproduces on any
project with:

- an imported mesh whose surface descends below some Z, and
- a `subtract` feature intersecting the model silhouette with `z_bottom` above
  the mesh's low point.

Generate the finish with `pocketPattern: 'parallel'` and `'constant_scallop'`
and count cut moves whose endpoints both sit within 1e-6 of the subtract's
bottom Z.

## Proposed fix

`liftFragment` already knows how to break a pass — it flushes the current
fragment when `safeToolTipZAt` returns a non-finite sample. The same break
should happen when the surface sits below the floor by more than a tolerance: a
point 0.18" under the limit is not a cut, it is a hole in the machinable region.
The equivalent break belongs in the parallel scanline builder.

### The judgement call this needs

Clamping is **not wrong in general.** "Cut as deep as you are allowed" is
defensible when nothing else machines that floor, and dropping the pass
unconditionally would leave stock in a project that has no pocket operation
there. What makes it waste here is that the floor belongs to a *related
subtract*, which by construction has its own clearing operation that owns it.

That discriminator is already in the data: `safeSubtractBottomZAtPoint`
returning non-null means "you are inside a subtract another operation clears".
So the candidate rule is to break the fragment where the surface is below the
floor **and** the floor came from a related subtract, and keep today's clamp
where the floor is the operation's own `effectiveBottom`. Whether the simpler
unconditional break is also acceptable is the thing to settle before
implementing, and it should be settled with a measurement on a project of each
shape rather than by argument.

A tolerance is needed either way: a point a micron under the limit is the cutter
legitimately bottoming out on the floor, not a hole.

## Non-goals

- No change to how `effectiveBottom` or `minCutZAtPoint` are *computed*; the
  depth limit itself is correct.
- No change to waterline, which does not take this path.
- Not #526, which is the opposite defect — a **non-target** subtract changing the
  solid model but not the toolpath. This one is about a *related* subtract whose
  limit is applied and then machined.

## Acceptance

- On a tracked fixture of the shape above, `parallel` and `constant_scallop`
  emit **no cutting motion** over area whose cutter-location surface lies more
  than the tolerance below the depth limit, and the surviving passes are
  unchanged elsewhere — asserted on the emitted move stream, not on a move
  count.
- Byte-identical output for a finish operation with no related subtract, and for
  every existing parallel/waterline fixture.
- Mutation-verified: restoring the unconditional clamp puts the flat floor pass
  back, and the test says so.
- Re-measured on `work/Oldman-splash-final.camj` in the primary checkout, with
  the 27.3 % / 26.9 % figures above expected to go to ~0 and the rest of the
  toolpath unchanged.

## Priority note

Filed `Medium` because `High` is at its cap of 10 today. On impact it is a
`High` — it is shipped behaviour costing a quarter of a real finish
operation's cutting time — so it is worth promoting as soon as a slot frees.

### LINKED ISSUE #712: fix(toolpaths): parallel 3D surface finish silently ignores Stock to leave (axial)
## Problem

The `parallel` 3D surface finish **silently ignores Stock to leave (axial)**. The
field is rendered in its properties panel, the value is stored on the operation,
and the emitted toolpath is unchanged. A user who asks to leave 0.5 mm for a
later pass gets a part cut to final size.

`waterline` and `constant_scallop` both apply it. `finish_surface_cleanup`
cannot and says so (`cleanupStockToLeaveOffsets`). `parallel` is the only
strategy that accepts the value and does nothing, with no warning.

Found while scoping #711 — constant scallop shifted by exactly the leave on a
fixture where parallel did not move at all.

## Measured

Tracked guitar-top fixture, 6 mm ball, Stepover at the 20 µm cusp spacing. Each
row is its own process, so nothing shares `getCachedHeightMap`'s per-mesh cache:

| strategy | axial leave | cut points | min Z | mean Z | max Z |
| --- | --- | --- | --- | --- | --- |
| `parallel` | 0 | 655,518 | 0.0000 | 8.4298 | 19.6255 |
| `parallel` | **0.5** | 655,518 | **0.0000** | **8.4298** | **19.6255** |
| `constant_scallop` | 0 | 1,392,402 | −0.0845 | 9.0053 | 19.6279 |
| `constant_scallop` | **0.5** | 1,392,402 | **0.4155** | **9.5053** | **20.1279** |
| `waterline` | 0 | 524,388 | −4.0000 | 11.1703 | 19.6253 |
| `waterline` | **0.5** | 1,126,740 | **−3.5000** | 12.7362 | 19.6323 |

Parallel is identical in point count, minimum, mean and maximum. Constant
scallop shifts by exactly 0.5. Waterline responds too.

## Root cause

`finishSurfaceParallel.ts` contains **no reference to `stockToLeaveAxial`**. The
live per-point Z chain is:

1. `buildTopSurfaceSegment` — `z = queryHeightMapTopZ(...)`, the raw mesh
   surface (verified: at (30, 18.5) on a 45° ramp it returns 20.5000).
2. `splitAndClampSurfaceSegmentToMinZ` — raises z to the floor, splits on a
   floor change.
3. `applyGougeProtection` — raises z to the kinematic cutter-location surface
   (same point returns 21.2911, and 21.2911 is what the operation emits).
4. `clampSurfaceSegmentToMinZ` — raises z to the floor again.

No step adds the axial leave.

The dispatcher computes `axialLeave` but feeds it only into `effectiveBottom`
([`finishSurface.ts:263-264`](../blob/main/src/engine/toolpaths/finishSurface.ts#L263)),
and **that clamp provably never binds**: `safeToolTipZAt` is a maximum over the
cutter footprint that includes `d = 0`, so `surfaceZ ≥ modelBottomZ` and
therefore `surfaceZ + axialLeave ≥ modelBottomZ + axialLeave = effectiveBottom`.
Measured: 0.00 % of emitted cut points sit at the floor on the guitar and hills
fixtures, for both patterns, with and without axial leave.

So for parallel the value has exactly one consumer, and that consumer is inert.

## Why no test caught it

Every existing finish-surface test sets `stockToLeaveAxial: 0`.

## Proposed fix

Apply the leave where the cutter-location surface is established — after
`applyGougeProtection` and before the final floor clamp — so the emitted Z
becomes `max(clSurface + axialLeave, floor)`. That mirrors
`finishSurfaceConstantScallop.ts`'s `liftFragment`, which is the reference
implementation.

The leave must be added **after** gouge protection, not to the raw sample before
it: `applyGougeProtection` raises z to a kinematic minimum, so an offset applied
first would be partly swallowed wherever that minimum binds.

## Acceptance

- A parallel run at axial leave `L` is the run at 0 shifted up by exactly `L` at
  every point where the floor does not bind — asserted point-by-point on the
  emitted stream, not on aggregate statistics.
- Byte-identical output at `stockToLeaveAxial: 0` for every existing fixture and
  posted G-code fingerprint.
- The same claim holds in mm and inch.
- Mutation-verified: removing the offset again reproduces today's
  identical-output failure, and the test says so.

## Notes

- Sequenced with #711 at the reporter's direction: both land in the same few
  lines of the parallel emission chain, so they share one branch and one PR.
- Filed `High`: this is silent wrong output on a user-visible control, and it
  affects the default 3D finish pattern.

zabooma:
Closes #711. Closes #712.

Two defects in the same few lines of the height-map finish strategies' per-point
Z chain. They are here together because a fix to either one touches the other's
code, and because the second was found by the first one's measurements.

## #711 — a flat pass at the depth limit, over ground already cleared

Where the cutter-location surface sits **below** a floor a related subtract
owns, the clamp turned *"there is nothing here I am allowed to cut"* into
*"machine a flat pass at the limit"* — at cutting feed, over area the subtract's
own clearing operation had already taken to that floor.

`hasMachinableSurface` now decides this once, in the dispatcher, and both
strategies break the pass rather than clamping onto it: parallel by splitting
the scanline, constant scallop the way `liftFragment` already breaks on a
missing height-map sample. Waterline is untouched — it never takes this path.

### Why the rule is narrow

Two things ruled out the obvious "skip anything below the floor":

- **Tabs must keep clamping.** Where a tab top sits above the surface, riding at
  the tab top is the entire point: it machines down to the tab and preserves it.
  Skipping there would punch a hole in the finish over every tab. There is a
  test for this, and a mutation that drops the exemption fails only that test.
- **The operation's own `effectiveBottom` can never trigger this.**
  `safeToolTipZAt` is a maximum over the cutter footprint that includes `d = 0`,
  so `surfaceZ ≥ modelBottomZ` and therefore
  `surfaceZ + axialLeave ≥ modelBottomZ + axialLeave = effectiveBottom`.
  Measured: **zero** emitted cut points sit at the floor on the guitar and hills
  fixtures, for either strategy, with and without axial stock. A floor that *is*
  above the surface therefore came from a subtract, so no subtract-specific
  plumbing was needed to tell the two cases apart.

### Measured on the reported project

`work/Oldman-splash-final.camj`, the file the report came from — a model buried
in the body below a pocket floored at Z 0.2, with 38.9 % of its
cutter-location surface under that floor:

| pattern | cut length | at the Z 0.2 floor | moves |
| --- | --- | --- | --- |
| constant scallop | 287.8 → **208.3 in** | 78.5 in → **0** | 43,874 → 30,580 |
| parallel | 242.0 → **174.4 in** | 65.0 in → **0** | 10,467 → 7,112 |

At that operation's 30 ipm, **159 minutes off a 576-minute finish**.

## #712 — parallel silently ignored Stock to leave (axial)

`finishSurfaceParallel.ts` contained **no reference to `stockToLeaveAxial`**.
The field is rendered in its panel, the value is stored on the operation, and
the emitted toolpath was unchanged — a parallel finish asked to leave 0.5 mm cut
the part to final size. Waterline and constant scallop both honoured it;
`finish_surface_cleanup` at least warns it cannot (`cleanupStockToLeaveOffsets`).

Guitar fixture, one process per row so nothing shares the height-map cache:

| strategy | leave | cut points | min Z | mean Z | max Z |
| --- | --- | --- | --- | --- | --- |
| parallel, before | 0.5 | 655,518 | 0.0000 | 8.4298 | 19.6255 |
| parallel, **after** | 0.5 | 655,518 | **0.5000** | **8.9298** | **20.1255** |
| parallel, either | 0 | 655,518 | 0.0000 | 8.4298 | 19.6255 |

Exactly the leave, on every statistic.

The offset rides `applyGougeProtection` — the step that establishes the
cutter-location surface — and not the raw sample before it, because that step
raises Z to a kinematic minimum and an offset applied first would be swallowed
wherever the minimum binds. At a leave of zero the arithmetic is unchanged.

**A note for review**, since it is easy to read the code the wrong way: the
emitted Z is *not* `buildTopSurfaceSegment`'s `queryHeightMapTopZ` value. That
step samples the **raw** mesh map; `applyGougeProtection` is what lifts it. On a
45° ramp the raw sample is 20.5000 and the operation emits 21.2911.

## The one baseline that moves

`finishSurfaceSlopeParity.test.ts`'s `model-in-pocket.camj` parallel row:
**4,852 → 786 moves**. That fixture is #711's shape exactly — a flat-topped
block in a pocket floored at Z 0.5 — and its cutter-location surface has only
two levels, **0.750 on 16.3 % of cells and 0.300 on 83.7 %, with nothing between
them**. So its 4,041 moves and 126.71 of 164.32 units of cutting at Z 0.500 were
provably air, and the sloped motion that went with them was the ramps into and
out of that pass. The real Z 0.750 machining is preserved and in fact grows
slightly, 752 → 777 moves and 23.50 → 24.79 of length, because the pass is now
split at the boundary. The waterline row is byte-identical. The reason is
recorded in the baseline comment.

## Tests

New `finishSurfaceDepthClamp.test.ts`, on the tracked `model-in-pocket.camj`
plus analytic ramps. Five claims: neither height-map strategy cuts at or below
the pocket floor; the real machining survives (the anti-vacuity guard, without
which an empty toolpath would pass); a tab over the sub-floor area is still
ridden at its top; and a parallel run at a given axial leave is the run at zero
shifted up by exactly that leave, point by point, in mm and inch.

Both defects were invisible to the old suite for two specific reasons, now
covered: every finish-surface test pinned `stockToLeaveAxial: 0`, and nothing
looked at what was emitted below a subtract's floor.

All three mutations were run and each fails exactly its own tests:

- `hasMachinableSurface` forced to `true` — both pocket-floor tests fail with
  4,041 parallel moves back at Z 0.500, and so does the tab test, whose
  anti-vacuity guard requires the untabbed fixture to have no cutting there.
- the tab clause dropped — only the tab test fails, which is why the claims are
  separate.
- `sp.z = safeZ + axialLeave` reverted — only the two axial tests fail.

## Verification

- `npm run build` — exit 0 (docs, lint, e2e-lane check, colours, portable paths,
  i18n, `tsc -b`, 236 unit test files, `vite build`)
- `npm run test:e2e:workflow-ui` — 61/61 passed, isolated server
- Branched from `8a98481`, current `main`


zabooma:
## Correction: the minute figures here are wrong by 60x

Feed is already in units per **minute**, so machining time is `length / feed`. I
computed `length / feed * 60`. Every minute figure in this issue and in PR #713
(including its commit message) is therefore 60x too large.

| stated | actual |
| --- | --- |
| "157 minutes of the finish's 576" | **2.6 minutes of 9.6** |
| "159 minutes off a 576-minute finish" | **~0.6 minutes off ~11.4** |

**The proportional claims are unaffected** and remain the reason the fix is
right: 27.3 % of constant scallop's cut length and 26.9 % of parallel's was a
flat pass over ground the pocket had already cleared, and it is gone. The
geometry, the tests and the parity-baseline reasoning all stand.

## And the fix made retraction worse, which matters more than the saving

Measured on `work/Oldman-splash-final.camj`, before (8a98481) and after
(915b0bf), at the operation's own feed 30 / plunge 20:

| | before | after |
| --- | --- | --- |
| cutting | 9.6 min | **6.9 min** |
| plunging | 1.6 min | **3.4 min** |
| rapids | 0.2 min | 0.5 min |
| **total** | 11.4 min | **10.8 min** |
| retract cycles | 61 | **108** |
| **non-cutting share** | 16.1 % | **35.6 %** |

Removing the air cutting gave back 2.7 min, but the extra retracts ate 2.1 min
of it. Breaking a pass where there is no surface necessarily creates retracts —
that is the unavoidable cost side of this change, and it is why the net win is
0.6 min rather than 2.7.

The retraction behaviour it exposed is now tracked as #716, which is the real
fix: constant scallop retracts to safe Z between every contour because
`emitContours` passes `maxLinkDistance: 0` and no `safeLinkCheck`.

